PR #14, opened 2022-02-09
Feature/stage abort

jardizzo:
Error checking was added to the stage task:

(1) Source file(s) do not exist
(2) Cannot create directory
(3) File already exists (destination is a link but a file exists with the same name)

All errors were simulated using a CYLC executed experiment and verified. In addition, an experiment was executed with stage files already copied. No errors were encountered as expected. The stage task will not abort if files already exist.

### DIFF
--- a/src/swell/tasks/stage.py
+++ b/src/swell/tasks/stage.py
@@ -121,7 +121,7 @@ class Stage(taskBase):
             if os.path.islink(dest):
                 os.remove(dest)
             if os.path.isfile(dest):
-                self.logger.abort('File already exists: "' + dest + '"')
+                self.logger.abort('File exists where link expected: "' + dest + '"')
             if not os.path.isfile(dest):
                 os.symlink(src, dest)
         except Exception:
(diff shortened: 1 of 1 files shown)


src/swell/tasks/stage.py
```diff
@@ -106,8 +113,16 @@ def link_file(self, src, dest):
...
-            os.symlink(src, dest)
+        self.logger.info('Linking "' + dest + '" to "' + src + '"')
+        if not os.path.isfile(src):
+            self.logger.abort('Source file does not exist: "' + src + '"')
+
+        try:
+            if os.path.islink(dest):
+                os.remove(dest)
+            if os.path.isfile(dest):
+                self.logger.abort('File already exists: "' + dest + '"')
```
danholdaway: What do you think about making this "File already exists and was created without using a symbolic link". Just to be clear why having it exist is a problem?

danholdaway:
Thanks @jardizzo. The changes look great. I just had one minor comment on the error message for the linking part.

jardizzo:
OK, I updated the abort message to better indicate the situation where a file exists in the place where a symbolic link is expected.

danholdaway (approved):

PR #16, opened 2022-02-10
Bring swell up to date with develop.

danholdaway:
## Description

This brings swell up to date with the develop branches of JEDI, which required some Yaml changes. In doing this I decided to untether swell from the yaml files that live in fv3-jedi/ewok/geos. You can see that the contents of these files are simply moved into the experiment.yaml instead of having the yaml:: path.

It also moves all the repos and modules back to develop after they were set to specific tags for the purpose of tagging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Swell Workflow Ecosystem, Layout and Launcher (SWELL)
+# Swell Workflow Ecosystem, Layout and Launcher
 
 ### Continuous integration:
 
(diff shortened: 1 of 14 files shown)


jardizzo (approved):
I see the explicit inclusion of the YAML. I suspect this means that we will no longer rely on the automatic inclusion of upstream changes to these YAMLs? Clean-up, consolidation, syntactic changes look reasonable to me.

danholdaway:
@jardizzo that is correct. This means we have to manually maintain these files in swell and keep swell up to date with the FV3-JEDI repo. In practice this was the case anyway. I've seen several yaml-affecting changes go into fv3-jedi over the course of the last few months without these files being changed in those PRs. It doesn't get caught because those yamls are not part of any CI testing. I also dislike that the yamls live in a directory called ewok, which could confuse folks.

PR #19, opened 2022-02-14
Feature/sat db bin

asewnath:
## Description

This PR adds the satellite database utility into swell that will take the active_channels table from GEOSana_GridComp and convert it into satellite yaml files. It also extends the jedi_config task to modify channel ranges and delete obs filters in the config. 

## Acceptance Criteria (Definition of Done)

Satellite database utility has been added and tested. Jedi_config task has been updated and tested.


src/swell/tasks/jedi_config.py
```diff
@@ -35,6 +35,100 @@ def execute(self):
         # ------------
         bundle_directory = self.config.get("bundle")
         application_config = self.config.get("application_config")
+        time_window = self.config.get('current_cycle')
```
danholdaway, line 48: I wonder if we could separate the code out here to make it a little more readable. Essentially moving all the code within the `if` block to a subroutine? That way it would look like:

```
if update_channels:

    set_channels_using_satdb(self.config)
```

And then have that subroutine extract what is needed from the config?

src/swell/tasks/jedi_config.py
```diff
@@ -35,6 +35,100 @@ def execute(self):
...
+        cycle_dt = dt.strptime(time_window, '%Y-%m-%dT%H:%M:%SZ')
+
+        if update_channels:
+
+            # Manage satellite database preparation
+            # -------------------------------------
+            if not use_sat_db_yaml:
+                # Need to make sure GEOSana_GridComp is cloned and ready
+                try:
+                    os.makedirs(geos_sat_db_dir)
```
danholdaway: Can the repo be cloned into the experiment directory somewhere? Perhaps into the bundle directory? Otherwise it would never get cleaned up.

danholdaway:
Thanks for adding all this @asewnath. For me it now works from beginning to end of the workflow so I'm not sure why you were seeing the errors you were seeing. I made a few 'big picture' edits that can be seen here: https://github.com/GEOS-ESM/swell/pull/19/commits/dd5d91c318c596a55faf29e3b218508859d4074c. Mainly this was to do away with the $NOBACKUP path for the yaml files that you generated and instead move the pre-generated yaml files into swell. In the future you should be able to use your utility to regenerate these files and add them to swell where we can version control them.

danholdaway (approved):

PR #20, opened 2022-02-16
Move all Yaml files into swell and cleanup the deployment phase

danholdaway:
## Description

This moves all the yamls that used to to come from the Jedi repos to within swell. If using an existing build it is now only necessary to clone fv3-jedi when creating the experiment.

The PR also cleans up the deployment phases after several hacks were added to get the first release out the door. Namely there had gotten to be too many classes, each with their own version of the dictionary so it because hard to know what was happening where. Now there is a single dictionary that is passed around to the various methods.


danholdaway:
Note that global.cylc will need to be updated, from nccs-discover to nccs_discover.

danholdaway:
Note that I changes the default path where experiments get staged to $NOBACKUP/JediExperiments. You can always change this back in the experiment.yaml if you wish.

jardizzo:
I'm having trouble understanding this part of the pull request: "Now there is a single dictionary that is passed around to the various methods." The base class for SWELL contains the config. All classes inherit from the base class. Where are the multiple copies coming from? Perhaps this refers to another dictionary?

danholdaway:
> I'm having trouble understanding this part of the pull request: "Now there is a single dictionary that is passed around to the various methods." The base class for SWELL contains the config. All classes inherit from the base class. Where are the multiple copies coming from? Perhaps this refers to another dictionary?

@jardizzo sorry that wasn't very clear. I was referring only to the swell_create_experiment code. Previously this method created two objects, one associated with preparing the experiment directories and one with expanding the yamls. Each of these objects contained dictionaries and the latter used the dictionary from the former. I found this a bit hard to keep track of and found I was including hacks to add things to these dictionaries. I felt it was easier (for me at least) to follow if there was one dictionary, created by reading from experiment.yaml) that was later modified in the methods associated with directory creation and yaml expansion. Once I did that it also seemed to negate the need for these methods to be classes.

jardizzo (approved):
Everything looks good to me. One comment: the YAMLs, as expected, can be very large and complicated. It will be nice to see if Matt Shin's new YAML include capabilities will enable us to perhaps decompose these YAMLs into smaller logical units that can then be combined at runtime. We need to learn more about his package capabilities.

danholdaway:
Thanks @jardizzo I've been thinking along the same lines. Perhaps we have each observation filter be included as a yaml file instead of in a large monolithic yaml.

PR #23, opened 2022-02-23
Updates to eva API

danholdaway:
## Description

Recent changes to the eva API means changes to where swell accesses eva.

## Dependencies

- [ ] waiting on https://github.com/danholdaway/eva/pull/14

### DIFF
--- a/src/swell/deployment/platforms/nccs_discover/modules
+++ b/src/swell/deployment/platforms/nccs_discover/modules
@@ -25,5 +25,5 @@ export DHJEDIOPT=/discover/nobackup/drholdaw/JediOpt/
 module use -a $DHJEDIOPT/modulefiles/core
 module load solo/swell-1.0.0
 module load r2d2/swell-1.0.0
-module load eva/1.0.0
+module load eva/1.1.0
 module load swell/1.0.1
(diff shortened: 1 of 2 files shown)


jardizzo (approved):
Looks like create_and_run() is now part of object instantiation? The only comment that I have is that "Run the diagnostic application" is obfuscated by the instantiation versus the invocation of a method of the object that has an explicit name (e.g. eva.run_diag()). I may be misinterpreting the code however.

danholdaway:
> Looks like create_and_run() is now part of object instantiation? The only comment that I have is that "Run the diagnostic application" is obfuscated by the instantiation versus the invocation of a method of the object that has an explicit name (e.g. eva.run_diag()). I may be misinterpreting the code however.

@jardizzo I'm not sure I understand your comment. Could you elaborate? In this case the thing provided by eva is just a method rather than an object. The method will create an object but it's not accessible to swell.

jardizzo:
OK, I understand now. The eva() method instantiates an object generated by a factory that is a function of the dictionary passed in. So my comment is simply that eva() does not denote an action. Perhaps eva_run()? Feel free to ignore.

danholdaway:
That was already decided in an eva code sprint development last week and merged into develop so we have to take it at face value.

PR #26, opened 2022-03-01
Feature/stage new

jardizzo:
New staging codes with support for multiple YAML data structures describing files to be acquired. Both the "STAGE" file structure and the "GetData" file structure are now supported. Both of these data structures are lists of dictionaries. However, the original "STAGE" file structure is somewhat limiting. The "GetData" file structure allows expanded options as follows:

(1) Support for multiple input source directories
(2) Sentinel parameters for implementing wait logic (minimum file count, minimum file age)
(3) Extensible (easy to add new dictionary parameters and extend the code)
(4) Flexible shortcuts for specifying files or whole directories to be copied
(5) Ability to rename files 

In addition, SWELL exceptions were added as a way of uniquely tagging errors.

Remaining Work:

(1) The current codes still use globs. Need to switch to os.listdir?
(2) GetData remains file-centric (i.e. recursive directory copies are not allowed). The parameter name for specifying items to be copied remains as "files". Entire directories can be specified using wildcards. We need to decide if this needs to be changed (e.g. "items" instead of "files" and allow directory names and recursive copies).

Examples:

(1) Copy all files in directory

src: /data/dir1
dst: /output/dir

(2) Copy random list of files

files:
  - /data/dir1/file1 /output/dirout
  - /data/dir2/file2 /output/dirout2
 
(3) Copy select files or files of select type

src: /data/dir1
dst: /data/output
files:
  - *.yaml
  - *.nc4
  - test.data

(4) Link files

src: /data/dir1
dst: /data/output
link: 1
files:
  - *.yaml


danholdaway:
Changes look good @jardizzo and the workflow ran all the way through for me. There's just some coding norms to resolve and then we should merge. You can improve/modify the things you mentioned in some subsequent PRs.

danholdaway (approved):

PR #33, opened 2022-03-28
Adding SaveObsDiag and CleanCycle

asewnath:
## Description

Adding the SaveObsDiag and CleanCycle tasks to the workflow.

## Acceptance Criteria (Definition of Done)

Both tasks execute successfully in the cylc environment.The SaveObsDiag task successfully stores merged ioda files from the current experiment cycle into a user's local R2D2DataStore. The CleanCycle task adds a note to a given cycle directory to indicate that the cycle is complete. It also removes all files specified by the "CLEAN" list in the experiment yaml file. 

## Dependencies

It requires that R2D2_CONFIG be set such that the Local database field points to the user's local datastore explicitly. It does not work with a config.yaml file that uses environmental variables. Currently this issue is resolved by creating a config.yaml file in $NOBACKUP and then setting R2D2_CONFIG explicitly to that file in the experiment modules.


danholdaway:
Thanks for doing all this Akira. The changes look good. The only concern I have is with the timing of setting ${R2D2_LOCAL_PATH} in the suites/r2d2_config.yaml file. Unless I'm missing something I think this should be done at swell_create_experiment time. It looks like you are doing this in the JediConfig task but I think it should be done earlier than that. Users who are experimenting with new data (needed at GetBackground or GetObservation time) might want to first make it available through some Local db. So can you move the scripting that fills this template in r2d2_config.yaml into swell_create_experiment?

src/swell/suites/hofx/r2d2_config.yaml
```diff
@@ -0,0 +1,21 @@
+databases:
+
+  local:
```
danholdaway: Can you change the names from `local` and `shared` to `user` and `gmao-shared`? I think that will make things clearer for users.

src/swell/tasks/clean_cycle.py
```diff
@@ -0,0 +1,47 @@
...
+        cfg = self.config
+        logger = self.logger
+        cycle_dir = self.config.get("cycle_dir")
+        clean_list = cfg.get('CLEAN')
+        current_dir = os.getcwd()
+
+        if os.path.isdir(cycle_dir):
+            os.chdir(cycle_dir)
+            # Remove all specified files
+            for pattern in clean_list:
```
danholdaway, line 39: For security the user should not be able to provide a search criteria that includes paths. Perhaps some additional safety could be provided by ensuring that the string is not just `'*'`? As in the utility cannot remove all files. There has to be some kind of characters in the string it will remove. `'*.nc4'` is allowed but `'*'` is not. I presume os.remove will not remove directories. I think this OK. We could always add the option to remove directories later on.

danholdaway (changes requested):
A couple of minor changes but otherwise this looks great. Thanks!

src/swell/deployment/bin/swell_create_experiment.py
```diff
@@ -159,6 +159,34 @@ def main(config, clean):
...
+        module_file.write('export R2D2_CONFIG={}'.format(r2d2_conf_path))
+
+    # Open the r2d2 file to dictionary
+    # ------------------------------------
+    with open(r2d2_conf_path, 'r') as r2d2_file:
+        r2d2_conf = yaml.safe_load(r2d2_file)
+
+    # Extract local path variable and replace with value from config
+    # --------------------------------------------------------------
+    element = r2d2_conf['databases']['local']['root']
```
danholdaway: Just a suggestion but you might be able to do this replacement with:

```
    from swell.utilities.string_utils import replace_vars

    with open(r2d2_file, 'r') as r2d2_file_open:
        r2d2_file_str = r2d2_file_open.read()
        r2d2_file_str = replace_vars(r2d2_file, **experiment_dict)

    with open(r2d2_file, 'w') as r2d2_file_open:
        r2d2_file_open.write(r2d2_file_str)
```

src/swell/tasks/clean_cycle.py
asewnath: yeah, os.rmdir() would be the function that removes directories

danholdaway:
Thanks Akira. The changes look great. I just made a couple of small tweaks after I prepared a new R2D2 module that does not define the path to the config. Just running the workflow one more time with cylc and then will merge.

danholdaway (approved):

PR #36, opened 2022-03-31
Add more IODA platforms and change the way they are included

danholdaway:
## Description

Add some more IODA platforms to swell and allow for easier addition of platforms not already added.


src/swell/configuration/observation_ioda_names.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - amsua_metop-a
+  - amsua_metop-b
+  - amsua_metop-c
+  - amsua_n15
+  - amsua_n18
+  - amsua_n19
+  - atms_n20
+  - atms_npp
+  - avhrr3_metop-a
+  - avhrr3_n18
```
HamidehGMAO, line 15: We should add avhrr3_metop-b here. We can do it when adding the avhrr3_metop-b yaml file too

danholdaway: Sounds good. I think that makes sense, to make the change when the YAML gets added.

gmao-jjin3: Can you add "amsr2_gcom-w1" as well? Thanks.

gmao-jjin3 (approved):

HamidehGMAO:
It works fine Dan. Thanks for the fix!

HamidehGMAO (approved):

src/swell/configuration/observation_ioda_names.yaml
danholdaway: Done

src/swell/configuration/observation_ioda_names.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - avhrr3_n19
+  - cris-fsr_n20
+  - cris-fsr_npp
+  - gpm_gmi
```
gmao-jjin3, line 19: It should be  "- gmi_gpm".

gmao-jjin3 (commented):
Need change "gpm_gmi" to "gmi_gpm".

PR #38, opened 2022-04-02
Make the version of swell consistent between create_experiement and workflow.

danholdaway:
## Description

Set the version of swell used in the workflow automatically so it is consistent with the version used at create_experiment time


gmao-jjin3 (approved):
I have to say that I don't really know these modules.

src/swell/deployment/prep_exp_dirs.py
```diff
@@ -72,3 +75,73 @@ def copy_suite_files(logger, experiment_dict):
...
+
+    # Swell bin path
+    # --------------
+    swell_bin_path = shutil.which("swell_task")
+    swell_bin_path = "/".join(list(swell_bin_path.split('/')[0:-1]))
+
+    # Swell lib path
+    # --------------
+    swell_lib_path = swell_install_path()
+    swell_lib_path = "/".join(list(swell_lib_path.split('/')[0:-1]))
```
jardizzo: You may want to use os.path.join and eliminate any explicit references to the OS separator ('/').

src/swell/deployment/prep_exp_dirs.py
```diff
@@ -72,3 +75,73 @@ def copy_suite_files(logger, experiment_dict):
...
+    suite_dir = experiment_dict['suite_dir']
+
+    # Modules file
+    # ------------
+    modules_file = os.path.join(suite_dir, 'modules')
+
+    # Swell bin path
+    # --------------
+    swell_bin_path = shutil.which("swell_task")
+    swell_bin_path = "/".join(list(swell_bin_path.split('/')[0:-1]))
```
jardizzo: You probably don't need to instantiate a copy of the split using "list". You could also try using os.path.split:

swell_bin_path = os.path.split(swell_bin_path)[0]

jardizzo (approved):
There are lots of references to '/' in the added code. This will be fine if there is never any reason to expect the code to run on a system that does not use "/" to delineate path components. Likewise, the replacement targets in the modules files also assume a "/" for pathnames. That may be ok since bash and csh are assumed to be running on a Unix-based OS.

danholdaway:
Thanks for the suggestions @jardizzo. It now makes use of `os.path.split` instead of relying on the `/` delineator.

PR #39, opened 2022-04-03
Adding gmi_gpm.yaml

gmao-jjin3:
Adding gmi_gpm.yaml and additional clouds qr and qs in experiment.yaml. Bias correction is not activated in gmi_gpm.yaml.

### DIFF
--- a/src/swell/configuration/observation_operators/gmi_gpm.yaml
+++ b/src/swell/configuration/observation_operators/gmi_gpm.yaml
@@ -19,30 +19,30 @@ obs operator:
     Sensor_ID: gmi_gpm
     EndianType: little_endian
     CoefficientPath: $(crtm_coeff_dir)
-#obs bias:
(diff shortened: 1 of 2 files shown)


danholdaway:
What's the reason that bias is not activated? Are there missing files?

gmao-jjin3:
> What's the reason that bias is not activated? Are there missing files?

No, there aren't.  Swell couldn't produce Hofx even before bias correction.  I am re-running x0044 with the CRTM in JEDI. Most of radiance observations have same forward results from different CRTM versions, but there are different results in GMI, AMSR2,  MHS, and ATMS observations. 

Anyway, they are activated now, so that we can have a check whether there are any missing files.

gmao-yzhu:
> we can have a check whether there are any missing files

OK I see, you added bias section there so we can have a check whether there are any missing files.

gmao-yzhu (approved):

danholdaway (approved):

PR #41, opened 2022-04-07
add mls ozone yaml.

karpob:
Adding mls55_aura.yaml , which adds the ozone level instrument microwave limb sounder on aura satellite. Includes a sanity check (<10000), and gross check following the GEOS-FP GSI configuration with a threshold of 5.0*sigma_o.


danholdaway (approved):
Looks great. Thanks!

PR #43, opened 2022-04-08
add workflow for converting GEOS backgrounds to R2D2

danholdaway:
## Description

This PR adds two new tasks. One for collecting the background from a previously run GEOS experiment, instead of R2D2. And another which stores any background to R2D2. 

It then adds a new workflow suite that will retrieve the backgrounds from the GEOS experiment and store them in R2D2. In essence converting the backgrounds generated from a GEOSadas experiment to R2D2, where they can be used in a pure GEOS-JEDI experiment.

This workflow is a bit simpler than the previously implemented workflow since it doesn't have any SLURM requirement. The PR includes some changes to the create experiment step to allow workflows to be able to not provide config for scheduling.


danholdaway:
Thanks @asewnath for adding the R2D2 config file. It makes me realize though. Do you think it's in the wrong place? I suppose this is not something that changes with the suite but maybe would change with the platform. Perhaps it needs to be in deployment/platform/discover_nccs directory?

asewnath:
> Thanks @asewnath for adding the R2D2 config file. It makes me realize though. Do you think it's in the wrong place? I suppose this is not something that changes with the suite but maybe would change with the platform. Perhaps it needs to be in deployment/platform/discover_nccs directory?

Ah okay I remember us talking about putting it in the platform directory. I will correct that.

asewnath (approved):

PR #54, opened 2022-04-22
Adding widget types to GUI

asewnath:
## Description

Adds radio buttons and dropdown menus into the GUI using the current framework. Removed lists and for loops so that check_widget calls the make widget functions directly.


CRWayman (approved):
looks great

PR #82, opened 2022-05-24
GUI changes into develop

danholdaway:
## Description

GUI changes into develop


jardizzo (approved):
Looks good. In a sense, a pulldown is a long-listed radio button. A radio button is a list with a default (denoting single value return) and a check box is a list with a multi-value return (no default). Can we possibly infer the widget type from the data structure and the presence of a "default"? Also , is it possible that the YAML data structures could self-describe the GUI. For example, a dictionary of dictionaries would yield nested clickable buttons that end with a widget operating on the leaf nodes. Just food for thought. I have not vetted these ideas through trial and error.

PR #83, opened 2022-05-25
New merge tool and new JEDI build

danholdaway:
## Description

- Update to new version of JEDI code
- Use bespoke tool for merging the IODA files.


src/swell/tasks/jedi_config.py
```diff
@@ -32,6 +34,26 @@ def ranges(i):
...
 
+    def replace_jedi_conf(self, jedi_config):
+
+        # Loop over dictionary and replace if element_template is a dictionary
+        for key, element_template in jedi_config.items():
+            if isinstance(element_template, dict):
+                self.replace_jedi_conf(element_template)
+            else:
+                # Strip special characters from element
+                element = copy.deepcopy(element_template)
```
jardizzo, line 45: Assumption here is that the element is a string? Have all lists been eliminated? Do you need a deep copy for a string var. You could also use the collections.abc Hashable method to determine if you have a scalar: https://www.programcreek.com/python/example/6735/collections.Hashable .

src/swell/tasks/jedi_config.py
```diff
@@ -32,6 +34,26 @@ def ranges(i):
...
+                element = element.replace("$", "")
+                element = element.replace("{", "")
+                element = element.replace("}", "")
+
+                try:
+                    # Replace with element from filled config
```
jardizzo, line 51: Could you use replace_vars() here? You just need to pass in **self.config or definitions from self.config: defs = {k:str(v) for k,v in iter(config.items()) if isinstance(v, Hashable)} . Note: from collections.abc import * . This would take care of variables that resolve to additional variables.

gmao-jjin3:
I don't have any comments about these updates, but I have a question:  Are the geoval files produced here identical to geoval files made from GSI outputs?  It is great if they are. We have to be careful is they are not.

danholdaway:
> I don't have any comments about these updates, but I have a question: Are the geoval files produced here identical to geoval files made from GSI outputs? It is great if they are. We have to be careful is they are not.

@gmao-jjin3 the intention is not to use these GeoVaLs for running tests in UFO. More just to make comparison plots against the GSI GeoVaLs to make sure the values look similar.

src/swell/tasks/jedi_config.py
danholdaway: What about variables that resolve to dictionaries? I wasn't sure if that would work with `replace_vars`?

src/swell/tasks/jedi_config.py
danholdaway: When I was developing something was going wrong that was fixed with using deepcopy but I don't recall what and it's possible that I made other changes that makes the deepcopy superfluous. I'll test without it.

jardizzo: You will certainly have a problem if it is list that is being copied. Otherwise you should not need deep copy for string, ints, floats and other native data types. One caution about YAML configs to mention: We need to be careful if we do make use of YAML anchors. This YAML syntax, dict2: *DICT1 is equivalent to dict2 = dict1 and both will reference the same memory space in the config dictionary. So far we are not using anchors so probably not an issue. This syntax, however, creates a new dictionary:

dict2:
  <<: *DICT1

Keep it in mind if we do decide to use anchors in YAML.

src/swell/tasks/jedi_config.py
jardizzo: replace_vars() will not work if the value is a dictionary. That is correct. However, it would work if you passed your resolved variable back through the YAML interpreter but that probably convolutes the code. It would work because str(dict) yields a JSON string that YAML can interpret (need to verify).

src/swell/tasks/jedi_config.py
```diff
@@ -32,6 +34,26 @@ def ranges(i):
...
+                    jedi_config[key] = self.config.get(element)
```
jardizzo, line 52: You might need a deep copy here if the retrieved dictionary contains any nested dictionaries or lists. Perhaps it is assumed here that the dictionary is flat?

jardizzo (approved):

PR #85, opened 2022-06-02
Diagnostic upgrade in Eva

danholdaway:
## Description

Brings in a new version of Eva that allows for plots that differentiate between observations that is filtered out by quality control and and that which is not.

It also brings in time interpolation in the observation operators, which should improve the fit to GSI.


src/swell/utilities/dictionary_utilities.py
```diff
@@ -52,3 +52,108 @@ def resolve_definitions(dictionary):
...
+    # Convert back to dictionary
+    d_interp = yaml.safe_load(d_string)
+
+    return d_interp
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+def dict_iterate_and_remove(d, key_to_remove, found):
```
danholdaway: @jardizzo I would appreciate if you could look at this code. It's possible you've thought about ways to do this in the past. What I'm trying to accomplish is to recursively search through a dictionary that may be nested and may include lists of dictionaries and remove matching keys that might be part of the lists or nested dictionaries. It doesn't work to convert to string and remove matching lines. Consider the following for example:

``` yaml
- key_to_remove: element
  next_key: next element
```

Removing the line `- key_to_remove: element` would remove the list indicator.

asewnath:
Hey @danholdaway I tried running this branch and I got stuck on MergeObsDiags with this error: ValueError: shape mismatch: objects cannot be broadcast to a single shape

src/swell/utilities/dictionary_utilities.py
jardizzo: I will take a look now. I may have something. So the goal is to remove all matching keys?

jardizzo: Dan, I have note tested this but we can give it a try:

    def delete_key(key, config):

        for node in config:

            if isinstance(node, dict):

                node.pop(key, None)

                for k,v in iter(node.items()):
                    if isinstance(v, dict): delete_key(key, v)

            elif isinstance(node, list):

                for v in node:
                    if isinstance(v, dict): delete_key(key, v)

jardizzo (commented):
The new method to remove dictionary keys was added and tested. All recent changes are bundled under this pull request. Please note that the new method, remove_matching_keys(), modifies the data structure in place and does not return a copy. The Eva_driver methods were also updated to be compatible with this change.

src/swell/utilities/dictionary_utilities.py
danholdaway: Thanks for adding the new code @jardizzo

Issue #59, opened 2022-05-12
YAML Configuration | AMSUA


danholdaway:
## Agreement with GSI using GeoVaLs in UFO (Hofx no bias correction)

- [x] AQUA
- [x] METOP-A
- [x] METOP-B
- [x] METOP-C
- [x] NOAA-15
- [x] NOAA-18
- [x] NOAA-19

## Agreement with GSI using GeoVaLs in UFO (Bias correction)

- [x] AQUA
- [x] METOP-A
- [x] METOP-B
- [x] METOP-C
- [x] NOAA-15
- [x] NOAA-18
- [x] NOAA-19

## Agreement with GSI using GeoVaLs in UFO (Quality control)

- [x] AQUA
- [x] METOP-A
- [x] METOP-B
- [x] METOP-C
- [x] NOAA-15
- [x] NOAA-18
- [x] NOAA-19

## Agreement with GSI using GeoVaLs in UFO (Hofx with bias correction)

- [x] AQUA
- [x] METOP-A
- [x] METOP-B
- [x] METOP-C
- [x] NOAA-15
- [x] NOAA-18
- [x] NOAA-19

## Agreement with GSI using Swell

- [ ] AQUA
- [ ] METOP-A
- [ ] METOP-B
- [ ] METOP-C
- [ ] NOAA-15
- [ ] NOAA-18
- [ ] NOAA-19

## Current issues:

- Need new GeoVaLs from GSI

gmao-yzhu:
AMSU-A NOAA19 has passed UFO tests, code changes and two new yaml files were merged into UFO develop.

gmao-yzhu:
When testing NOAA18, I found that UFO results didn't match with GSI. Jianjun made plots and further narrowed the issue down to the locations over ice, i.e., when ice-fraction=1 and sst=260.

I diff-ed CRTM code and fixed files and confirmed we are using the same version in GSI with UFO. 
When I read the CRTM codes, I noticed that over ice, sst will be re-set to 260 if it is >= 280. 

So this issue may have something to do with how surface type is decided and sst values in geoval file. In GSI, it is considered as ice when ice-fraction>0.99.  Jianjun is re-generating testing files to see if new files will make any difference.

gmao-yzhu:
@gmao-jjin3 I added some printout from UFO calculation.

Here are the locations where sst=260.0 with inconsistent Tb with GSI:
```
         profile.   ice-frac.     sst.           Tb
[...]

PR #90, opened 2022-06-08
changes of AMSU-A yaml files for simulation, bias correction, quality control

gmao-yzhu:
This is corresponding to Issue #59. 
With the codes and yaml file changes for GEOS merged into JEDI fv3-bundle UFO, this pull request will get the yaml files into Swell.


danholdaway (approved):
Thanks @gmao-yzhu! Great to have the first official YAML files in Swell! And this will be very useful for testing the surface geovals.

PR #101, opened 2022-07-25
Cylc Obs Workflow

asewnath:
## Description

Cylc workflow for preprocessing observation and satbias files into JEDI-compliant formats for SWELL experiments.

## Acceptance Criteria 

All observations for a given cycle should be retrieved in the Cylc run directory, preprocessed, and saved into the user's local R2D2DataStore with the correct datetimes


danholdaway:
@asewnath can you check on the coding norms please? It looks like that test is failing.

danholdaway:
In the list of changed files I see a few unexpected changes. For example a series of files in a directory called `/bin`. And then a bunch of files called `*.sw*`. Also the `satwind.yaml`. Could you remove those or revert them to be the same as develop?

src/swell/suites/process_obs/experiment.yaml
```diff
@@ -0,0 +1,74 @@
+# Experiment name and location
+experiment_id: x0044-jedi-process_obs
+experiment_root: /discover/nobackup/${USER}/JediExperiments/
```
danholdaway:
```suggestion
experiment_root: /discover/nobackup/${USER}/SwellExperiments/
```
So it's consistent with the other experiments.

danholdaway (changes requested):
Just one small change request. Otherwise this runs for me and looks like it produces the correct thing.

asewnath:
satbias processing isn't working properly due to HDF5 version mismatch errors

src/swell/suites/retrieve_obs_list.py
```diff
@@ -0,0 +1,15 @@
+from swell.configuration.configuration import return_configuration_path
```
danholdaway: What is the role of this file? I don't see it being referenced anywhere. And it seems like it's doing task like work so I'm nor sure it belongs in the suite directory.

asewnath: I actually don't even remember why I made this file, it probably had to do with logic I made for the GUI sprint because, like you say, it wouldn't be in the suites directory otherwise. I'll delete that

danholdaway (approved):

PR #108, opened 2022-08-10
Use JEDI Spack modules and update Eva

danholdaway:
## Description

- Update to use the JEDI spack modules and use a new JEDI build based on Spack.
- Update to eva 1.2.2 and utilize the features of that tag to include histogram and map plots.
- Clean up the CI workflows
- Bump version ready for tagging


jardizzo (approved):
Looks like mostly code style changes outside of EVA plot additions and spack stack. Do we need an updated on-prem workflow to build the code on discover?

danholdaway:
That's correct @jardizzo. For now I've turned off the on-prem workflow until we have things a bit more stable. If possible can you work from a branch from now on, instead of develop?

PR #112, opened 2022-09-09
Feature/add ompslp

karpob:
## Description
This adds omps-lp instrument for future versions of FP (for use when MLS dies). I ran one cycle through x0044 background and it seems fine.  See correlation plot from EVA:
image
 

## Dependencies
None

## Impact
None


danholdaway (approved):
Thanks for adding this. Looks good.

PR #125, opened 2022-10-17
Feature/update jedi

danholdaway:
## Description

Changes to Swell to allow building with the latest develop versions of the JEDI repos.


jardizzo (approved):
The nightly test that builds JEDI and runs the HOFX use-case ran without errors.

PR #127, opened 2022-10-28
New config and ready for coupled mode

danholdaway:
Big refactoring that achieves two main goals:

1. Refactor the initial config

- Experiment.yaml only contains things the users will adjust.
- Config that is more fixed is moved the configuration directory and loaded on the fly.
- Config files are all copied to the experiment directory to allow reproducibility.
- Config is ready for TUI and GUI with baked in questions.
- Options for CI/CD are available through the platform config.
- Remove bespoke templating and use jinja2 instead.

2. Prepare to run coupled
- Jinja2 templating in the suite file that are resolved based on user model choices.
- Task receive extra argument of model when approprirate.
- Config will combine top level config with the config with the chosen model so tasks only see 'their model'.


danholdaway:
@jardizzo we should get the CI running from this branch. It adds this file: https://github.com/GEOS-ESM/swell/blob/feature/newconfusage/src/swell/deployment/platforms/nccs_discover/experiment.yaml where you can specify CI experiment info automatically. Jinja2 will complete the templates including a datetime for the experiment ID so you can always have a unique experiment ID.

jardizzo (approved):
So it looks like all templating is jinja2 with some additional variable substitutions of the form $(var) ? Seems like a good decision. I think initially we were trying to remain backward compatible with EWOK conventions. Regarding the new CICD hook in the platform files: having the ability to encode compound alternative options within YAML definitions is a great configuration mechanism. I suppose the only negative is that the SWELL repo will now dictate the CICD implementation regarding the experiment names/directories. The only other approach is to have the CICD workflow dynamically resolve the experiment details on the fly be assuming that the pertinent parameter names will never change. A simple string edit (sed for example) would substitute the values. Alternatively, we could adopt global variables for the CICD workflow to resolve. For example:

ci_cd:
  experiment_id: $CICD_EXPERIMENT_ID
  experiment_root: $CICD_EXPERIMENT_ROOT

The CICD workflow could do a simple envsubst on the experiment file before swell_create_experiment is invoked. A third option could be to let swell_create_experiment override defaults with additional command-line options: --exp_id, --exp_root . That may negate the need for special configuration blocks (i.e. always resolve the parameters on the fly). Users can still choose to hardcode values by removing the templated placeholders ($EXP_ID, $EXP_ROOT).

mathomp4:
Note: I think the YAML Linter is having issues with this because when YAML sees a brace right after a colon a la:
```yaml
      obsfile: {{cycle_dir}}/atms_npp.{{window_begin}}.nc4
```
it assumes it's a flow mapping. And I guess the parser sees the right-brace at the end of  `{{cycle_dir}}` as the end of the mapping and then expects the block end:
```bash
$ yamllint -d relaxed --no-warnings atms_npp.yaml
atms_npp.yaml
  6:29      error    syntax error: expected <block end>, but found '<scalar>' (syntax)
```
About the only way I can think of fixing this is to quote the path:
```yaml
      obsfile: '{{cycle_dir}}/atms_npp.{{window_begin}}.nc4'
```

but not sure if this then runs.

danholdaway:
> Details

Thanks @mathomp4. Makes sense. I should be able to quote it since the value is supposed to be a string. It would be a problem if Jedi were expecting an integer, float etc but since it expects a string it should work OK. Will give it a try.

mathomp4:
> > Details
> 
> Thanks @mathomp4. Makes sense. I should be able to quote it since the value is supposed to be a string. It would be a problem if Jedi were expecting an integer, float etc but since it expects a string it should work OK. Will give it a try.

Of course, then the fun is double- vs single-quotes, but every time I try and [read the spec on this](https://yaml.org/spec/1.2.2/#73-flow-scalar-styles), I just try both and see what works. 😄

danholdaway:
> > > Details
> > 
> > 
> > Thanks @mathomp4. Makes sense. I should be able to quote it since the value is supposed to be a string. It would be a problem if Jedi were expecting an integer, float etc but since it expects a string it should work OK. Will give it a try.
> 
> Of course, then the fun is double- vs single-quotes, but every time I try and [read the spec on this](https://yaml.org/spec/1.2.2/#73-flow-scalar-styles), I just try both and see what works. 😄

It's passing the tests now and the workflow still runs

danholdaway:
> So it looks like all templating is jinja2 with some additional variable substitutions of the form $(var) ? Seems like a good decision. I think initially we were trying to remain backward compatible with EWOK conventions. Regarding the new CICD hook in the platform files: having the ability to encode compound alternative options within YAML definitions is a great configuration mechanism. I suppose the only negative is that the SWELL repo will now dictate the CICD implementation regarding the experiment names/directories. The only other approach is to have the CICD workflow dynamically resolve the experiment details on the fly be assuming that the pertinent parameter names will never change. A simple string edit (sed for example) would substitute the values. Alternatively, we could adopt global variables for the CICD workflow to resolve. For example:
> 
> ci_cd: experiment_id: $CICD_EXPERIMENT_ID experiment_root: $CICD_EXPERIMENT_ROOT
> 
> The CICD workflow could do a simple envsubst on the experiment file before swell_create_experiment is invoked. A third option could be to let swell_create_experiment override defaults with additional command-line options: --exp_id, --exp_root . That may negate the need for special configuration blocks (i.e. always resolve the parameters on the fly). Users can still choose to hardcode values by removing the templated placeholders ($EXP_ID, $EXP_ROOT).

Sounds good about using environment variables. I think that makes good sense. I'll put some login in to take care of that.

danholdaway:
@jardizzo the latest commit should add what you are requesting. If cicd mode is selected you have to have two environment variables set `CICD_EXPERIMENT_ID` and `CICD_EXPERIMENT_ROOT`. The code will resolve those variables in the experiment YAMLs as it creates the experiment. If this needs further tweaking as you put the CI/CD together you can make those changes in your branch post develop merge of this branch.

PR #128, opened 2022-11-02
Improve the prints from launch experiment script

danholdaway:
## Description

A bit of clean up of the prints that come out of the `swell_launch_experiment` job.


jardizzo (approved):
Looks good to me. I assume the control sequences cause the text to be highlighted in different colors. Cool!

danholdaway:
@jardizzo that's correct. This is what all the output now looks like for create and then launch experiement.

Screen Shot 2022-11-03 at 12 46 56 PM

PR #129, opened 2022-11-17
Soca HofX is turned on in the workflow

Dooruk:
## Description

- Turning on SOCA in the workflow


jardizzo (approved):
Hi Doruk, Changes look reasonable. Looks like you qualified the stage subdirectories to accommodate ocean and atmosphere files. I cannot discern any problems based on a quick review. Did tests yield success for atmosphere / ocean experiments? I suspect the answer is yes. Apologies if I missed this in PR overview.

danholdaway (approved):

PR #132, opened 2022-12-05
Build JEDI in the workflow

danholdaway:
## Description

- Add the capability to build the JEDI code in a workflow.
- Adds workflow for building the JEDI code separately from other things.
- Splits the JEDI building into a clone and build task. The latter runs as a slurm task.


src/swell/deployment/prep_config.py
```diff
@@ -36,16 +36,15 @@ def platform_fill(logger, experiment_dict, ci_cd, comment_dict):
...
-
     # Resolve any templates
-    platform_exp_str = template_string_jinja2(logger, platform_exp_templated, temp_experiment_dict)
+    platform_exp_str = template_string_jinja2(logger, platform_exp_templated, experiment_dict)
 
     # Load to dictionary
     platform_dict = yaml.safe_load(platform_exp_str)
 
+    # Copy experiment dictionary
+    experiment_dict_new = experiment_dict
```
jardizzo, line 46: Is this intended to be a deep copy? Any nested dictionaries or lists will be address copied.

danholdaway: I don't think a deep copy would be needed here.

jardizzo (approved):
Looks good as far as I can tell. We need to eventually merge the minor changes that I made in feature/cicd_mods.

PR #136, opened 2022-12-20
Feature/3dvar generateb

Dooruk:
Introduces 'bump' background error model in the workflow.

Generates or obtains bump files depending on the number of processors requested in the suite file.


Dooruk:
Exception could be improved. Currently bump generation gives segmentation fault error.  I will fix both in the 3dvar_execute sub branch.

PR #139, opened 2023-01-25
Feature/ioda v3

danholdaway:
## Description

Upgrade to the use IODA version 3 variable name conventions. 

Other side changes:
- Add the `BuildJediByLinkingTask`. If this task fails because the code needs to actually be built then the `BuildJedi` task is triggered. But if `BuildJediByLinkingTask` succeeds then `BuildJedi`, which requires compute resources, never runs.
- Removes the need to know where the build is coming from in the modules file. I.e. no longer depend on modules in the nightly build. `module use` on the nightly modulefiles path was taking a long time for some reason so this also speeds up the workflows quite a bit.

## Dependencies

N/A

## Impact

N/A


jardizzo (approved):
Really cool what can be done with the CYLC syntax for describing conditional workflow dependencies! I saw something about not depending on the nightly build modules because it was too slow to load? I suppose I need to understand that one. There were some symbolic links used to replicate the expected directory structure that seemed to differ from  what is delivered from the JediBundle task under SWELL.

gmao-jjin3 (approved):

danholdaway:
Hi @jardizzo. That's right. We had this line in the modules file:
`module use -a /discover/nobackup/gmao_ci/swell/nightly/modulefiles/`
Executing this was very slow for me and for the swell tasks that were running in slurm I saw it take up to 20mins. I'm not sure what the problem is in that directory structure is but in any case it's not quite the right thing to do. In the case that the user selects to create their own JEDI build they would want to use IODA's Python API from their own build. So I took out of your modules file the PYTHONPATH bits and put it into the modules files but pointing to the experiment build. In the case that the user chooses to link to the nightly then they will have the same thing, but in the case that they choose to build in the experiment it will also point to the right thing.

src/swell/tasks/merge_ioda_files.py
```diff
@@ -1,280 +0,0 @@
-# (C) Copyright 2021-2022 United States Government as represented by the Administrator of the
-# National Aeronautics and Space Administration. All Rights Reserved.
-
-# This software is licensed under the terms of the Apache Licence Version 2.0
-# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
-
-
-# --------------------------------------------------------------------------------------------------
-
```
Dooruk, line 9: Is this task not needed anymore? Is it done automatically somewhere else?

src/swell/deployment/platforms/nccs_discover/modules
```diff
@@ -20,15 +20,17 @@ module load jedi-fv3-env/1.0.0
 module load nco/5.0.6
 module load sp/2.3.3
```
Dooruk, line 22: While this file is being changed, may I suggest adding:
```
# Unlimited Stacksize
# --------------------
ulimit -S -s unlimited
ulimit -S -v unlimited
umask 022
```

Dooruk (approved):
The modules file change can be included since yaml check still fails, other than that it looks fine. BuildJedibylinking is a great addition.

src/swell/tasks/merge_ioda_files.py
danholdaway: I removed the task because it wasn't working with the new IODA conventions. I didn't really want to update it because IODA now has the capability of writing a single file, and that's how we run it. Previously IODA would only write the the same number of files as processors and they would have to be combined with this tool. We can revive it from an old tag if we ever need.

src/swell/deployment/platforms/nccs_discover/modules
danholdaway: Good call, I will do that.

danholdaway:
> Really cool what can be done with the CYLC syntax for describing conditional workflow dependencies

Definitely really nice to be able to have a task trigger only on failure. The only thing that is a slight annoyance is that it appears read in the TUI when it fails. I wonder if they could have had it present orange or something that denotes that it's failure wont hold up the workflow. It means users might need some extra education on these non-blocking failures.

danholdaway:
> BuildJedibylinking is a great addition.

We should do the same thing for the static background model in the 3DVar to avoid waiting in the queue just to link some files. There it is of particular importance too because that task may require a larger number of nodes for high resolution cases.

PR #141, opened 2023-02-01
Customize the bundles which get built in BuildJedi task

danholdaway:
## Description

Instead of building all available bundles, let the user choose which bundles they want to build. Set the default in the sample suite to build fv3-jedi, soca and iodaconv, which are the only bundles that swell currently uses.

In a single test the build took 53 minutes of the 4 hours permitted.

`/home/drholdaw/cylc-run/swell-build_jedi-suite/run1/log/job/1/BuildJedi/01/job.out`

## Dependencies

N/A

## Impact

N/A

### DIFF
--- a/src/swell/deployment/bin/swell_create_experiment.py
+++ b/src/swell/deployment/bin/swell_create_experiment.py
@@ -45,7 +45,7 @@ def main(config):
     # Generate the configuration file
     # -------------------------------
     if config is None:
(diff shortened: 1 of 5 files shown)


Dooruk:
I spent quite some time in queue (4hours might be overkill) but buildjedi worked via swell_task so I'm approving.

Dooruk (approved):

jardizzo (approved):
Looks good to me.

PR #147, opened 2023-02-06
add new task for generating static b by linking

danholdaway:
## Description

Adds a new task for generating the static B matrix by linking. This will avoid the need to run the full task when a larger number of  nodes are needed and there is no dependency on the cycle.

### DIFF
--- a/src/swell/tasks/generate_b_climatology_by_linking.py
+++ b/src/swell/tasks/generate_b_climatology_by_linking.py
@@ -46,11 +46,11 @@ class GenerateBClimatologyByLinking(taskBase):
 
         # Source path base the part that looks like /path/to/static_background_error_model/
         source_path_base = os.path.join('jedi', 'interfaces', self.get_model(), 'model',
-                                     'static_background_error', static_background_error_model)
+                                        'static_background_error', static_background_error_model)
 
         # Model specific part of the path, i.e. anything that goes between the base path above,
         # and the actual files that need to be linked into the directory.
-        if  static_background_error_model == 'bump':
+        if static_background_error_model == 'bump':
             source_path_model = self.append_source_path_bump()
         else:
             self.logger.abort("Only bump is currently supported")
--- a/src/swell/tasks/generate_b_climatology_by_linking.py
+++ b/src/swell/tasks/generate_b_climatology_by_linking.py
@@ -1,0 +1,94 @@
+# (C) Copyright 2023 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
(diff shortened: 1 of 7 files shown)


src/swell/utilities/file_system_operations.py
```diff
@@ -0,0 +1,97 @@
...
+    # Check that at least one path was found
+    if not any(found_paths):
+        logger.abort(f'In link_all_files_from_first_in_hierarchy_of_sources none of the ' +
+                     f'directories being searched were found to exist. Directories: {source_paths}')
+
+    # Second sweep to establish if directories contain files
+    print('found_paths', found_paths)
+    found_files = False
+    for ind, source_path in enumerate(source_paths):
+        print("AAA", ind, source_path, found_paths[ind])
```
Dooruk: I'm guessing these print statements were for testing?

src/swell/tasks/generate_b_climatology_by_linking.py
```diff
@@ -0,0 +1,94 @@
...
+            All inputs are extracted from the JEDI experiment file configuration.
+            See the taskBase constructor for more information.
+        """
+
+        # Get the flavor of static background error model
+        static_background_error_model = self.config_get('static_background_error_model')
+
+        # Extract general parts of the config
+        swell_static_files_main = self.config_get('swell_static_files')
+        swell_static_files_user = self.config_get('swell_static_files_user', None)
```
Dooruk, line 41: Is this defined in any of the suites right now?

Dooruk (changes requested):
It works on my end, just a few print statements left.

Also `swell_static_files_user` is not defined in any of the suites.

danholdaway:
I'm thinking that `swell_static_files_user` will be a highly customized thing that folks can manually add to their YAML files.

src/swell/utilities/file_system_operations.py
danholdaway: Oops! Thanks.

src/swell/tasks/generate_b_climatology_by_linking.py
danholdaway: Note that it defaults to `None` if the user does not provide this in their YAML. Generally I don't expect people to be using this much.

Dooruk: I see. I guess my point is that people won't see it if it's not shown in or documentation.

danholdaway: It's a good point. I guess the best thing would be to use R2D2 for these kinds of things.

PR #150, opened 2023-02-08
Add the suite file and task shells for running a forecast of GEOS

danholdaway:
## Description

This adds the empty suite and empty tasks for running a forecast of GEOS. At the moment the suite does not do anything but provides the framework for adding our respective parts.

## Dependencies

N/A

## Impact

N/A


sanAkel (commented):
Thanks @danholdaway.

I need to follow through and help myself with your [doc](https://geos-esm.github.io/swell/). Only after that I think I can be capable of reviewing- though they all look okay to my 👁️.

Dooruk:
It looks like a good starting point. My only comment is that some of the same tasks were created during the following:

https://github.com/GEOS-ESM/swell/pull/133/files

Something to keep in mind while merging 3dvar.

Dooruk (approved):

danholdaway:
> It looks like a good starting point. My only comment is that some of the same tasks were created during the following:
> 
> https://github.com/GEOS-ESM/swell/pull/133/files
> 
> Something to keep in mind while merging 3dvar.

Good point. Might have a few conflicts to resolve.

PR #153, opened 2023-02-09
Remove constraint from the build_jedi suite file

danholdaway:
## Description

Bug fix for error reported by @jardizzo when running the nightly


jardizzo (approved):
So the constraint was to use cascade/skylake? . We are also experiencing long wait-times in the queue. That could be because the haswell nodes are less in number.

PR #155, opened 2023-03-01
Use skylake or cascade lake for the default nodes

danholdaway:
## Description

Batch jobs on Discover still seem to be using Haswell unless specified otherwise. This is leading to longer queue times than necessary. This PR allows jobs to choose either Cascade Lake or Skylake nodes. Eliminating Haswell, since these are being decommissioned. 

Hopefully this will eliminate long queue times on 

## Dependencies

N/A

## Impact

N/A

### DIFF
--- a/src/swell/__init__.py
+++ b/src/swell/__init__.py
@@ -9,4 +9,4 @@ import os
 repo_directory = os.path.dirname(__file__)
 
 # Set the version for swell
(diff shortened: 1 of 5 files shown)


jardizzo (approved):
Hopefully this will speed up the nightly runs also; though queue wait times have generally been less than 1.5 hours.

PR #156, opened 2023-03-10
Add workflow for building GEOS and populate relevant tasks

danholdaway:
## Description

Add tasks for clone, build (by linking) and build GEOS.

Add workflow for building GEOS. First part of having a workflow for making a forecast with GEOS.

### DIFF
--- a/src/swell/utilities/build.py
+++ b/src/swell/utilities/build.py
@@ -26,10 +26,8 @@ def build_and_source_dirs(package_path):
     package_source_path = os.path.join(package_path, 'source')
 
     # Remove trailing slash if needed
-    if package_build_path.endswith('/'):
-        package_build_path = package_build_path[:-1]
-    if package_source_path.endswith('/'):
-        package_source_path = package_source_path[:-1]
+    package_build_path = package_build_path.rstrip('/')
+    package_source_path = package_source_path.rstrip('/')
 
     return package_build_path, package_source_path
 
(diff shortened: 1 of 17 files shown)


danholdaway:
@mathomp4 would appreciate if you have a chance to check over the `clone_geos.py ` and `build_geos.py` tasks to make sure I haven't done something incorrect.

src/swell/utilities/build.py
```diff
@@ -16,22 +16,22 @@
...
-    jedi_bundle_source_path = os.path.join(jedi_bundle_path, 'source')
+    package_build_path = os.path.join(package_path, 'build')
+    package_source_path = os.path.join(package_path, 'source')
 
     # Remove trailing slash if needed
-    if jedi_bundle_build_path.endswith('/'):
-        jedi_bundle_build_path = jedi_bundle_build_path[:-1]
-    if jedi_bundle_source_path.endswith('/'):
-        jedi_bundle_source_path = jedi_bundle_source_path[:-1]
+    if package_build_path.endswith('/'):
```
jardizzo: You might be able to just use rstrip('/') here.

jardizzo (approved):
Looks like lots of subprocess code consolidation in addition to new tasks. I need to make sure that the nightly CI workflow(s) are not coupled with a now outdated modules file since a new module load was added for GEOS. See one minor comment on possible benefit of "rstrip()" to remove trailing characters (can eliminate the endswith check).

Dooruk:
The GEOS build worked for me. Only issue (not related to this PR) I had was the long wait time for the compute node.

Dooruk (approved):
The GEOS build worked for me. Only issue (not related to this PR) I had was the long wait time for the compute node.

src/swell/utilities/build.py
danholdaway: Thanks @jardizzo, always happy to eliminate an `if`!

PR #166, opened 2023-04-12
R2d2v2

ghost:
## Description

THIS IS A DRAFT;  NOT READY FOR REVIEW.

Update SWELL for R2D2 Version 2.  Three components to this update:
- API calls.
- Config restructure (no longer a r2d2 config yaml).
- SWELL must work in conjunction with a MySQL server.  We can install this "locally" (on Discover) or leverage the AWS cloud server for this. 

Some, but not necessarily all, issues:
- Do we need to account for `fc_date_rendering` (e.g. `store_background.py`)?  TBD.
- `window_length` for observations (e.g. `save_obs_diags.py` and `obs_process_setup.py`?
- `model` not specified in old API calls (e.g. `obs_process_setup.py` bias correction), just use 'geos' here? 
- Shall we preregister metadata in the database (e.g. experiments, models, etc.)?  Or should SWELL code register them on the fly? 

Every API call must be successfully tested before considering a merge into develop.

### DIFF
--- a/src/swell/deployment/bin/swell_create_experiment.py
+++ b/src/swell/deployment/bin/swell_create_experiment.py
@@ -83,21 +83,23 @@ def main(config):
     # Create R2D2 database file
     # -------------------------
     data_assimilation_run = dict_get(logger, experiment_dict, 'data_assimilation_run', False)
-    if data_assimilation_run:
-        r2d2_conf_path = os.path.join(exp_suite_path, 'r2d2_config.yaml')
 
-        # Write R2D2_CONFIG to modules
-        with open(os.path.join(exp_suite_path, 'modules'), 'a') as module_file:
-            module_file.write(f'export R2D2_CONFIG={r2d2_conf_path}')
-
-        # Open the r2d2 file to dictionary
-        with open(r2d2_conf_path, 'r') as r2d2_file_open:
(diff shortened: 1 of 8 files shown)


danholdaway:
Thanks for starting these changes @tariqhamzeyssai. I'm on PTO next week so we'll have to discuss the week afterwards. I suspect we can neglect fc_date_rendering. We'll need window_length I think. I understood that R2D2 will eventually prepare a file with all the observations for that window length. Why is 'model' needed when getting observations? I wouldn't imagine that would depend on the model. Swell should be able to add outputs to the database. But we should pre-register all the inputs. At the moment they are all in `/discover/nobackup/drholdaw/R2D2DataStore`. Could we start copying files from there to the database to start testing the new code?

danholdaway:
If you want to run the coding norms locally you can issue `python pycodestyle_run.py` from the root directory of the swell repo.

ghost:
> Why is 'model' needed when getting observations? I wouldn't imagine that would depend on the model.

We're getting bias correction. Is that not model-specific?  

EWOK scripts were specifying `model` for R2D2V**1** calls, as it is for V**2**.  I can't log into Discover at the moment to view your R2D2 build, but I think the old 'bc.yaml' file validated `model` with no default.  It is not immediately clear how SWELL `bc` fetches were passing without `model`.

PR #176, opened 2023-05-10
Fix MarineVertInterp and update some modules to sky4

danholdaway:
## Description

- Keep up with JEDI (https://github.com/JCSDA-internal/soca/pull/874)
- Keep up with Skylab (to version 4.0)

## Dependencies

N/A

## Impact

N/A

### DIFF
--- a/src/swell/deployment/platforms/nccs_discover/modules
+++ b/src/swell/deployment/platforms/nccs_discover/modules
@@ -17,14 +17,15 @@ umask 022
 module use /discover/swdev/jcsda/spack-stack/modulefiles
 module load miniconda/3.9.7
 module load ecflow/5.8.4
-module use /gpfsm/dswdev/jcsda/spack-stack/spack-stack-v1/envs/skylab-3.0.0-intel-2022.0.1/install/modulefiles/Core
+module load mysql/8.0.31
+module use /gpfsm/dswdev/jcsda/spack-stack/spack-stack-1.3.0/envs/unified-env/install/modulefiles/Core
 module load stack-intel/2022.0.1
 module load stack-intel-oneapi-mpi/2021.5.0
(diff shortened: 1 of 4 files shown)


jardizzo (approved):

src/swell/deployment/platforms/nccs_discover/modules
```diff
@@ -42,7 +43,7 @@ module use -a /discover/nobackup/drholdaw/JediOpt/modulefiles/core
 module load solo/swell-1.0.0
 module load r2d2/swell-1.0.5
 module load eva/1.3.4
```
Dooruk: Eva is giving the following error:

`ABORT EvaFactory: Expecting to find a class called 'FigureDriver' in a file called '/discover/nobackup/dardag/JediOpt/src/eva/latest/src/eva/plot_tools/figure_driver.py' but no such file was found or an error occurred during import. `

`Reported error: No module named 'scipy.stats._hypotests_pythran'`

Same issue with EVA 1.3.5 as well.

Dooruk (changes requested):
Hofx and 3dvar worked on my end. I just had to make a new build with jedi_bundle 1.0.10. I will leave the Cylc test to you since you are already in queue.

src/swell/deployment/platforms/nccs_discover/modules
danholdaway: Thanks @Dooruk. I reinstalled `eva/1.3.5` using skylab 4 modules and that should hopefully fix that issue.

danholdaway: ...and also updated this branch to point to that version of eva.

danholdaway:
Since the nightly is already broken I'm going to merge this so we can see if it works tonight.

PR #178, opened 2023-05-22
Allow only explicit use of configuration

danholdaway:
## Description

As I see it there are two aspects of the config that make it harder to understand than we would like:

1. Things are dynamically added to the configuration.
2. The config it used to template certain configuration file in a way that is invisible to users.

These two factors make it hard to accomplish the next major feature of attaching the questions to the tasks. We need a way of cleaning saying this question belongs to these tasks. We don't want to have questions where it is not clear which tasks are using that part of the config. The first step to doing this is to remove the ability to add things to the config, which has a number of knock-on effects. The second is to force the creation of a specific dictionary that is used in templating other files, with each task that uses that feature being forced to reference the configuration and copy the variable to the templating dictionary. 

In doing this I created two new classes, one for obtaining data assimilation window variable and another for templating JEDI config files. These are variables of the base class with delegation methods added to the base class. I cleaned up the base directory of the tasks leaving only the actual base classes. 

This work will enable the next step of moving all the questions from the suites part of the code to the tasks part of the code.

Sorry for the large number of changes. I opted to unify the date in the headers of the files such that they don't change in the future.


danholdaway:
@jardizzo can you trigger the nightly for this?

Dooruk (approved):
I made two quick fixes. hofx and 3dvar runs on my end. I will adapt to these changes with the new suites.

I didn't test build_geos or build_jedi.

PR #182, opened 2023-06-02
Refactor the creation of config variables ahead of question implementation

danholdaway:
## Description

The introduces a new way of managing the configuration. Currently each suite re-writes all the questions that the suite needs and this is of course not extensible. This PR keeps the old infrastructure in place but starts to add the new infrastructure for storing the question infrastructure. The heart of this is `tasks/task_questions.yaml`. The parts of the config that a task is allowed to access is now controlled by `task_questions.yaml`. A new test is added to ensure that future PRs do not allow the YAML file and what the tasks actually need gets out of sync.


Dooruk (approved):
Hofx and 3dvar suites work on my end. I started updating tasks with cycling and I find it straightforward to adapt to these questionnaire changes with the incoming cycling/forecast_geos tasks. It is much more clear now.

My one comment would be the `execution_time_limit`s in `prep_suite.py` causing long queue times if requested time is too high.

PR #183, opened 2023-06-07
Use spack-stack 1.4 modules to maintain compatibility with JEDI

danholdaway:
## Description

JEDI was updated to require spack-stack 1.4. THe nightly will fail tonight without an update to the modules.


danholdaway:
https://github.com/GEOS-ESM/swell/actions/runs/5202807938

jardizzo (approved):

PR #184, opened 2023-06-09
Geos cycling tasks

Dooruk:
# Description

Major refactoring of GEOSgcm `gcm_run.j` within Swell and the addition of relevant tasks and utilities. Includes recent updates regarding configurations and task questions. 

New suite `forecast_geos` allows running coupled GEOSgcm with 5 deg ocean horizontal resolution and `PE12x72-CF` atm. resolution. 0.66 and 0.25 degree tests were also simulated but they will require a few more modifications to run. 

New suite `3dvar_cycle` is capable of running coupled GEOSgcm with 5 deg ocean horizontal resolution and PE12x72-CF atm. resolution with 24 hour DA window. 2/3- and 1/4-degrees runs are also possible. Appropriate IC files need to be provided and the rest is handled by `3dvar_cycle`

These changes allow hot starting 3DVar cycling GEOS from a previous run using configuration files created with gcm_setup. Swell assumes `$HOMDIR` and `$EXPDIR` are collocated (if not, only the `RC` folder needs to be moved from `$EXPDIR` to the `$HOMDIR`).

User needs to point to the `$HOMDIR` of the previous run (for hotstarting) as well as the `*_rst` files, `MOM.res.nc` (can be multiple), and `tile.bin` created by a previous GEOS simulation. Rest is handled by Swell tasks internally.

## Capabilities
- Parsing `AGCM.rc`, `CAP.rc`, and `gcm_run.j` into dictionaries to obtain grid, BC, and physics information. This allows modifications of BCs according to cycle time step similar to what's done in `gcm_run.j`.
- Using dictionary or string replacement for modifying `NUM_SGMT` and `JOB_SGMT` entries in `CAP.rc` and `'n'` to `'r'` change in `input.nml` .
- Creating `cap_restart` automatically.
- Renaming `*_checkpoint` to `*_rst`.
- Can handle optional restart output option in `AGCM.rc` called `RECORD_FREQUENCY` which will be useful for DA outputs.
- Obtaing next cycle dir information using forecast duration.
- Works with Skylab v4.
- Works with GEOSgcm Krok
- Time information for GEOS is obtained from the `*_rst` files. This could pose an issue between yaml setup and restart files so probably requires some time difference check in Swell. GEOSgcm handles this internally and quits if mismatch is too high.
- Updating `MOM.res.nc`  with the `ocn.analysis` file from SOCA with `PrepareAnalysis` task. This task could easily be updated to handle `ice`, `sfc`, and `bgc` files using SOCA `nomenclature`. 
- GEOS model files are now confined in `geosdir`.

## Some current issues
- Only 6 x 1 processor layout is supported for now as that is hard coded in the `AGCM.rc` and `MOM_override` which is obtained by gcm_setup.
- `RunGeosExecutable` does not allow repeating runs in the same folder if the model progressed some time steps during the first try and outputted history files. GEOS fails and exits if it sees those history files during next try.
- MPI arguments from `gcm_run.j` were not included. mpirun automatically assumes MOM6 (`LD_PRELOAD=$GEOSDIR/lib/libmom6.so` was not necessary). These did not cause issues on my end but something to keep in mind.
- RPLY works for 0.25 degree tests but only works for certain cycle start times, more relevant for cycling DA considerations.
- `-P1D` in `flow.cylc` requires manual change for geos suites, `-P1` does not work in UTC mode.
- `cycle_times` need to be added to `flow.cylc` for `3dvar_cycle` suite.
- DA cycling works with restarts, gives NaN error with history files. 
- Storing restart and obs files outside of run directories.


src/swell/suites/3dvar_cycle/suites-3dvar_cycle.yaml
```diff
@@ -0,0 +1,128 @@
...
+  type: iso-datetime
+
+# Each forecast cycle consists of a set duration until DA
+# Should be less than or equal to 1 day due to the ISO -> string conversion
+forecast_duration:
+  default_value: 'P1D'
+  prompt: What is the GEOS forecast duration?
+  type: string
+
+# Run ahead limit for workflow
```
danholdaway, line 20: You can remove the run ahead limit. Since it is cycling cycles cannot move ahead since one cycle depends on the previous one.

src/swell/tasks/base/task_base.py
```diff
@@ -159,6 +170,17 @@ def cycle_dir(self):
 
     # ----------------------------------------------------------------------------------------------
 
+    def forecast_dir(self):
+
+        # Combine datetime string (directory format) with the model
+        forecast_dir = os.path.join(self.experiment_path(), 'run',
```
danholdaway, line 177: Should the forecast directory be something like `experiment_path/run/datetime/forecast`? I think the way it is would see a piling up of all those files in the datetime directory and then the data assimilation directories would be a little hidden in the mess.

src/swell/tasks/get_restart.py
```diff
@@ -7,18 +7,67 @@
...
+        # TODO: Users might choose to point to a random local directory, where
+        # to keep restart files?
+        # -------------------------------------------------------------------
+        self.logger.info('GEOS restarts are copied from a previous forecast')
+
+        src = os.path.join(self.swell_static_files, 'geos', 'restarts', rst_path, '*_rst')
+
+        for filepath in list(glob.glob(src)):
+            filename = os.path.basename(filepath)
+            self.geos.copy_to_geosdir(filepath, self.geos.at_cycle_geosdir(filename))
```
danholdaway: Do you need the helper `copy_to_geosdir` function for this? Can it just be `shutil.copy2(filepath, os.path.join(self.forecast_dir(), filename)`? If you want to do some checking around copying perhaps it could go into a more generically named function that lives in `utilities/file_system_operations.py`?

src/swell/tasks/get_restart.py
```diff
@@ -7,18 +7,67 @@
...
+
+        src = os.path.join(self.swell_static_files, 'geos', 'restarts', rst_path, 'tile.bin')
+        self.geos.copy_to_geosdir(src, self.geos.at_cycle_geosdir('tile.bin'))
```
danholdaway: Same as above.

src/swell/tasks/get_restart.py
```diff
@@ -7,18 +7,67 @@
 
 # --------------------------------------------------------------------------------------------------
 
+import shutil
```
danholdaway, line 10: Would it make more sense to name this task `get_geos_restarts` since it's specific to GEOS?

src/swell/tasks/get_restart.py
```diff
@@ -7,18 +7,67 @@
...
     def execute(self):
 
-        self.logger.info('GetRestart')
+        self.logger.info('Obtaining GEOS restarts for the coupled simulation')
+
+        self.swell_static_files = self.config.swell_static_files()
+
+        # Create cycle_dir and INPUT
+        # ----------------------------
+        if not os.path.exists(self.geos.at_cycle_geosdir('INPUT')):
```
danholdaway: Perhaps you can eliminate the need for this `at_cycle_geosdir` method if you change the base class such that it returns the full forecast directory. Then commands like this can just be `os.path.join(self.forecast_dir(), 'INPUT')`. It saves having to go look in another file to see what is happening.

src/swell/tasks/link_geos_output.py
```diff
@@ -0,0 +1,85 @@
...
+        type (history vs. restart), DA method, and window length.
+
+        Options could be offered in yamls, for instance history vs. restart.
+        """
+
+        self.current_cycle = os.path.basename(self.forecast_dir())
+        src, dst = self.link_restart()
+
+        if os.path.exists(src):
+            self.geos.linker(src, dst, dst_dir=self.cycle_dir())
```
danholdaway: Linker also feels like something quite generic that could live in utilities rather than the geos class.

src/swell/tasks/prepare_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+import shutil
+
+from swell.tasks.base.task_base import taskBase
+from swell.utilities.datetime import datetime_formats
+
+# --------------------------------------------------------------------------------------------------
+
+# TODO: This could be restructured according to model type (MOM6, CICE etc.)
+# --------------------------------------------------------------------------
+SOCA_dict = {
```
danholdaway: Can you put this into the `jedi/interface/geos_ocean/geos_ocean.yaml` metadata?

src/swell/tasks/prepare_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+
+        """
+        Updates variables in restart files with analysis variables.
+        """
+
+        self.logger.info('Preparing analysis and updating restarts')
+
+        # Current and restart time objects
+        # --------------------------------
+        self.current_cycle = os.path.basename(self.forecast_dir())
```
danholdaway, line 42: This might need updating with the base class forecast_dir suggestion.

src/swell/tasks/prepare_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+        self.cc_dto = dt.strptime(self.current_cycle, datetime_formats['directory_format'])
```
danholdaway: Can you use `self.cycle_time_dto()` here?

src/swell/tasks/prepare_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+        for soca_var in self.soca_ana:
+            var = SOCA_dict[soca_var]
+            self.logger.info(f'Updating {var} in restart')
+
+            ds_rst.variables[var][:] = ds_ana.variables[var][:]
+            ds_rst.sync()
+        ds_ana.close()
+        ds_rst.close()
+
+        shutil.move(f_rst, self.geos.at_cycle_geosdir(['RESTART', 'MOM.res.nc']))
```
danholdaway: After the above changes you might be able to use `(self.forecast.dir(), 'RESTART', 'MOM.res.nc')`.

src/swell/tasks/run_geos_executable.py
```diff
@@ -7,18 +7,47 @@
...
 
 class RunGeosExecutable(taskBase):
 
     def execute(self):
 
-        self.logger.info('RunGeosExecutable')
-
+        # Total number of processors
+        # ----------------------------
+        np = self.config.total_processors()
```
danholdaway: I wonder if this key should be forecast processors? I think we have total_processors somewhere else that probably also shouldn't be called total_processors. But we could change that one later on. The reason I mention this is that in the future we might run the forecast with x processors then the atmospheric da with y processors and ocean da with x-y processors.

src/swell/tasks/run_geos_executable.py
```diff
@@ -7,18 +7,47 @@
...
+        # ---------------
+        geos_modules = 'g5_modules.sh'
+        geos_modules_path = os.path.join(self.experiment_path(), 'GEOSgcm', 'source',
+                                         '@env', geos_modules)
+
+        # Run the GEOS executable
+        # -----------------------
+        self.logger.info('Running '+geos_executable_path+' with '+str(np)+' processors.')
+
+        self.geos.run_executable(self.geos.at_cycle_geosdir(), np, geos_executable_path,
```
danholdaway: Do we need this method? I think it's only called from here and I don't anticipate there would be other tasks needing to do this. Perhaps the code in the method could just be in the task?

src/swell/utilities/bin/generate_task_question_default_dicts.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python
```
danholdaway: Can you remove this file? I think it came in by mistake during merging.

src/swell/utilities/bin/generate_task_questions_dict.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python
```
danholdaway: Can you remove this file? I think it came in by mistake during merging.

src/swell/utilities/geos.py
```diff
@@ -0,0 +1,436 @@
+# (C) Copyright 2023 United States Government as represented by the Administrator of the
```
danholdaway, line 1: I have a feeling that this class could be reduced a bit. As a rule of thumb if the method is pretty generic it could go into one of the utilities, for example file operations. If the method is only called in one place and we don't anticipate it being called from other places then the method could be moved there. This is especially true for methods that are only called by the task that creates the GEOS run directory. Eventually that task will hopefully be removed completely when GEOS can provide that functionality in a callable way.

danholdaway (commented):
A few comments after looking through the code. I haven't tried to run anything yet.

danholdaway:
Thanks for putting together this great and extensive piece of work. I haven't tested things yet but started adding a few minor comments.

src/swell/utilities/geos.py
```diff
@@ -0,0 +1,436 @@
...
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+
+# --------------------------------------------------------------------------------------------------
+
+from datetime import datetime
+import f90nml
```
danholdaway, line 11: This probably ins't a standard pacakge so should be added to the setup.py and requirements*.txt files.

Dooruk: I see this in the loaded swell modules: `py-f90nml/1.4.3`. Is that irrelevant?

src/swell/utilities/geos.py
Dooruk: I agree and thanks for bringing this up. I refrain from making changes to the tasks in the `utilities` as it might break some unrelated task that I'm not aware of. Some of the existing `utilities` file operations did not work properly with symlinks regarding `RC` files. So I just created link and copy methods that work for GEOS tasks but I agree that we could combine them with `utilities` ones.

src/swell/utilities/bin/generate_task_questions_dict.py
Dooruk: done

src/swell/utilities/bin/generate_task_question_default_dicts.py
Dooruk: done

src/swell/tasks/run_geos_executable.py
Dooruk: fixed

src/swell/tasks/link_geos_output.py
Dooruk: I agree, please see my `utilities` reply below.

src/swell/tasks/get_restart.py
Dooruk: Agreed & changed.

src/swell/suites/3dvar_cycle/suites-3dvar_cycle.yaml
Dooruk: Ok, is that an optional entry for cylc?

src/swell/tasks/run_geos_executable.py
Dooruk: Another alternative is to use the .rc parsing method to obtain these as dictionary items from `AGCM.rc`. `NX` and `NY` for atmosphere and `OGCM.NX` and `OGCM.NY` for ocean model processor layouts/numbers. One has to check consistency between the cycling workflow and GEOS input files anyway.

src/swell/utilities/geos.py
danholdaway: That's why it works. But it's necessary to be explicit which non-standard dependencies are needed.

src/swell/tasks/run_geos_executable.py
danholdaway: That would probably make sense. Presumably later on we would remove this key from Swell anyway since it should be a question asked as part of the GEOS configuration.

src/swell/suites/3dvar_cycle/suites-3dvar_cycle.yaml
danholdaway: Yeah it's optional for when there is no inter-cycle dependency.

src/swell/tasks/prepare_analysis.py
Dooruk: Yes, fixed

src/swell/tasks/base/task_base.py
Dooruk: The way I organized it, GEOS scratch directory resides in `experiment_path/run/datetime/geosdir`. That way `RemoveGeosRunDir` can rm that folder (once the `MoveRestart` and `SaveRestart` is done for the cycle) and DA folders are confined in `experiment_path/run/datetime/model` i.e., `cycle_dir`.

I like your suggestion swapping `geosdir` with `forecastdir` and have that in base class though, let me give that a try.

src/swell/tasks/get_restart.py
Dooruk: Yeah, I agree. This made sense in terms of code brevity while I was using inheritance lavishly. I will make some modifications.

src/swell/utilities/geos.py
Dooruk: I merged the utilities except linker.

danholdaway (approved):
Swell cycles!

PR #189, opened 2023-07-07
Feature/questions to tasks callum

CRWayman:
## Description

Large update to prep config python codes in line with new task questions yaml structure.

## Dependencies

## Impact

- [ ] repo
- [ ] ...


danholdaway:
Thanks again for putting this all together @CRWayman. It's looking really nice. In trying to run some things I realized we (I) missed something fairly fundamental, which is that we will in fact need model based 'suite questions'. Primarily this the cycle_times directive. This wouldn't be information that is needed by a particular task but does depend on the model at hand. What I've done is [push to your branch](https://github.com/GEOS-ESM/swell/pull/189/commits/e69b34ea4deed96a72c9930d294831034944ec16) a `suite_questions.yaml` to each of `geos_atmosphere` and `geos_ocean` and put in there the default and options for cycle_times. I've also added that question to the `suites/hofx/suite_questions.yaml` (note that I renamed suites_new to suites and suites to suites_old. I think your code will need to essentially accumulate any questions that occur in the `suite/hofx/suite_questions.yaml` and ask them later on during the model loop. With a quick glance of your code I couldn't see how to do that easily but perhaps you can see a way?

CRWayman:
Hey Dan, I noticed an issue with the build_jedi task questions: "Do you want to use an existing JEDI build or create a new build?" should probably be a string-drop-list so users don't have to guess the syntax for "create" instead of use_existing

CRWayman:
Ok I've fixed the experiment config output for suite_to_run and platform by adding them to my dictionary in the "suite_setup" function in prep_config_base.py

I also removed unnecessary print statements, and removed the "Please answer the following questions..." text when defaults is called, this question is still printed when questionary is used.

danholdaway:
> Hey Dan, I noticed an issue with the build_jedi task questions: "Do you want to use an existing JEDI build or create a new build?" should probably be a string-drop-list so users don't have to guess the syntax for "create" instead of use_existing

Oh yeah. That makes sense. It should be a drop list.

danholdaway:
Thanks for the latest updates.

danholdaway (approved):

danholdaway:
Thanks for this substantial piece of work @CRWayman. The changes are now merged into develop. Cheers to the removal of a significant amount of duplicate YAML and a clearer experiment generation system for everyone.

PR #192, opened 2023-07-17
Rename Discover workflows

danholdaway:
## Description

Rename the workflow files


jardizzo (commented):
Is the intent to have one tiered file (1,2,3) and launch all workflows under that tier from one dispatch file? It looks like the nightly workflow is under the tier-1 dispatch. Will we have another cron workflow that also uses it?

jardizzo (approved):

PR #217, opened 2023-08-24
UFO tests - Scatwind Yaml Operator Change

asewnath:
## Description

In order for Scatwind to agree with the new x48 data (with Jianjun's pressure changes), the product operator needs to be used instead of vertical interpolation. Note that vertical interpolation was used for the original data without the pressure changes.


danholdaway:
@asewnath I'm going to close this PR as I think it's not required. And something went very wrong with the history along this branch. Can you check the scatwind file that's in develop and confirm that it looks as you would expect?

danholdaway:
If you see a problem please open a new PR against the current develop so you have the correct history.

PR #226, opened 2023-09-07
Surface obs and bump time by one second

danholdaway:
## Description

- Turn on all the surface observations
- Bump the time by one second if equal to the beginning of the window in GSI to IODA converter
- Adjust all the passedBenchmarks that change by adding a second


gmao-jjin3 (approved):

danholdaway:
Tier 1 passing https://github.com/GEOS-ESM/swell/actions/runs/6116396789

PR #233, opened 2023-09-20
No bias correction for atms channel 15 in GEOS.

gmao-jjin3:
No bias correction for atms channel 15 in GEOS.


danholdaway (approved):

danholdaway:
Thanks @gmao-jjin3. Looks good to me but will hold off on merging until the tests are fixed.

PR #235, opened 2023-09-21
Renaming test suite files and adding overrides to prep config softwar…

CRWayman:
…e. also advanced options added

## Description
- advanced options flag lets users view every question, even those which are normally hidden from users
- renaming tests from stable suffix to tier1 suffix
- adding override capability from designated file
- fixing a bug so that when test suites are called during cli, options for check lists become equal to default option provided in test file

## Dependencies

## Impact

- [ ] repo
- [ ] ...


danholdaway (approved):
Thank you @CRWayman

PR #247, opened 2023-09-27
Remove passive_bc option

danholdaway:
## Description

Turn off everywhere we have 

```yaml
use passive_bc: true
```
 Either commented or not. The deafult is false so removing sets it to false. This will probably change counts so @gmao-jjin3 @gmao-wgu are we OK with this in principle before we start testing it all out?

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/airs_aqua.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/airs_aqua.yaml
@@ -362,7 +362,6 @@ obs filters:
     channels: *airs_aqua_channels
     options:
       channels: *airs_aqua_channels
-#        use passive_bc: true
       use_flag: *useflag_airs_aqua
   minvalue: 1.0e-12
   action:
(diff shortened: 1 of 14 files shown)


gmao-jjin3 (approved):

gmao-jjin3:
> Either commented or not. The deafult is false so removing sets it to false. This will probably change counts so @gmao-jjin3 @gmao-wgu are we OK with this in principle before we start testing it all out?

I don't think it changes counts to remove it.   Counts were calculated without this passive BC configuration which is a flawed design because it simply turns on all these passive channels.

gmao-yzhu:
I talked this with Ben Ruston last week, he took notes and will check with people in JCSDA. But this won't affect what we are doing here at this moment.

gmao-wgu:
I observed use passive_bc also be used in the inter-channel consistency check for MW, it will not change the count of active channels, so it should be still set as .true. @

danholdaway:
Thanks @gmao-wgu. I've updated the PR. The only thing really happening now is to turn it off for ATMS in the use flag filter.

gmao-jjin3:
@gmao-wgu is right about which one should be "true".  The benchmark values came from my tests in which both were true, so they should be changed.

danholdaway:
@gmao-jjin3 @gmao-wgu these changes require a change in count for ATMS. Are we OK with this?

gmao-jjin3:
@danholdaway I am OK with that.  I can get new counts if you don't have them already.

gmao-jjin3:
@danholdaway  See the numbers in https://github.com/GEOS-ESM/swell/pull/249

danholdaway:
> @danholdaway See the numbers in #249

Those are the same counts we have here so I'll go ahead and merge. Thank you.

PR #252, opened 2023-10-04
Single entry point for all applications and utilities

danholdaway:
## Description

Instead of having many scripts we have a single entry point `swell`.

Examples of new use:

`swell create ufo_testing`  will create the experiment directory. The old `swell_prepare_experiment_config` and `swell_create_experiment` wrapped into one command.

`swell launch /path/to/yaml` instead of `swell_launch_experiment`

`swell task GetObservations ... ` instead of `swell_task`.

Note that now you can do `swell --help` and see all features of the entire package. Then you can do things like `swell create --help` to go down into each application to see the help. So the user has only one entry point and can figure out the entire application from there.

h/t @aerorahul for this nice suggestion.

## Dependencies

- https://github.com/GEOS-ESM/CI-workflows/pull/7

## Impact

N/A


danholdaway:
I still need to make the corresponding changes to the tier1 and tier2 testing.

Dooruk:
I see a few module changes but is this still using spack 1.4 modules?

danholdaway:
Yes we are still using spack-stack 1.4. The package changes are just for GitHub. Now everything goes through the top level script, which includes an import of everything in one way or another. Previously calling a test meant only needing to import what that test needed. We could have probably avoided that with a less monolithic top level but I figured this makes it easier for folks to look at and understand what's going on in the future.

src/swell/deployment/platforms/generic/suite_questions.yaml
```diff
@@ -2,7 +2,7 @@ experiment_id:
   default_value: swell-{{suite_to_run}}
 
 experiment_root:
-  default_value: /home/${USER}/SwellExperiments
+  default_value: $HOME/SwellExperiments
 
 r2d2_local_path:
   default_value: /home/${USER}/R2D2DataStore/Local
```
CEgerer93: For consistency with `experiment_root: default_value` and a potentially changing /home filesystem: 
```suggestion
  default_value: $HOME/R2D2DataStore/Local
```

CEgerer93 (commented):
I quite like the single entry point implementation. Indeed, it makes understanding the application easier. Looking forward to working with these changes when merged.

src/swell/deployment/platforms/generic/suite_questions.yaml
danholdaway: Thanks for the suggestion.

src/swell/deployment/prep_config.py
```diff
@@ -1,119 +0,0 @@
...
-
-    # Dictionary file to write
-    # ------------------------
-    cwd = os.getcwd()
-    experiment_id = dict_get(logger, experiment_dict, 'experiment_id')
-    exp_dict_file = os.path.join(cwd, f'{experiment_id}.yaml')
-
-    # Write dictionary to YAML file
-    # -----------------------------
-    with open(exp_dict_file, 'w') as file:
```
Dooruk, line 110: `swell_prepare_experiment_config` used to create a local `${experiment_id}.yaml` file, outside the experiment directory. With the new combined approach this is not happening, was this intentional?

Dooruk:
This is great, `--help` outputs are much more clear and helpful now!

src/swell/deployment/prep_config.py
danholdaway: Good spot and thanks for flagging. That is intentional yes. One thing we will want to avoid in the future is folks being too loose about their experiment IDs. Since experiments will be tied to databases, either R2D2 or GMAO, experiment IDs will need to be unique. By creating the entire experiment in one go it becomes clearer to users not to just be changing things in the YAML that shouldn't be changed.

PR #255, opened 2023-10-05
Bring YAMLs up to date with new eva and parallelize the EvaObservation task

danholdaway:
## Description

- Bring swell into compliance with a more modern Eva 1.3.5 to the upcoming 1.6.0 version. This makes for a number of YAML changes in the eva Yamls.
- Parallelize the EvaObservations task, reducing the cost of producing plots for the atmospheric runs by a factor of 10. _But bringing in the cost of being in the queue._ 

This would be merged once the Eva PR is merged, incase of any requested changes there. However the module for eva is built on Discover so this can be tested now.

## Dependencies

- [ ] https://github.com/JCSDA-internal/eva/pull/158

## Impact

N/A

### DIFF
--- a/src/swell/tasks/eva_observations.py
+++ b/src/swell/tasks/eva_observations.py
@@ -141,5 +141,5 @@ class EvaObservations(taskBase):
 
         # Call eva in parallel
         # --------------------
-        with Pool(processes=20) as pool:
+        with Pool(processes=40) as pool:
             pool.map(run_eva, eva_dicts)
--- a/src/swell/tasks/eva_observations.py
+++ b/src/swell/tasks/eva_observations.py
@@ -8,15 +8,25 @@
 # --------------------------------------------------------------------------------------------------
 
 
+from multiprocessing import Pool
 import os
 import yaml
 
-from eva.eva_base import eva
+from eva.eva_driver import eva
 
 from swell.tasks.base.task_base import taskBase
 from swell.utilities.dictionary import remove_matching_keys, replace_string_in_dictionary
 from swell.utilities.jinja2 import template_string_jinja2
 from swell.utilities.observations import ioda_name_to_long_name
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+# Pass through to avoid confusion with optional logger argument inside eva
+def run_eva(eva_dict):
+    eva(eva_dict)
+
 
 # --------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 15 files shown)


asewnath (approved):
Was able to run and everything looks good

Dooruk:
I tested it with my CLI and it works for some of my previous cycling outputs. To test the parallel aspect, I can run it with cylc as well but currently my EvaObs task takes less than a minute for all of my ocean obs. So this might be detrimental for my ocean DA cycle compute times with a 6h window 😃

Would it be possible to make parallelization optional or is that too much to ask? Otherwise, I can just erase the eva job submission part from my flow.cylc.

danholdaway:
What you have to watch out for now @Dooruk is that the task will use up to 40 cores. So if you run it on a login node it will use 40 cores of the login node. It's not like an MPI run where it just wont run on a login node.

danholdaway:
I'm going to reach out to NCCS to see if there is a convenient way know if on the login node or not. Then we can set the number of workers accordingly. Say 4 for a login node and 40 for a slurm node.

Dooruk:
Ok, I thought running on CLI does not initiate parallelization.

danholdaway:
When you run on CLI it will still use the same number of cores that it would use in Cylc. It just doesn't first put you onto a compute node. If it's a task like RunJediVariationalExecutable then it wont run on login because NCCS put a fence in place. But for Python threading no such fence exists.

src/swell/tasks/eva_observations.py
```diff
@@ -130,6 +135,11 @@ def execute(self):
...
-            # Call eva
-            # --------
-            eva(eva_dict)
+            # Add eva dictionary to list
+            # --------------------------
+            eva_dicts.append(eva_dict)
+
+        # Call eva in parallel
+        # --------------------
+        with Pool(processes=40) as pool:
```
Dooruk, line 144: One tiny comment, maybe we could have dictionaries according to cas | sky (using 40 cores) vs. SCU 17 (more than 40 cores) instead of hard coding here. `prep_suite` could also benefit from that but not imperative.

Dooruk (approved):
This is great, it will save a lot of time!

It works on compute node for me, it takes 25 seconds instead of 50 for ocean obs😄

src/swell/tasks/eva_observations.py
danholdaway: When it comes to scu17 we'll have to rethink a lot of things I feel. Note that this isn't like MPI tasks where you try to make strategic choices to balance memory vs compute with available resources. This is more like threading. In fact in some cases you might choose a number here that is greater than the number of physical cores (if the work being done by the workers was heavily IO bound). In that case you might have multiple workers (threads) working on the same core, i.e. one working while another it waiting on IO. In this case 40 is a safe choice pretty much regardless of the number of cores. It would even work just fine on the 24 cores of haswell if we had them.

PR #256, opened 2023-10-05
LETKF for single cycle

CEgerer93:
## Description

Completes support of LETKF in `geos_atmosphere` for a single cycle.
Need to establish conventions for updated ensemble members before cycling in earnest.

## Dependencies

No dependencies, though the following needs to be addressed for an error-free `Cylc` workflow described by `suites/letkf/flow.cylc`:
- [x] `EvaObservations`

## Impact

None.


danholdaway:
Thanks so much for adding this @CEgerer93. You can evaluate the Coding Norm and Code Testing failures on Discover. To do that you would install as follows:

Swell Code Testing:
``` shell
module use -a /discover/nobackup/drholdaw/opt/modulefiles/core/
module load miniconda/py39_23.3.1
pip install --prefix=/path/to/core/swell/version --no-cache-dir -r requirements-standalone.txt .
module load swell/version
swell_tests_code
```

Python Coding Norms
``` shell
cd /path/to/srd/swell/version
python pycodestyle_run.py
```

The YAML test can't be run on Discover so you have to look at the logs here. It looks like it doesn't like Jinja2 templates in the YAML but what you have is legit. Maybe we should ask @mathomp4 if he knows how to skip a file in the YAML linting.

CEgerer93:
Thank you @danholdaway for the reminder on how to locally check the swell/python norms; minor changes, as expected.

To be clear, `swell_tests_code` isn't supported in a standard `swell` build? There are quite a few additional modules (e.g., `flake8`, etc.) required when installing for the swell code testing.

danholdaway:
@CEgerer93 that is correct. The problem is that we normally build swell with the spack-stack modules so swell is installed in a manner consistent with how things it will run (i.e. JEDI) are built. However I found that flake8, which the test needs, requires some versions of dependencies newer than what's in spack-stack. So until we can add more to the spack-stack modules we have two installs.

CEgerer93:
Okay, great - that makes sense. I will await suggestions from Matt concerning YAMLlint.

mathomp4:
I guess I'm a bit late to this, but it looks like you should be able to ignore a file with action-yamllint. Per https://github.com/ibiqlik/action-yamllint/issues/8, you might be able to add to `.yamllint.yml`:
```yaml
    allow-non-breakable-inline-mappings: true
  truthy: disable

  ignore: |
    src/swell/configuration/jedi/oops/LocalEnsembleDA.yaml
```

There actually seems to be a lot of different ways per https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths

If you anticipate more Jinja2/YAML files, might be best to start a `.yamlignore` file and use `ignore-from-file` perhaps?

CEgerer93:
Thank you, @mathomp4 for your suggestions. @danholdaway has implemented a change to allow the yaml norms to pass.

I trust we will be sticking with just a single file with jinja2 conditional formatting. The ensemble-based schemes I am working to add into swell are the only components, as far as I can tell, that feature duplicate yaml entries based on what outputs are produced. A later version of [`fv3-jedi`](https://github.com/JCSDA-internal/fv3-jedi) will likely clear this up.

danholdaway:
Thanks @mathomp4 that suggestion works and is better than what I was doing since it doesn't diverge from the community maintained action for YAML linting.

src/swell/configuration/jedi/interfaces/geos_atmosphere/geos_atmosphere.yaml
```diff
@@ -6,3 +6,4 @@ executables:
   variational3D: fv3jedi_var.x
   variational4D: fv3jedi_var.x
   variational4DEnsVar: fv3jedi_var.x
+  letkf: fv3jedi_letkf.x
```
danholdaway: Should we call this `localensembleda: fv3jedi_letkf.x`? That would be consistent with the oops YAML name and what I think the application is actually called in JEDI.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
```diff
@@ -66,6 +79,51 @@ jedi_forecast_model:
   options:
   - pseudo-model
 
+local_ensemble_inflation_mult:
+  default_value: 1.1
+  options:
```
danholdaway: Do we really want options here? Or should the user be free to pick any float they like?

danholdaway: Same comment for the two below.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
```diff
@@ -66,6 +79,51 @@ jedi_forecast_model:
...
+  - 0.6
+
+local_ensemble_inflation_rtps:
+  default_value: 0.5
+  options:
+  - 0.5
+
+local_ensemble_save_posterior_ensemble:
+  default_value: true
+  options:
```
danholdaway: For boolean questions we don't need to defer the options to each model since they will always be true/false. Can you set the options in the main `task_questions.yaml` instead? Same for next three questions.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
```diff
@@ -174,6 +232,35 @@ produce_geovals:
   - true
   - false
 
+vertical_localization_apply_log_transform:
+  default_value: true
+  options:
+  - true
```
danholdaway: Same as above.

src/swell/deployment/prep_suite.py
```diff
@@ -126,6 +126,7 @@ def prepare_cylc_suite_jinja2(logger, swell_suite_path, exp_suite_path, experime
         'RunJediHofxExecutable',
         'RunJediLetkfExecutable',
         'RunJediVariationalExecutable',
+        'RunJediLetkfExecutable',
```
danholdaway: Could this be `RunJediLocalEnsembleDaExecutable`? For future proofing if this suite could be used for other things, like EnKF, GETKF etc.

src/swell/suites/letkf/flow.cylc
```diff
@@ -6,7 +6,7 @@
 
 # --------------------------------------------------------------------------------------------------
 
-# Cylc suite for executing JEDI-based h(x)
+# Cylc suite for executing JEDI-based LETKF Algorithm
```
danholdaway:
```suggestion
# Cylc suite for executing JEDI-based LocalEnsembleDA Algorithm
```

src/swell/tasks/run_jedi_letkf_executable.py
```diff
@@ -69,6 +69,51 @@ def execute(self):
         self.jedi_rendering.add_key('crtm_coeff_dir', self.config.crtm_coeff_dir(None))
         self.jedi_rendering.add_key('window_begin', window_begin)
 
+        # Ensemble Localizations
```
danholdaway, line 72: As above I would suggest renaming the task to `run_jedi_local_ensemble_da_executable`.

src/swell/tasks/task_questions.yaml
```diff
@@ -370,6 +404,94 @@ jedi_forecast_model:
   - RunJediVariationalExecutable
   type: string-drop-list
 
+local_ensemble_inflation_mult:
```
danholdaway, line 405: As I alluded to in a previous comment not everything necessarily needs options. When it comes to numbers like floats and integers they might be omitted to allow users to pick anything they want. If you have options I think the type would generally need to be string and then drop list or check list. Perhaps we actually need to put some protection in place for that.

src/swell/tasks/task_questions.yaml
```diff
@@ -370,6 +404,94 @@ jedi_forecast_model:
...
+  tasks:
+  - RunJediLetkfExecutable
+  type: float
+
+local_ensemble_save_posterior_ensemble:
+  ask_question: false
+  default_value: defer_to_model
+  models:
+  - geos_atmosphere
+  options: defer_to_model
```
danholdaway: boolean options don't need to be deferred.

src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -106,7 +124,7 @@ def render_oops_file(self, config_name):
 
         # Path to configuration file
         config_file = os.path.join(self.jedi_config_path, 'oops', f'{config_name}.yaml')
-
+        print(config_file)
```
danholdaway: Remove rogue print.

danholdaway (changes requested):
Thanks again for adding this great new feature. Just a few minor things to change please. Also, as I'm reviewing the code I realize that "letkf" might not actually be the correct nomenclature for the suite and tasks etc. Perhaps we should go with `LocalEnsembleDa` / `local_ensemble_da` ?

src/swell/configuration/jedi/interfaces/geos_atmosphere/geos_atmosphere.yaml
CEgerer93: I agree - this would be consistent on both accounts.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
CEgerer93: Ideally no, the floats chosen for `local_ensemble_inflation_mult`, `local_ensemble_inflation_rtpp`, and `local_ensemble_inflation_rtps` would be left to the user to decide on.

I've removed the `options` for each, as the `default_value` sets values that appear standard.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
CEgerer93: Okay, this makes sense. I've moved `local_ensemble_save_posterior_ensemble{,_increments}` and `local_ensemble_save_posterior_mean{,_increment}` to the top-level `task_questions.yaml`.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
CEgerer93: Noted.

src/swell/deployment/prep_suite.py
CEgerer93: Implemented this suggestion. I wonder about the "Local" aspect however, as EnKF, and I believe GETKF, don't compute local updates. I'll review those schemes, and potentially return to a better suite name.

src/swell/utilities/render_jedi_interface_files.py
CEgerer93: Thanks. Done.

src/swell/tasks/run_jedi_letkf_executable.py
CEgerer93: Done. I've changed the variable `jedi_application` to reflect this as well.

src/swell/tasks/task_questions.yaml
CEgerer93: Understood. Removed options from `local_ensemble_inflation_mult`, `local_ensemble_inflation_rtpp`, and `local_ensemble_inflation_rtps`.

src/swell/tasks/task_questions.yaml
CEgerer93: Set `local_ensemble_save_posterior_*` options at top-level - not deferred to model.

CEgerer93 (commented):
Addressed/implemented comments on PR #256.

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
danholdaway: I think you still want them here but they should be just:

```
local_ensemble_save_posterior_ensemble:
  default_value: true
```

I.e. without the options part. The options don't differ between models, they are always true or false regardless of model.

src/swell/tasks/task_questions.yaml
```diff
@@ -313,6 +313,40 @@ gradient_norm_reduction:
   - RunJediVariationalExecutable
   type: float
 
+horizontal_localization_lengthscale:
+  ask_question: false
+  default_value: defer_to_model
+  models:
+  - geos_atmosphere
+  options: defer_to_model
```
danholdaway:
```suggestion
```

src/swell/tasks/task_questions.yaml
```diff
@@ -313,6 +313,40 @@ gradient_norm_reduction:
...
+  tasks:
+  - RunJediLocalEnsembleDaExecutable
+  type: float
+
+horizontal_localization_max_nobs:
+  ask_question: false
+  default_value: defer_to_model
+  models:
+  - geos_atmosphere
+  options: defer_to_model
```
danholdaway:
```suggestion
```

src/swell/tasks/task_questions.yaml
```diff
@@ -560,10 +687,80 @@ total_processors:
...
+vertical_localization_function:
+  ask_question: false
+  default_value: defer_to_model
+  models:
+  - geos_atmosphere
+  options: defer_to_model
+  prompt: Which localization scheme should be applied in the vertical?
+  tasks:
+  - RunJediLocalEnsembleDaExecutable
+  type: string
```
danholdaway: If you have a string with options can you select either `string-drop-list` or `string-check-list` depending on whether one or multiple choices are valid.

src/swell/tasks/task_questions.yaml
```diff
@@ -560,10 +687,80 @@ total_processors:
...
+
+vertical_localization_lengthscale:
+  ask_question: false
+  default_value: defer_to_model
+  models:
+  - geos_atmosphere
+  options: defer_to_model
```
danholdaway:
```suggestion
```

src/swell/utilities/run_jedi_executables.py
```diff
@@ -48,7 +48,18 @@ def jedi_dictionary_iterator(jedi_config_dict, jedi_rendering, window_type, obs,
                     for ob in obs:
                         # Get observation dictionary
                         obs_dict = jedi_rendering.render_interface_observations(ob)
+
+                        # Append ensemble localizations
+                        obsLocalize = []
```
danholdaway: I think this might add the `obs localizations` dictionary to the observations even when the applciation is not LocalEnsembleDa. I think you need to fence this somehow. A better thing might just be to add is in the RunJediLocalEnsembleDa task though, instead of here with an if statement.

danholdaway:
I sorted out the conflicts and have a few extra comments.

src/swell/tasks/task_questions.yaml
CEgerer93: Missed this mistaken deferral in `horizontal_localization_max_nobs` and `horizontal_localization_lengthscale` above - thanks.

src/swell/tasks/task_questions.yaml
CEgerer93: I had left `type` as `string`, as Gaspari-Cohn is the only scheme currently supported for vertical localization. This will probably change, so I will change `type` to a `string-drop-list`.

src/swell/utilities/run_jedi_executables.py
CEgerer93: Whoops, yes that would add localizations to observations for any task. I've removed the offending code block from `src/swell/utilities/run_jedi_executables.py` and adapted `src/swell/tasks/run_jedi_local_ensemble_da_executable.py` to accomplish the same behavior but for the `localensembleda` application only.

CEgerer93:
Addressed suggestions - perhaps the most important being the relocation of obs localization block into `run_jedi_local_ensemble_da_executable.py`.

danholdaway (approved):
Thanks for working on this great new feature @CEgerer93.

danholdaway:
Just running the tier 1 tests to make sure there is no impact elsewhere https://github.com/GEOS-ESM/swell/actions/runs/6562412615

PR #259, opened 2023-10-09
IAU for MOM6 cycling runs

Dooruk:
Some changes required to have IAU for the ocean model. This is a more stable and easier way of introducing increments to the ocean state rather than updating restart files.

Updated `background_error` parameters to conform with the 0.25 degrees setup.

IAU requires a file called `MOM_oda_incupd` in the experiment directory which is augmented with the MOM_input if IAU switch is turned on. Contents of the file is below. Increment is distributed equally throughout `ODA_INCUPD_NHOURS`.  For now it does not have U,V increments due to some changes required within `SOCA` and static bump files would need to be updated to consider U,V fields.

```
! === module MOM_oda_incupd ===
ODA_INCUPD = True   ! [Boolean] default = False
                             ! If true, oda incremental updates will be applied
                             ! everywhere in the domain.
ODA_INCUPD_FILE = "mom6_increment.nc"   ! The name of the file with the T,S,h increments.

ODA_TEMPINC_VAR = "Temp"        ! default = "ptemp_inc"
                                ! The name of the potential temperature inc. variable in
				      	   ! ODA_INCUPD_FILE.
ODA_SALTINC_VAR = "Salt"        ! default = "sal_inc"
                                ! The name of the salinity inc. variable in
                                ! ODA_INCUPD_FILE.
ODA_THK_VAR = "h"               ! default = "h"
                                ! The name of the int. depth inc. variable in
                                ! ODA_INCUPD_FILE.
ODA_INCUPD_UV = false            !
ODA_UINC_VAR = "u"              ! default = "u_inc"
                                ! The name of the zonal vel. inc. variable in
                                ! ODA_INCUPD_UV_FILE.
ODA_VINC_VAR = "v"              ! default = "v_inc"
                                ! The name of the meridional vel. inc. variable in
                                ! ODA_INCUPD_UV_FILE.
ODA_INCUPD_NHOURS = 3.0            ! default=3.0
```


src/swell/tasks/eva_observations.py
```diff
@@ -71,6 +71,14 @@ def execute(self):
 
             # Split the full path into path and filename
             obs_path_file = observation_dict['obs space']['obsdataout']['engine']['obsfile']
+
+            # Prevent Eva from failing if there are observation files with 0 observations
+            with nc.Dataset(obs_path_file, 'r') as ds:
+                loc_size = len(ds.dimensions['Location'])
+
+            if (loc_size) < 1:
+                continue
```
danholdaway, line 95: Can you put a `logger.info` here telling the user that no plots were produced for the instrument? It might help when they are confused as to why nothing it produced.

src/swell/tasks/task_questions.yaml
```diff
@@ -381,6 +381,18 @@ minimizer:
   - RunJediVariationalExecutable
   type: string-drop-list
 
+mom6_iau:
+  ask_question: false
+  default_value: defer_to_model
+  models:
+  - geos_ocean
+  options: defer_to_model
```
danholdaway: Can we just put the two options here instead of deferring to the model?

danholdaway (changes requested):
Thanks for adding the ocean IAU, great to have this. Two minor changes but otherwise LGTM

src/swell/tasks/task_questions.yaml
Dooruk: so have the default as false?

danholdaway: Default remains defer_to_model options becomes [true, false]

Dooruk:
All should be good now. 

Eventually, this will require some static Swell / JEDI files to be updated, but now that CICE6 updates are on the way (as well as GEOS version change) we can just merge these changes and we can update static files all together. Default for IAU is off so things should run.

danholdaway (approved):

PR #260, opened 2023-10-10
finding that Hyperspectral errors are not in sync w/ GSI numbers

rtodling:
## Description

I am finding that the errors in the hyperspectral IR yaml's are not consistent w/ those in sat-info ... these have substantial impact in results (making that too different that GSI's).

## Dependencies

None

## Impact

Correction of numbers that don't seem right.

I am attaching a fig here summarizing results of multiple cycles - and you can see that hyper IR stands out as very different from GSI's ... AMVs are also trouble as are radiosondes, but those are still being looked at.

imp0_ave


danholdaway:
This does cause a failure of the UFO tests. https://github.com/GEOS-ESM/swell/actions/runs/6470216995 Likely because it changes the observation counts. So we might need to update the reference counts. @gmao-jjin3 @gmao-wgu can you take a look at what the updated counts would need to be?

rtodling:
Just for completion here is the count over the same period the stats above were computed.

job_obscnt

As you can see the counts are not identical, but are reasonably close for most data ... the issue w/ the diagnostic in the previous plot is certainly obs error setting for IR. Again, AMVs and Radiosondes are still a mystery.

rtodling:
Sorry that my branch here is out of sync w/ develop ...

danholdaway:
No problem. Thanks for those count plots.

gmao-wgu:
@rtodling The obs errors we put in the yaml for AIRS, IASI and CrIS-FSR are consistent with the GSI, which are not specified from the satinfo file but from the diagonals of R matrix for all active channels. That is the reason you see the differences when compared with the satinfo file.

gmao-jjin3:
@gmao-wgu Thanks. Are those values in your input already inflated? Are they final observational errors?  You know there is error inflation in the configuration:     
https://github.com/GEOS-ESM/swell/pull/260/files#diff-efe795a1f98c099239af21d15b9ad360347370607c788a71388010ee7d8dad45L147
https://github.com/GEOS-ESM/swell/pull/260/files#diff-efe795a1f98c099239af21d15b9ad360347370607c788a71388010ee7d8dad45L147

gmao-wgu:
@gmao-jjin3 These are errors after inflation because they are obtained in the ana.log in GSI output not directly from the diagonal elements of R matrix.

gmao-jjin3:
No. That is the reason we were talking about doing the error inflation offline. There are two error inflation approaches implemented  in JEDI/UFO, but not the way we use in GSI. I will do inflation offline and generate the correlation matrix for all hyper spectral IR instruments.  JEDI/UFO either use the inter-channel correlations directly or convert to inter-channel correlations internally if a covariance matrix is provided.  The obs errors are provided from the error vector in the yaml.


> @gmao-jjin3 These are errors after inflation because they are obtained in the ana.log in GSI output not directly from the diagonal elements of R matrix.

@gmao-wgu  Then, we are inflating those again with your inputs, aren't we?

rtodling:
> @rtodling The obs errors we put in the yaml for AIRS, IASI and CrIS-FSR are consistent with the GSI, which are not specified from the satinfo file but from the diagonals of R matrix for all active channels. That is the reason you see the differences when compared with the satinfo file.

Wei, to your point. Indeed, I know that the values in the yaml files in "develop"  are the deflated values (as compared to what's in the satinfo files) - so here is what happens:

Jo = omb inv(R) mob 

The Jo calculation that comes out of JEDI is fully consistent, since R is diagonal in the run I have.
In the control run for GSI, R is not diagonal, but the Jo (in the figure - not the actual GSI calculation) uses diag(R) - and R is the same as that going in JEDI (develop). So the calculations of what shows in the plot are consistent between the two cycled system ... what I should be doing to be correct is to use the fully correlated inv(R) to do the offline Jo calculation that I am doing. I thought I implementing this a long time ago - but haven't done; I will work on it soon.

Regardless of this inconsistence I believe the obs errors that are placed in the yaml should be the ones in satinfo. This is what's consistent when no correlations are used .... what I should have done is simply turn off the correlated R in my control.

The other thing is, in your testing of the correlated R in JEDI, what should happen is that when we specify a correlated R, the code should ignore the default errors and use the ones in the given matrix (those will be fully consistent w/ GSI's - assuming you recreated R after reconditioning to give the JEDI. 

I am re-running the JEDI exp w/ the satinfo errors - what I am not doing is re-running my control w/o corr(R); perhaps I will do that ...

rtodling:
Folks I am going to add to what I showed above. This time I have taken a true step back and looked again at analyzing only radiosondes. I did two runs one using the swell yaml for sondes; another using what I had used in the past that had worked truly well for me - basically using the gsi QC decisions.

I can produce a bunch of figs similar to those I am popping here - but you should get the gist that the swell yaml does not allow me to get the data to behave as it should. 

All figures show increments, the top panel is GSI, the bottom is JEDI.

This is for PS using SWELL:

ps_swell_yaml

This is for PS using the simple yaml I had used in the past relying on GSI's qc.

ps_simple_yaml

The PS results are not ideal, but not horrible ... now T and Wind!

here is for T-swell (at 850):

t_850_swell_yaml

Here is for the simply yaml settings: 

t_850_simple_yaml

here is for U-200 SWELL:

u_200_swell_yaml

and the simple yaml:

u_200_simple_yaml


You get the picture ... 

I placed the two yaml's I am using under:  /discover/nobackup/rtodling/Trouble

I believe similar issues apply to sat winds ... perhaps others too. I will be testing sat winds next.

rtodling:
What is odd is that:
GSI gives:
           Jo Global         117976    1.6685721483494402E+05       1.414
JEDI gives:
 Nonlinear Jo(Radiosondes) = 86905.1, nobs = 117913, Jo/n = 0.737027, err = 7.9987

The obs counts are reasonably close; Jo and Jo/n too - remember the cost function in GSI does not include the 0.5.

rtodling:
Dan and I figured out the issue to be related to how the linear obs operator is been set for some of the conventional obs. I modified the raob as I had in the "good" above and now the "swell" case works. I am not revising the other conventional obs to handle the linear operator more consistently w/ the NL one ... let's see what I get.

danholdaway:
@rtodling is this a different issue than what we were looking at yesterday?

rtodling:
Now I have been looking at the radiance again ... AMSUA seems reasonable - not perfect, but acceptable. But the hyperspectral IR is completely wrong. 

Here is a fig similar to the ones I've been posting - comp of inc. - notice the increments in this case are inflated by a factor of 4 for visualization purposes.  

gsi_vs_jedi_hyperir_t_200

The the Jo(b) in this case for GSI and JEDI look like this:


GSI
20211201 120000 , Jo(AIRS-AQA) =  5.2113E+04, nobs =  370267 , Jo/n =  1.4075E-01, err =  1.2225E+00
20211201 120000 , Jo(CRIS-N20) =  8.4228E+03, nobs =  202452 , Jo/n =  4.1604E-02, err =  5.9820E-01
20211201 120000 , Jo(CRIS-NPP) =  6.6266E+03, nobs =  198678 , Jo/n =  3.3354E-02, err =  5.7773E-01
20211201 120000 , Jo(IASI-MTB) =  6.2515E+04, nobs =  593983 , Jo/n =  1.0525E-01, err =  9.0647E-01
20211201 120000 , Jo(IASI-MTC) =  5.6673E+04, nobs =  592469 , Jo/n =  9.5656E-02, err =  8.9960E-01
JEDI
CostJo   : Nonlinear Jo(airs_aqua)    = 87071.2, nobs = 319212, Jo/n = 0.272769, err = 1.07835
CostJo   : Nonlinear Jo(cris-fsr_n20) = 21786.4, nobs = 111001, Jo/n = 0.196272, err = 0.52304
CostJo   : Nonlinear Jo(cris-fsr_npp) = 18463.7, nobs =  96358, Jo/n = 0.191616, err = 0.500366
CostJo   : Nonlinear Jo(iasi_metop-b) =  143377, nobs = 449994, Jo/n = 0.318619, err = 0.685874
CostJo   : Nonlinear Jo(iasi_metop-c) =  145346, nobs = 451077, Jo/n = 0.32222 , err = 0.696811

Remember again that GSI(Jo) much be multiplied by 0.5.  The numbers are not ideal in terms of how they compare, but it's hard to reconcile the diffs above w/ those in the increment. 

I played games of removing, say, CrIS-FSR from the blend ... the incs discrepancy between JEDI and a consistent GSI control are similar ... something is really amiss.

rtodling:
Interesting to see what the PS increment due to Hyperspectral observations looks like:

Untitled

quite odd! 

Maybe there is a geoVal not quite correct in the case of IR ... 

and btw: the increment of Q due to HyperIR is got a funny look (likely because the Q control variable is different right now in JEDI and GSI), but it's too bad; at least it's not exploding.


Untitled

rtodling:
I've done a few more tests and I am getting the impression the problem here is the setting of the sat-bias-correction (VarBC) ... don't have a fig to illustrate, but I'd like to have somebody to talk to about the settings ...

danholdaway:
> I've done a few more tests and I am getting the impression the problem here is the setting of the sat-bias-correction (VarBC) ... don't have a fig to illustrate, but I'd like to have somebody to talk to about the settings ...

While @gmao-jjin3 and I were setting up a adas-swell test we noticed that GSI renames the bias correction coefficients after it's finished. That is to say, if JEDI used the same bias correction file that GSI used it would actually be the one for the next cycle. Where do you grab these coeffs from?

gmao-jjin3:
@rtodling  It looks that you numbers of cris-fsr in JEDI are about half of those in GSI.

gmao-jjin3:
@rtodling  Here is the PR to save the original bias correction coefficient files in the GSI FVWORK directory:
https://github.com/GEOS-ESM/GEOSana_GridComp/pull/152

danholdaway:
I believe this was resolved a different way so closing the PR. Feel free to reopen if you think it's still relevant.

PR #262, opened 2023-10-11
Replace obs filters by obs prior and post filters.

gmao-jjin3:
Move the "Variable Transforms" filter into the section of "obs prior filters."


danholdaway (approved):
Thanks @gmao-jjin3

PR #266, opened 2023-10-18
Feature/single obs test

gmao-jjin3:
Add an option to make and test single-observation files. Those files are saved from the first data in radiance ioda files or un-combined conventional ioda files.  In other words , there are more than one observation in some combined conventional files, but they are different observations such as temperature, station pressure, specific humidity, and winds.


gmao-jjin3:
I couldn't find the branch "feature/single_entry_point" which is the base of this PR.

gmao-jjin3:
A new JEDI/UFO PR is made https://github.com/JCSDA-internal/ufo/pull/3082, and is needed in order to successfully conduct single-observation tests with multiple processors.

src/swell/tasks/task_questions.yaml
```diff
@@ -529,6 +529,20 @@ produce_geovals:
   - GsiNcdiagToIoda
   type: boolean
 
+single_observations:
+  ask_question: true
```
danholdaway: Can you change this to false since we can consider it an advanced option and not something we want to ask users every time they are setting up an experiment.

danholdaway (changes requested):
One minor comment. THank you for adding this.

src/swell/tasks/task_questions.yaml
gmao-jjin3: Done.  Should we use "single_observations" or "single_observation"?

danholdaway: Probably doesn't matter too much.

danholdaway (approved):

PR #271, opened 2023-10-25
Remove the Yaml linting action

danholdaway:
## Description

Removing the YAML linting action

- Many YAMLs need to have Jinja2 templating and according to the developer of yamllint these kinds of files are not YAMLs and therefore cannot be linted.
- It is questionable that the linting actually enforces much of a consistent sytle.


asewnath:
On my github this looks like it's been stuck on the yaml coding norms for two days. Is that the case for anyone else? Just wondering if my github is glitching

asewnath (approved):

danholdaway:
> On my github this looks like it's been stuck on the yaml coding norms for two days. Is that the case for anyone else? Just wondering if my github is glitching

After the merge I will delete that check from GitHub,

PR #273, opened 2023-10-30
Accessing GSI observing system records for active and available channels

asewnath:
## Description

Initial PR making changes required to update yaml use flags during cylc workflow run. PR does the following:
- Change amsua_n19.yaml to include active channel jinja templating in `use_flag` field
- Adding two new tasks: `CloneGeosAna` and `GenerateObservingSystemRecords`
- Created new object `ObservingSystemRecords` to handle parsing active and available channel tables from GSI
- Modified `render_interface_observation` function to call new `get_active_channels` function. 
- Modifying relevant tasks to get `observing_system_record_path` for `render_interface_observation` or `jedi_dictionary_iterator` function

Changes that still need to be made: (can be in this PR or another one)
- Update all yaml files with `use_flag` field to have active channel jinja templating
- Available channels jinja templating in all yaml files
- Add logger into functions for proper aborts


src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_n19.yaml
```diff
@@ -191,9 +191,10 @@ obs post filters:
         channels: *amsua_n19_available_channels
         use passive_bc: true
         sensor: *Sensor_ID
-        use_flag: [-1, -1, -1,  1,  1,
-                    1, -1, -1,  1,  1,
-                    1,  1,  1,  1, -1]
+        use_flag: &amsua_n19_use_flag {{amsua_n19_active_channels}}
+        #use_flag: [-1, -1, -1,  1,  1,
```
danholdaway: Can you remove the comments? Below as well.

src/swell/suites/ufo_testing/flow.cylc
```diff
@@ -34,11 +34,18 @@
...
+
+            # Clone geos ana for generating observing system records
+            CloneGeosAna
         """
 
         {% for cycle_time in cycle_times %}
         {{cycle_time.cycle_time}} = """
 
+            # Generate satellite channel records
+            #GenerateObservingSystemRecords
```
danholdaway: Delete this commented line

src/swell/suites/ufo_testing/flow.cylc
```diff
@@ -101,6 +109,12 @@
             --partition={{scheduling["BuildJedi"]["partition"]}}
             {% endif %}
 
+    [[CloneGeosAna]]
+        script = "swell task CloneGeosAna $config -d $datetime -m geos_atmosphere"
```
danholdaway: Can you remove the -d and -m arguments from this call. The tasks should not depend of either.

src/swell/tasks/base/task_registry.py
```diff
@@ -0,0 +1,48 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
```
danholdaway: I thought this file got removed?

src/swell/tasks/gsi_bc_to_ioda.py
```diff
@@ -47,11 +51,17 @@ def execute(self):
         sensors = []
         sensors_satbias = []
         sensors_tlapse = []
+
+        cycle_time = os.path.normpath(cycle_dir).split('/')[-2]
```
danholdaway: There should be a `self.cycle_time_dto()` and `self.cycle_time()`. Please use one of these instead.

src/swell/tasks/task_questions.yaml
```diff
@@ -847,3 +848,26 @@ window_type:
   - StoreBackground
   type: string-drop-list
 
+observing_system_records_gsi_path:
```
danholdaway: Can you add a:

```
models:
- all
```

danholdaway: And below

src/swell/test/code_tests/generate_observing_system_test.py
```diff
@@ -0,0 +1,11 @@
+import os
+from swell.utilities.observing_system_records import ObservingSystemRecords
+
+observations = ['amsua_n19']
+observing_system_records_path = './yaml_output/'
+path_to_geosana_gridcomp = '/discover/nobackup/asewnath/github/GEOSana_GridComp/'
```
danholdaway: Ideally a unit test should be able to run regardless of where it is run, for example if we want to run it on GitHub. Is there a way of doing this without the absolute path to something in your space?

src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -149,8 +153,31 @@ def render_interface_model(self, config_name):
     # ----------------------------------------------------------------------------------------------
 
     # Prepare path to interface observations file and call rendering
-    def render_interface_observations(self, config_name):
+    def render_interface_observations(self, config_name, path_to_observing_sys_yamls, cycle_time):
```
danholdaway: Instead of passing cycle_time to this method each time can you pass it to the constructor along side cycle_dir?

src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -149,8 +153,31 @@ def render_interface_model(self, config_name):
...
+
+        # Add active channels to template dictionary
+        self.__template_dict__[f'{config_name}_active_channels'] = active_channels
+
+        # Render templates in file and return dictionary
+        return self.__open_file_render_to_dict__(config_file)
+
+    # ----------------------------------------------------------------------------------------------
 
+    def render_interface_ufo_test(self):
```
danholdaway: Why was a specific function for ufo_tests needed?

src/swell/utilities/run_jedi_executables.py
```diff
@@ -46,8 +52,11 @@ def jedi_dictionary_iterator(jedi_config_dict, jedi_rendering, window_type, obs,
                 if value_special == 'observations':
                     observations = []
                     for ob in obs:
-                        # Get observation dictionary
-                        obs_dict = jedi_rendering.render_interface_observations(ob)
+                        obs_dict = jedi_rendering.render_interface_observations(
+                            ob,
+                            observing_system_records_path,
```
danholdaway: Because of this being called recursively and from multiple places...what about adding a method to `jedi_rendering` called `set_observing_system_records_path`? Then you don't have to change the interface to this method. You could just add a check internally to `render_interface_observations` that makes sure the variable was set. This way if there are (in the future) calls to run_jedi_executable that doesn't involve observation we wont end up asking a question about the this path. It would avoid making any changes to the `jedi_dictionary_iterator` API.

danholdaway (changes requested):
Thank you very much for adding this very important functionality @asewnath. I just have a few minor comments about code readability. The overall functionality looks great.

The only other comment I would make is in having these three files: `sat_db_utils`, `instr_state_machine` and `observing_system_records`. Do we still need all three? What is the difference between what's in the files? It's not very easy to tell by the names of the files.

asewnath:
> Thank you very much for adding this very important functionality @asewnath. I just have a few minor comments about code readability. The overall functionality looks great.
> 
> The only other comment I would make is in having these three files: `sat_db_utils`, `instr_state_machine` and `observing_system_records`. Do we still need all three? What is the difference between what's in the files? It's not very easy to tell by the names of the files.

The `observing_system_records` file contains the `ObservingSystemRecords` object, which has methods to parse the GSI records and save the yaml files.`instr_state_machine` is an old name that can be changed to something like `gsi_tbl_parser`. It contains the code that actually parses the GSI records. Since it's a lot of code, it seems cleaner to have it in a separate file. The `sat_db_utils` file contains different utilities for the channel update process. `read_sat_db` is used only by `ObservingSystemRecords` and can be moved to `observing_system_records`. `run_sat_db_process` is legacy code that can be deleted. The only place it is called in is `utilities/scripts/swell_sat_db_processing.py`, which can also be deleted. Then there are functions that are used for the channel update that are not associated with the `ObservingSystemRecords` object. This includes `get_active_channels` and other functions necessary for the active channel retrieval. A separate file seems appropriate for these functions, and the file itself can be changed to `get_active_channels` or something along those lines. I can also add more documentation to all these files.

danholdaway:
> > Thank you very much for adding this very important functionality @asewnath. I just have a few minor comments about code readability. The overall functionality looks great.
> > The only other comment I would make is in having these three files: `sat_db_utils`, `instr_state_machine` and `observing_system_records`. Do we still need all three? What is the difference between what's in the files? It's not very easy to tell by the names of the files.
> 
> The `observing_system_records` file contains the `ObservingSystemRecords` object, which has methods to parse the GSI records and save the yaml files.`instr_state_machine` is an old name that can be changed to something like `gsi_tbl_parser`. It contains the code that actually parses the GSI records. Since it's a lot of code, it seems cleaner to have it in a separate file. The `sat_db_utils` file contains different utilities for the channel update process. `read_sat_db` is used only by `ObservingSystemRecords` and can be moved to `observing_system_records`. `run_sat_db_process` is legacy code that can be deleted. The only place it is called in is `utilities/scripts/swell_sat_db_processing.py`, which can also be deleted. Then there are functions that are used for the channel update that are not associated with the `ObservingSystemRecords` object. This includes `get_active_channels` and other functions necessary for the active channel retrieval. A separate file seems appropriate for these functions, and the file itself can be changed to `get_active_channels` or something along those lines. I can also add more documentation to all these files.

Thanks for the explanation. I think some renaming and clean up of any unused code should be sufficient.

src/swell/utilities/run_jedi_executables.py
asewnath: I'll also do this to set the cycle time so that `render_interface_observations` can go back to passing in just `config_name`

danholdaway: Sounds good.

src/swell/utilities/render_jedi_interface_files.py
asewnath: Ah, missed that you already talked about passing cycle_time to the constructor earlier!

danholdaway (approved):

PR #276, opened 2023-11-15
Adding inter-channel corrrelations into obs errors for IR

gmao-wgu:
## Description

The inter-channel correlations are added into the obs errors for hyperspectral IR observations.

## Dependencies

- [ ] NA

## Impact

- [ ] Need further testing


rtodling:
I think this can be closed ... what's in develop already incorporates correlated R.

rtodling:
As we talked about Wei: the most important thing on this is make sure the obs counts make sense and that the proper changes are being used - there seems to be too much going on at high-levels ... way more than in a corresponding GSI run. And this I see independently of any corr(R)  stuff.

As we talked about: some of what might cause this could be associated w/ the geovals still not being quite right.

PR #278, opened 2023-11-30
Adding jinja templating to observation yamls for active/available channels

asewnath:
## Description

This PR is a continuation of the work to incorporate active/available channel information from GEOSadas into swell. Logic was added to handle available channel templating, and observation yamls containing qc filters with use_flags were templated to receive active/available channel fields.


src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -191,12 +194,20 @@ def render_interface_observations(self, config_name):
             # Check that the config_name is not ufo_tests
             if config_name != 'ufo_tests':
 
-                # Get active channels
-                active_channels = get_active_channels(self.observing_system_records_path,
-                                                      config_name, self.cycle_time)
+                # Get available and active channels
+                result = get_channels(self.observing_system_records_path,
```
danholdaway: Can you just make the return here `None, None`? If so it would save the if statement below.

danholdaway (changes requested):

danholdaway (approved):

PR #298, opened 2024-02-15
Towards ensemble hofx support

### LINKED ISSUE #288: Add new Swell CLI option
A potential workflow to support computation of ensemble observers in the `localensembleda` suite includes the definition of a new variable `ensemble_hofx_strategy`; this variable will allow toggling between serial and parallel run patterns. It would be helpful for Swell to know how to distribute subsets of H(x) computations between allocated nodes when `ensemble_hox_strategy = parallel`. One thought is to include a new command line option to `swell task`, such as the following:

https://github.com/GEOS-ESM/swell/blob/1af9b232063f7c33648dbcf1b67b55d134f058e6/src/swell/suites/localensembleda/flow.cylc#L163
where `-e {{packet}}` is a new option.

I have an idea where this could be done, but I worry I could be breaking some functionality. I do realize this may be a little premature as an issue, especially given the holding pattern Swell is currently in. Nonetheless, I'm curious what insight might be gained.

CEgerer93:
## Description

Extends the `localensembleda` suite by separating the ensemble H(x) computation from the analysis step. Introduced structures to compute ensemble observers in either a serial or parallel workflow.

- In addition, a new option (`-p`) has been added to `swell task` that should be triggered when ensemble workflows are run. This new option specifies to an ensemble-related task which **"packet"** of ensemble members it should handle. Together with additional suite/task questions, ensemble-related tasks can operate on mutually exclusive subsets of ensemble members.

## Impact

- [ ] introduces a default parameter to `RunJediHofxExecutable` class in `run_jedi_hofx_executable.py` to produce hofx yamls for each ensemble member; minor rearrangement of said script as well.

@Dooruk edit: 

@asewnath and I had to make some additional changes to make sure `hofx_executable` wasn't impacted. 

I confined the ensemble_packet related entries to `task_questions`, which worked fine but we should see how do they work with the questionnaire. 

There still needs to be another task to combine individual h(x)_* outputs as LETKF executable requires observations to be in the same IODA file with the hofx0** naming convention.

Thanks @CEgerer93 for working on this and for your changes.


asewnath:
task_question_dicts swell test passing locally but not on github actions

rtodling:
I think you are requesting for too many reviewers for this. It delays things too much.

src/swell/suites/localensembleda/flow.cylc
```diff
@@ -60,12 +60,37 @@
...
 
-            # Run Jedi hofx executable
-            BuildJediByLinking[^]? | BuildJedi[^]  => RunJediLocalEnsembleDaExecutable-{{model_component}}
-            StageJedi-{{model_component}}[^] => RunJediLocalEnsembleDaExecutable-{{model_component}}
-            StageJediCycle-{{model_component}} => RunJediLocalEnsembleDaExecutable-{{model_component}}
-            GetEnsemble-{{model_component}} => RunJediLocalEnsembleDaExecutable-{{model_component}}
-            GetObservations-{{model_component}} => RunJediLocalEnsembleDaExecutable-{{model_component}}
+            # Run hofx for ensemble members according to strategy
+            {% if ensemble_hofx_strategy == 'serial' %}
+                BuildJediByLinking[^]? | BuildJedi[^] => RunJediHofxEnsembleExecutable-{{model_component}}
```
rtodling, line 65: What are all these instances of the same thing here? It looks like this should/could be wrapped in some sort of loop, no?

src/swell/tasks/run_jedi_hofx_ensemble_executable.py
```diff
@@ -0,0 +1,169 @@
...
+                                            jedi_executable)
+
+        # Run the JEDI executable
+        # -----------------------
+        if not generate_yaml_and_exit:
+            self.logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
+            run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
+                           jedi_config_file, output_log_file)
+
+        # TODO: save_geovals combination, if ever needed
```
rtodling, line 165: I don't think anyone will look at GeoVals for the ensemble of hofx - for most part, geovals is a means to compare against GSI - once the comp has been done for a single hofx - and deemed satisfactory - all others will be good.

There might be an instance when geovals could be of relevance, this would be when dealing w/ OSSEs - but I think the OSSE hook is still somewhat in the air (under work and investigation) in the deterministic system - really far from being used in the ensemble.

rtodling (approved):
Other than minor comments, and w/o trying to run this - things look ok by me.

src/swell/tasks/run_jedi_hofx_ensemble_executable.py
Dooruk: Ok, thanks. I figured there might be a fringe case but I will erase that line at some point after merging to develop.

src/swell/suites/localensembleda/flow.cylc
Dooruk: It depends if one wants to run in serial (no ensemble packets) vs. parallel (with ensemble packets). That's why there is repetition and some has {{packet}} at the end.

This flow.cylc needs another task to combine hofx_* outputs (which requires IODA_v3 handling) and final organizing before LETKF is executed. It is good for now though.

PR #297, opened 2024-02-15
Cloning GEOS_mksi instead for GSI channel records

asewnath:
## Description

Thanks to @mathomp4, we have a lightweight repository for accessing GSI channel records. 

Changes include:
- Renaming task references to CloneGeosMksi and task script is called clone_geos_mksi.py
- Modifying clone_geos_mksi.py to clone GEOS_mksi instead and updating all paths as necessary
- Adding ability to use a tag for GEOS_mksi (question shows up in experiment.yaml)

Closes #289

PR #299, opened 2024-02-16
Fix to check for missing obs in SaveObsDiag

### LINKED ISSUE #295: missing observations require modifying experiment.yaml
[This previous pull request ](https://github.com/GEOS-ESM/swell/pull/286)updates executable configs inside the model directories but does not modify the `experiment.yaml`. So when there are additional tasks looping over observations, they fail when there is an observation input with 0 locations. This is what's happening with `SaveObsDiags`.

We didn't catch this with Eva because I have the following section: 

https://github.com/GEOS-ESM/swell/blob/6037ab6c230a868209c292b0e02e7590e7932af5/src/swell/tasks/eva_observations.py#L110-L113

So I need to take the empty observation sources out of the `experiment.yaml` once we identify it.

@asewnath

asewnath:
## Description

Adds `check_obs` function to `SaveObsDiag` to make sure code downstream isn't trying to execute if obs is missing. Also removed some logic added earlier in `eva_observations.py` because the `check_obs` function takes care of it now.

Added `sst_ostia` back to geos-ocean hofx swell run and everything seems to be working as expected

Closes #295 
(Thanks for finding this bug @Dooruk!)

### DIFF
--- a/src/swell/__init__.py
+++ b/src/swell/__init__.py
@@ -9,4 +9,4 @@ import os
 repo_directory = os.path.dirname(__file__)
 
 # Set the version for swell
-__version__ = '1.8.1'
+__version__ = '1.8.3'
--- a/src/swell/tasks/eva_observations.py
+++ b/src/swell/tasks/eva_observations.py
@@ -87,7 +87,6 @@ class EvaObservations(taskBase):
 
         # Set the observing system records path
         self.jedi_rendering.set_obs_records_path(self.config.observing_system_records_path(None))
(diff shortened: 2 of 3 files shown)


Dooruk (approved):
Looks good to me, thanks @asewnath. Can I merge https://github.com/GEOS-ESM/swell/pull/297 first so you just need to pull those changes?

Dooruk:
Also would you mind bumping the swell version number?

PR #303, opened 2024-03-08
Modifying Airs Aqua passed benchmark

asewnath:
## Description

Passed benchmark change to fix tier testing


rtodling (approved):
This looks quite reasonable to me.

Dooruk:
Thanks @asewnath. Just FYI, there is an issue with eva_log and 3dvar output so tier1 test will still fail for 3dvar. I identified the issue (instead of gradient reduction the residual norm is displayed) but I will fix it on another PR.

PR #312, opened 2024-03-18
Add 4D-Ens-Var option for GEOS-atmosphere

rtodling:
## Description

This puts the pieces together to allow running hybrid-4D-Ens-Var for the GEOS atmosphere option of SWELL.

In adding this I noticed that the existing variational settings distinguishing only between 3D and 4D were not enough to accommodate options related to 4D-Var vs 4D-Ens-Var. For this reason, the oops interfaces were changed to instead of being "{variational}{window_type}" to "{variational}{cost_type}"  - this is way more general and easier to adjust for different options of JEDI.

One or two minor yaml's had to be adjusted on the geos-ocean side of things to fit in the new change ... still need to be tested to make sure nothing is broken.

With respect to the hybrid 4d-ens-var added here, there are still a couple of issues pending:

- 4d-background needs to be properly templated or generalized to expand syntax for 4d: https://github.com/JCSDA-internal/fv3-jedi/issues/1163
- it used to be possible to get 4d increments out of each outer loop in jedi - somehow I can't get the anymore: https://github.com/JCSDA-internal/fv3-jedi/issues/1164
- need to teach SWELL to fetch and link (lat-lon) ensemble.
## Dependencies

None

## Impact

None


rtodling:
I seem to be getting zero feedback in the two issues I refer to above. My plan to bypass the need for those to be resolved  is the following:

- have logic in swell (whoever ugly) to handle the 4d-background settings in the yaml
- have a task added to swell that, in 4d, will look for the analysis, diff them from the background, and convert the corresponding increment to lat-lon - this is what I've been doing offline (outside the context of SWELL).

The two items above are temporary stop gaps to allow for the 4d changes here to move onto develop. I will continue to pursue the answer to the two issues placed in fv3-jedi (above).

PR #324, opened 2024-04-22
Refactor the prep config

danholdaway:
## Description

Finishing up something I started a while back. Of course now I unfortunately cannot properly test it. However I did check that the code produces the exact same `experiement.yaml` files as develop and the exact same `flow.cylc` files. Will happily hear about failures of the tests that run on Discover and try to modify things.

The basic gist is that the previous structure was limited in the sense that you could not easily reference the suite and ask questions in a dynamic way. With these change you can now put essentially unlimited templating into the suite files. You can check the suite files, ask questions, then modify the suite file before asking more questions. So for example lets say you wanted to get 3D backgrounds for ocean and 4D for atmosphere you can put a template in the suite file and have two tasks, one for 3D and one for 4D. You have a 'suite' question that is window_type (promoted from task questions) and then template the suite file using window type before you ask the task questions for ocean and the task questions for atmosphere.


src/swell/utilities/r2d2.py
```diff
@@ -0,0 +1,52 @@
...
+
+    with open(r2d2_config_file_template, 'r') as f:
+        r2d2_config_file_template_str = f.read()
+
+    # Create a dictionary containing r2d2_local_path
+    r2d2_config_dict = {'r2d2_local_path': r2d2_local_path}
+
+    # Replace the template with the dictionary
+    r2d2_config_file_template_str = template_string_jinja2(logger, r2d2_config_file_template_str,
+                                                           r2d2_config_dict)
```
Dooruk, line 42: `template_string_jinja2` is not defined here.

Dooruk (changes requested):
Thanks @danholdaway for revisiting this branch to finish up your work. This will help us switch from `geos_ocean` to `geos_marine` (to include sea-ice DA)!

I locally tested it and it only requires the following changes (plus develop merge. Afterwards, I want to go through the code in detail and wait for input from the other reviewers. I can't push changes to your branch:

```
from swell.utilities.jinja2 import template_string_jinja2

def create_r2d2_config(platform, cycle_dir, r2d2_local_path):

    # Create a logger
    logger = Logger('CreateR2D2config')
```

I also ran the Tier1 workflow with it (after carefully examining for malicious code snippets 🦺 ). It provides an error that I didn't encounter on my end but the failed script is what you just added in the PR so not sure what went wrong, I will check it in detail.

https://github.com/GEOS-ESM/swell/actions/runs/8790732293

danholdaway:
@Dooruk I had forgotten about the R2D2 part of this PR. I've cleaned that up now. I'd started to make the R2D2 parts task instead of suite questions. Hopefully that's working now though I can't test it because I don't remember the hash of r2d2 that we were using with Swell so I can't install it myself. If you know it let me know and I can test it out. I can't replicate that import error. At one point in the history of the PR there was a missing __init__.py file in the deployment part and I saw the same error but it's working for my currently.

ghost:
@danholdaway  I think the old r2d2 commit was 06854f0100e56f44f7bf69d4bbd3fbe59e8c21c8

Dooruk (approved):

Dooruk:
This overall looks good thank you again @danholdaway. I want to test with the CLI input before I merge it but Discover is super slow right now. 

Are there any other reviews, before I merge this?

Dooruk (changes requested):
Eh, I spoke too soon..

After the latest commit, `fetch` is not seeing the `r2d2_local_path`, and resorting to the defaults and failing. It wasn't happening before that. I looked briefly but couldn't identify the issue, will look again on Monday.

Screenshot 2024-05-03 at 6 09 04 PM

rtodling:
I want to make sure none of the changes here would prevent us, GMAO, from getting background that are in R2D2. I haven't tried grabbing  the stuff in this PR and running myself .. I also think the changes here are very extensive and GMAO needs to have a conversation about the ordering in which we take the PRs that are lined up right. Something might be way easier to incorporate before taking the changes from this PR.

danholdaway:
Makes sense. It's tough to take such a change when the author also cannot really run the code. For example it might be tricky for me to debug what Doruk is seeing around R2D2. Perhaps the way to do this is to get to a point where you feel ready to take it (if you want it) and then we try to have a short code sprint (or something like that) to finalize any issues before you merge. That might save a lot of back and forth.

Dooruk:
I want to clarify, this PR is separate from the config builder tool (see [here](https://github.com/NOAA-EMC/jcb)) @danholdaway is working on and I mentioned a few times. As Dan described, it's a more flexible way to handle how the task questions are structured and R2D2 config yamls are created in the cycle directories. The issue I'm having is that R2D2/fetch does not see that R2D2_config yaml currently, I'm working on it right now though.

That's what I was going to suggest @danholdaway, to move branch to Swell so I can test and modify locally rather than bugging you after each obstacle. One quick question though, what is the purpose of setting `R2D2_CONFIG_FILE` environment variable? Does R2D2/fetch use it? I don't see it in R2D2 repo.

danholdaway:
I thought that the env variable was used internally to R2D2.

Dooruk:
> I thought that the env variable was used internally to R2D2.

Well, that was the bug 😄  It's `R2D2_CONFIG` not `R2D2_CONFIG_FILE` and also `SaveObsDiags` task is missing `create_r2d2_config`. Anyhow, I can move the branch to Swell, so you don't get pinged every time @danholdaway, thanks again!

@rtodling I tested suites and they work (other than 3dvar_atmos but that is due to a recent JEDI change) but happy to discuss changes further with you.

PR #328, opened 2024-04-25
Re-instate hofx scaling in observation operators.

### LINKED ISSUE #304: 3DVar Fails: Surface wind scaling pressure
The following:

hofx scaling field: SurfaceWindScalingPressure
hofx scaling field group: DerivedVariables

in the nonlinear and linear operators of most conventional obs causes 3dvar to fails.

Likely this needs a geoval that is improperly defined. Or the linear model has something incorrectly set.

gmao-jjin3:
Add back hofx scaling in linear and non-linear observation operators for convectional data.

Dependencies

- [x] waiting on https://github.com/JCSDA-internal/ufo/pull/3299

Solves the issue https://github.com/GEOS-ESM/swell/issues/304

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/aircraft.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/aircraft.yaml
@@ -24,8 +24,25 @@ obs operator:
     - name: windNorthward
 
     # Hofx scaling
-#   hofx scaling field: SurfaceWindScalingPressure
-#   hofx scaling field group: DerivedVariables
+    hofx scaling field: SurfaceWindScalingPressure
+    hofx scaling field group: DerivedVariables
+
+  - name: VertInterp
+    observation alias file: '{{experiment_root}}/{{experiment_id}}/configuration/jedi/interfaces/{{model_component}}/observations/obsop_name_map.yaml'
+    variables: [airTemperature]
+
+linear obs operator:
+  name: Composite
+  components:
+
+  - name: VertInterp
+    variables:
+    - name: windEastward
+    - name: windNorthward
+
+    # Hofx scaling
+    hofx scaling field: SurfaceWindScalingPressure
+    hofx scaling field group: DerivedVariables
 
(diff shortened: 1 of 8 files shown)


rtodling (approved):
These definitely fix the issue w/ JoJc getting to be negative - this no longer happens; 3DFGAT runs to the end and produces a healthy increment. 

Now, I seem to be seeing more difference w/ GSI results than I was seeing before ... I will make some plots and illustrate this elsewhere. I think for the purposes here - the changes should be taken and move to develop. The difference in increments is another issue.

This is good by me ... and btw: did pull the latest JEDI to be in sync w/ the yaml adjustments here.

Thanks for the hard work JJ - and I know Fabio was heavily involved too. Thanks to both of you.

Dooruk:
`UFO` test failed due to the infamous `passedBenchmark`. Could someone update those numbers please?

I took a quick look of the definition of the `passedBenchmark` in `UFO` and I don't see a tolerance option (see below code section). Could there be an another benchmark that we could use to avoid this issue keep happening? 

https://github.com/JCSDA-internal/ufo/blob/46b95c8b7da3e11b6d4795e2b8a1b5a553f71033/test/ufo/ObsFilters.h#L492-L497

gmao-jjin3:
@Dooruk  I don't see any failed tests in my JEDI built yesterday or today.  I suggest you have another run after checking out the latest JEDI.

Dooruk:
> @Dooruk I don't see any failed tests in my JEDI built yesterday or today. I suggest you have another run after checking out the latest JEDI.

Ok, triggering another with the last evening build.

Dooruk:
@gmao-jjin3 it failed again, there is just minor difference in `scatwind` and likely few other instruments that needs to be tested:

Screenshot 2024-04-26 at 9 29 19 AM

gmao-jjin3:
Okay. Updated some benchmark numbers.

Dooruk:
Thanks @gmao-jjin3 that worked.

I can't tag your github handle but Fabio thank you for your help!

PR #331, opened 2024-04-30
use-flag cannot be wired; comes from satdb

rtodling:
## Description

Use flag settings were wired ... these need to come from sat-db (mksi)

## Dependencies

None

## Impact

Correct use of data in general way - for now this is correct since in the test cases the choice wired was consistent w/ satdb, but not in general. This removes the wiring.


gmao-jjin3 (approved):

Dooruk:
Would this require UFO test changes?

Dooruk:
> Would this require UFO test changes?

It didn't, merging.

PR #336, opened 2024-05-07
Refactor prep config

Dooruk:
Same thing as https://github.com/GEOS-ESM/swell/pull/324 that @danholdaway introduced with additional fixes.

This shouldn't change (and didn't change as far as I could test) any of the configuration YAML outputs. I tested creating experiments with the defaults/overrides and Callum and I tested the CLI interface. 

A few highlights:
- It replicates the `task_questions` logic with the `suite_questions` so we don't have to have `suite_questions` in each suite folder.  
- It changes Jinja2 utility slightly to add a custom "undefined" class, which essentially gives Jinja2 an option to skip certain template variables without breaking. 
- It moves R2D2 config creation to utilities but again, doesn't change the configuration.
- This should help with the model dependent tasks (e.g., a variable change task only required for CICE)

I made a few minor fixes and merged with SLURM changes by @ashiklom (I had to add dashes to avoid whitespace between SLURM directives and added `items` to jinja2 undefined class).

So please take a look and test if you can and see if anything breaks. I'm planning on merging this next week.


Dooruk:
This is ready for merge, please take a look if you can.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+2. The code then builds a list of suite questions that do not depend on the model. At around the
+   same time the code determines all the tasks within that suite that do not depend on the model
+   component. At that point all these questions are asked using the question client.
+
+3. If there is some dependency on model component(s) in the flow.cylc then the code will ask the
+   user which model components should be included. Then all the suite questions depending on the
+   model will be
+
+3. All the
```
ashiklom: Incomplete docs here?

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+    def __init__(self, logger, suite, platform, config_client, override, advanced):
+
+        # Store local copy of the inputs
+        self.logger = logger
+        self.suite = suite
+        self.platform = platform
+        self.override = override
+        self.advanced = advanced
+
+        # Assign the client that will take care of providing responses
```
ashiklom, line 65: Suggest adding an assertion here, since we only support 2 right now. (If we go with Pydantic, #349, we can replace this with an Enum).

```suggestion
        # Assign the client that will take care of providing responses
        assert config_client.lower() in ('cli', 'defaults'), f"Invalid client {config_client}. Must be either 'cli' or 'defaults'"
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+class PrepareExperimentConfigAndSuite:
+
+    # ----------------------------------------------------------------------------------------------
+
+    def __init__(self, logger, suite, platform, config_client, override, advanced):
```
ashiklom: `advanced` is never used in this Class, so we should remove it.

Also, per #349, as long as we are committing to Python >3.5 (which I strongly recommend!), I suggest adding some type hints:

```suggestion
    def __init__(
        self,
        logger: Logger,
        suite: str,
        platform: str,
        config_client: str,
        override: Union[str, dict, None]
    ) -> None:
```

Also, I'm not a huge fan of `override: Union[str, dict]` --- I would rather it just be `dict` (technically, `Optional[dict]`) and the `str` case (path to override file) be handled much earlier, when we parse command line arguments.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+        # Store local copy of the inputs
+        self.logger = logger
+        self.suite = suite
+        self.platform = platform
+        self.override = override
+        self.advanced = advanced
```
ashiklom: `advanced` is not used in this class (as far as I can tell), so we should remove it.

```suggestion
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+        # Assign the client that will take care of providing responses
+        if config_client.lower() == 'cli':
+            self.config_client = GetAnswerCli()
+        elif config_client.lower() == 'defaults':
+            self.config_client = GetAnswerDefaults()
```
ashiklom, line 69: ...or just raise an error here.

```suggestion
            self.config_client = GetAnswerDefaults()
        else:
            logger.abort(f"Invalid config_client {config_client}")
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+        # Perform the assembly of the dictionaries that contain all the questions that can possibly
+        # be asked. This
+        self.prepare_question_dictionaries()
+        self.override_with_defaults()
+        self.override_with_external()
+
+    # ----------------------------------------------------------------------------------------------
+
+    def prepare_question_dictionaries(self):
```
ashiklom:
```suggestion
    def prepare_question_dictionaries(self) -> None:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+            for key, val in self.question_dictionary_model_dep[model].items():
+                if val['models'] != ['all'] and model not in val['models']:
+                    keys_to_remove.append(key)  # Remove if not needed by this model
+
+            for key in keys_to_remove:
+                del self.question_dictionary_model_dep[model][key]
+
+    # ----------------------------------------------------------------------------------------------
+
+    def override_with_defaults(self):
```
ashiklom:
```suggestion
    def override_with_defaults(self) -> None:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                if key == 'experiment_id':
+                    val['default_value'] = f'swell-{self.suite}'
+
+                if key == 'model_components':
+                    val['default_value'] = self.possible_model_components
+                    val['options'] = self.possible_model_components
+
+    # ----------------------------------------------------------------------------------------------
+
+    def override_with_external(self):
```
ashiklom:
```suggestion
    def override_with_external(self) -> None:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+        if self.suite_needs_model_components:
+            for model, model_dict in self.question_dictionary_model_dep.items():
+                for key, val in model_dict.items():
+                    if model in override_dict['models'].keys():
+                        if key in override_dict['models'][model].keys():
+                            val['default_value'] = override_dict['models'][model][key]
+
+    # ----------------------------------------------------------------------------------------------
+
+    def ask_questions_and_configure_suite(self):
```
ashiklom: As currently designed, the type annotation here should be:

```suggestion
    def ask_questions_and_configure_suite(self) -> Tuple[dict, dict]:
```

However, the whole point of having the class is that we can just directly access the corresponding attributes in the calling code. I.e., We currently have:

```python
experiment_dict, comment_dict = prepper.ask_questions_and_configure_suite()
```

...when it is probably more idiomatic to have this method just modify the attributes (i.e., without a return statement) and then just grab the attributes directly. So, in the calling code:

```python
prepper.ask_questions_and_configure_suite()
experiment_dict = prepper.experiment_dict
comment_dict = prepper.comment_dict
```

In that case, the type annotation here would be:

```python
      def ask_questions_and_configure_suite(self) -> None:
```

(It's probably even _more_ idiomatic Python to just pass around the entire `prepper` object through various function calls. But that's a bigger change).

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                        # Ask the question
+                        self.ask_a_question(self.question_dictionary_model_dep[model], question_key,
+                                            model)
+
+        # Return the main experiment dictionary
+        return self.experiment_dict, self.questions_dict
+
+    # ----------------------------------------------------------------------------------------------
+
+    def ask_a_question(self, full_question_dictionary, question_key, model=None):
```
ashiklom:
```suggestion
    def ask_a_question(
        self,
        full_question_dictionary: dict,
        question_key: str,
        model: Optional[str]=None
    ) -> None:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                self.experiment_dict[question_key] = self.config_client.get_answer(question_key, qd)
+                self.questions_dict[question_key] = qd['prompt']
+            else:
+                self.experiment_dict['models'][model][question_key] = \
+                    self.config_client.get_answer(question_key, qd)
+                self.questions_dict[f'models.{model}.{question_key}'] = qd['prompt']
+
+    # ----------------------------------------------------------------------------------------------
+
+    def get_suite_task_list_model_ind(self, suite_str):
```
ashiklom:
```suggestion
    def get_suite_task_list_model_ind(self, suite_str: str) -> list:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+        # Ensure there are no duplicate tasks
+        tasks = list(set(tasks))
+
+        # Return tasks
+        return tasks
+
+    # ----------------------------------------------------------------------------------------------
+
+    def get_suite_task_list_model_dep(self, suite_str):
```
ashiklom:
```suggestion
    def get_suite_task_list_model_dep(self, suite_str: str) -> Tuple[dict, list]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+import copy
+import os
+import yaml
+
```
ashiklom, line 15: Let's begin our commitment to type hints! #349 

```suggestion
from typing import Union, Tuple, Optional

```

https://docs.python.org/3/library/typing.html#typing.Union

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+   model will be
+
+3. All the
+"""
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+class PrepareExperimentConfigAndSuite:
```
ashiklom, line 46: Strictly speaking, a class corresponds to an object and should therefore should be a noun, not a verb. I know this seems pedantic, but I think it's important because it reflects the unique roles of functions vs. classes.

Is there any reason not to ambitiously call this something like `SwellExperiment`? You have effectively encapsulated everything you need to run an experiment in here, right?

Or, if we think that means something different, I might at least call this something like `ExperimentConfigFactory`, to reflect the fact that it's an object that produces experiment configs.

src/swell/deployment/prepare_config_and_suite/question_and_answer_cli.py
```diff
@@ -8,20 +8,17 @@
...
 import questionary
 from questionary import Choice
 
-from swell.deployment.prep_config_base import PrepConfigBase
 
 # --------------------------------------------------------------------------------------------------
 
 
-class PrepConfigCli(PrepConfigBase):
+class GetAnswerCli:
```
ashiklom, line 21: Ditto to my other comment about class names. Strictly speaking, this should be a noun --- maybe `CliResponder` or `CliAnswerer`?

src/swell/deployment/prepare_config_and_suite/question_and_answer_defaults.py
```diff
@@ -8,25 +8,9 @@
...
 
 
-from swell.deployment.prep_config_base import PrepConfigBase
-
-
-# --------------------------------------------------------------------------------------------------
-
-
-class PrepConfigDefaults(PrepConfigBase):
+class GetAnswerDefaults:
```
ashiklom, line 11: Ditto to my other comment about class names --- maybe mirror the `CliResponder/CliAnswerer` with `DefaultResponder/DefaultAnswerer`.

src/swell/deployment/prepare_config_and_suite/question_and_answer_cli.py
ashiklom: This is definitely overkill for this PR, and probably in general, (i.e., ignore this comment except as a curiosity 😉), but just thinking out loud: This (and `GetAnswerDefault`) are a very natural use case for python's [Abstract Base Classes](https://docs.python.org/3/library/abc.html).

```python
from abc import ABC, abstractmethod

class Answerer(ABC):
    
    @abstractmethod
    def get_answer(self, key: str, val: list):
      pass

class DefaultAnswerer(Answerer):
    def get_answer(self, key: str, val: list):
        # Default implementation...

class CLIAnswerer(Answerer):
    def get_answer(self, key:str, val: list):
        # CLI implementation
    def other_method(...):
        # ...
    def other_method_2(...):
        # ...
```

src/swell/utilities/jinja2.py
```diff
@@ -7,28 +7,75 @@
 # --------------------------------------------------------------------------------------------------
 
 
-import jinja2
+import jinja2 as j2
```
ashiklom, line 13: To allow type hints on the `template_string_jinja2` function below.

```suggestion
from swell.utilties.logger import Logger

```

src/swell/utilities/jinja2.py
```diff
@@ -7,28 +7,75 @@
...
-                        f'{string_rendered}')
+    def __repr__(self):
+        return str(self)
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+def template_string_jinja2(logger, templated_string, dictionary_of_templates,
+                           allow_unresolved=False):
```
ashiklom:
```suggestion
def template_string_jinja2(
    logger: Logger,
    templated_string: str,
    dictionary_of_templates: dict,
    allow_unresolved: bool = False
) -> str:
```

src/swell/utilities/jinja2.py
```diff
@@ -7,28 +7,75 @@
...
+    # Extra safety checks
+    # -------------------
+    if not allow_unresolved:
+        logger.assert_abort('{{' not in string_rendered, f'In template_string_jinja2 ' +
+                            f'the output string still contains template directives. ' +
+                            f'{string_rendered}')
+
+        logger.assert_abort('}}' not in string_rendered, f'In template_string_jinja2 ' +
+                            f'the output string still contains template directives. ' +
+                            f'{string_rendered}')
```
ashiklom: I was staring at this for a while trying to figure out why there were two identical statements before I spotted the difference:

```suggestion
        logger.assert_abort(
            not (('{{' in string_rendered) or ('}}' in string_rendered)),
            f"""
            In template_string_jinja2, the output string still contains template directives:
            '''
            {string_rendered}
            '''
            """
        )
```

src/swell/utilities/jinja2.py
```diff
@@ -7,28 +7,75 @@
...
+
+    # Load the algorithm template
+    # ---------------------------
+    template = env.from_string(templated_string)
+
+    # Render the template hierarchy
+    # -----------------------------
+    try:
+        string_rendered = template.render(dictionary_of_templates)
+    except j2.exceptions.UndefinedError as e:
```
ashiklom, line 79: Since we're just re-throwing the exception, any reason not to just do this?

```suggestion
    except Exception as e:
```

Or, is the point that we do actually want to silently ignore all exceptions here _except_ the jinja2 `UndefinedError` exception? If so, I suggest a comment explaining why.

src/swell/utilities/r2d2.py
```diff
@@ -0,0 +1,52 @@
...
+
+import os
+
+from swell.swell_path import get_swell_path
+from swell.utilities.jinja2 import template_string_jinja2
+
+# --------------------------------------------------------------------------------------------------
+
+
+def create_r2d2_config(logger, platform, cycle_dir, r2d2_local_path):
```
ashiklom:
```suggestion
def create_r2d2_config(
    logger: Logger,
    platform: str,
    cycle_dir: str,
    r2d2_local_path: str
) -> None:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+    # ----------------------------------------------------------------------------------------------
+
+    def ask_a_question(self, full_question_dictionary, question_key, model=None):
+
+        # Set flag for whether the question should be asked
+        ask_question = True
+
+        # Has the question already been asked?
+        if question_key in self.experiment_dict.keys():
```
ashiklom: `"x" in somedict` automatically checks for presence of keys, and is much faster (uses dict hashing, so O(1) performance, whereas searching the list `somedict.keys()` is O(n). Here, it doesn't matter, but get into the habit!

```suggestion
        if question_key in self.experiment_dict:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+            ask_question = False
+
+        # Dictionary for this question
+        qd = full_question_dictionary[question_key]
+
+        # If model is not none then ensure the experiment dictionary has a dictionary for the model
+        if model is not None:
+            if 'models' not in self.experiment_dict.keys():
```
ashiklom:
```suggestion
            if 'models' not in self.experiment_dict:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                self.experiment_dict['models'] = {}
+                self.questions_dict['models'] = f"Configurations for the model components."
+            if model not in self.experiment_dict['models'].keys():
```
ashiklom:
```suggestion
            if model not in self.experiment_dict['models']:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                self.experiment_dict['models'][model] = {}
+                self.questions_dict[f'models.{model}'] = \
+                    f"Configuration for the {model} model component."
+
+        # Check the dependency chain for the question
+        if 'depends' in qd.keys():
```
ashiklom:
```suggestion
        if 'depends' in qd:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+            # Check is dependency has been asked
+            if qd['depends']['key'] not in self.experiment_dict.keys():
```
ashiklom:
```suggestion
            if qd['depends']['key'] not in self.experiment_dict:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                    # Ask the question
+                    self.ask_a_question(self.question_dictionary_model_ind, question_key)
+
+        # 9.2 Iterate over the model_dep dictionary and ask task questions
+        # ----------------------------------------------------------------
+        for model in self.experiment_dict['model_components']:
+
+            # Iterate over the model_dep dictionary and ask questions
+            # -------------------------------------------------------
+            for question_key in self.question_dictionary_model_dep[model].keys():
```
ashiklom: Dict iterator already iterates over keys, so `.keys()` here is redundant.

```suggestion
            for question_key in self.question_dictionary_model_dep[model]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+                # Ask only the task questions first
+                # ----------------------------------
+                if 'suites' not in self.question_dictionary_model_dep[model][question_key].keys():
```
ashiklom:
```suggestion
                if 'suites' not in self.question_dictionary_model_dep[model][question_key]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+        # --------------------------------------------------------
+        if not self.suite_needs_model_components:
+            return self.experiment_dict, self.questions_dict
+
+        # 6. Iterate over the model_dep dictionary and ask suite questions
+        # ----------------------------------------------------------------
+
+        # At this point the user should have provided the model components answer. Check that it is
+        # in the experiment dictionary and retrieve the response
+        if 'model_components' not in self.experiment_dict.keys():
```
ashiklom:
```suggestion
        if 'model_components' not in self.experiment_dict:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+        9. Ask the model dependent task questions.
+        """
+
+        # If the client is CLI put out some information about what is due to happen next
+        if self.config_client.__class__.__name__ == 'GetAnswerCli':
+            self.logger.info("Please answer the following questions to configure your experiment ")
+
+        # 1. Iterate over the model_ind dictionary and ask questions
+        # ----------------------------------------------------------
+        for question_key in self.question_dictionary_model_ind.keys():
```
ashiklom:
```suggestion
        for question_key in self.question_dictionary_model_ind:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+            # Ask only the suite questions first
+            # ----------------------------------
+            if 'suites' in self.question_dictionary_model_ind[question_key].keys():
```
ashiklom:
```suggestion
            if 'suites' in self.question_dictionary_model_ind[question_key]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+        # -------------------------------------------------------------
+        suite_str = template_string_jinja2(self.logger, self.suite_str, self.experiment_dict, True)
+
+        # 3. Get a list of tasks that do not depend on the model component
+        # ----------------------------------------------------------------
+        model_ind_tasks = self.get_suite_task_list_model_ind(suite_str)
+
+        # 4.1 Iterate over the model_ind dictionary and ask task questions
+        # ----------------------------------------------------------------
+        for question_key in self.question_dictionary_model_ind.keys():
```
ashiklom:
```suggestion
        for question_key in self.question_dictionary_model_ind:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+            # Ask the task questions
+            # ----------------------
+            if 'suites' not in self.question_dictionary_model_ind[question_key].keys():
```
ashiklom:
```suggestion
            if 'suites' not in self.question_dictionary_model_ind[question_key]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+        for model in self.experiment_dict['model_components']:
+
+            model_dict = self.question_dictionary_model_dep[model]
+
+            # Loop over keys of each model
+            for question_key in model_dict:
+
+                # Ask only the suite questions first
+                if 'suites' in model_dict[question_key].keys():
```
ashiklom:
```suggestion
                if 'suites' in model_dict[question_key]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+        suite_str = template_string_jinja2(self.logger, self.suite_str, self.experiment_dict,
+                                           True)
+
+        # 8. Build a list of tasks for each model component
+        # -------------------------------------------------
+        model_dep_tasks, all_tasks = self.get_suite_task_list_model_dep(suite_str)
+
+        # 9.1 Ask the new task questions that do not actually depend on the model
+        # -----------------------------------------------------------------------
+        for question_key in self.question_dictionary_model_ind.keys():
```
ashiklom:
```suggestion
        for question_key in self.question_dictionary_model_ind:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+
+            if 'tasks' in self.question_dictionary_model_ind[question_key].keys():
```
ashiklom:
```suggestion
            if 'tasks' in self.question_dictionary_model_ind[question_key]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+            # Get all elements of swell_task_lines that contains "-m {model}"
+            model_tasks_this_model = [line for line in swell_task_lines if f'-m {model}' in line]
+
+            # Get task name
+            tasks = []
+            for line in model_tasks_this_model:
+                tasks.append(line.split('swell task ')[1].split(' ')[0])
+
+            # Unique model tasks
+            model_tasks[model] = list(set(tasks))
```
ashiklom, line 601: In general, `list(set(...))` seems like a Python antipattern, especially for multiple calls. If we are sure that `model_tasks` should _always_ be unique (here elsewhere), we should just define it as a `set` and always pass it around as such.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+            if key in override_dict.keys():
+                val['default_value'] = override_dict[key]
+
+        # Iterate over the model_dep dictionary and override
+        # --------------------------------------------------
+        if self.suite_needs_model_components:
+            for model, model_dict in self.question_dictionary_model_dep.items():
+                for key, val in model_dict.items():
+                    if model in override_dict['models'].keys():
+                        if key in override_dict['models'][model].keys():
```
ashiklom:
```suggestion
                    if model in override_dict['models']:
                        if key in override_dict['models'][model]:
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -0,0 +1,601 @@
...
+                self.logger.abort(f'Override must be a dictionary or a path to a yaml file.')
+
+        # In this case the user is sending in a dictionary that looks like the experiment dictionary
+        # that they will ultimately be looking at. This means the dictionary does not contain
+        # default_value or options and the override cannot be performed.
+
+        # Iterate over the model_ind dictionary and override
+        # --------------------------------------------------
+        for key, val in self.question_dictionary_model_ind.items():
+            if key in override_dict.keys():
```
ashiklom:
```suggestion
            if key in override_dict:
```

ashiklom:
@Dooruk I gave this a pretty thorough review. Functionally, I think it looks good and I didn't find any bugs. My suggestions above are primarily related to style and, in a few cases, adding additional robustness or clarity.

That said, don't let my suggestions hold up this PR. Take a quick look through them and feel free to either directly commit them (I tried to make it a one-click thing using "Suggestions") or just "resolve" the conversation without implementing the changes. Especially for the type hinting stuff (#349), I can always just open a separate PR on that later.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: Hmm, I've been doing this all around the workflow, we should dedicate a `.keys[]` removal PR :)

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: Better software practices are always welcome!

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: Good thought, that is currently done under `deployment/create_experiment.py` :

```
    # Assert valid method
    # -------------------
    valid_tasks = ['defaults', 'cli']
    if method not in valid_tasks:
        logger.abort(f'In Suites constructor method \'{method}\' not one of the valid ' +
                     f'tasks {valid_tasks}')
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: Yes, Dan said below comments in the methods would be sufficient. I will erase these but we should revisit them when we improve the documentation.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: I would like to defer renaming classes and/or methods to a seperate discussion, perhaps involving the new hire.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: I'm keeping the `Tuple` suggestion for now.

src/swell/deployment/prepare_config_and_suite/question_and_answer_cli.py
Dooruk: Again, I'm suggesting this as a future PR :)

src/swell/deployment/prepare_config_and_suite/question_and_answer_defaults.py
Dooruk: Future concern #3

src/swell/utilities/jinja2.py
Dooruk: It's the latter. There is a description under `ask_questions_and_configure_suite` method in  `prepare_config_and_suite.py`. Essentially, with the way it is designed now, there are two jinja passes one of which being non-template resolving (non-exhaustive) vs. template resolving. Does this make sense?

For instance, I had to add `items(self)` in the `SilentUndefined` class to make sure jinja2 ignores the below part:

```
        {%- for key, value in scheduling["RunJediVariationalExecutable"]["directives"][model_component].items() %}
```

Dooruk: I will add a comment though.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
Dooruk: So you are saying if the order doesn't matter `list` is redundant since `set` already returns a list? Sure, I can test that and accept the change.

ashiklom: The point is more that, if `model_tasks[model]` should _always_ have unique values then the correct data type for it is a set (which always has only unique values) rather than a list.

```python
mylist = ["a", "a", "b", "c"]
myset = set(mylist)
myset
# {'a', 'b', 'c'}
myset.add("a")
myset
# {'a', 'b', 'c'}
```

At a minimum, that lets us remove a lot of these kinds of `list(set(...))` calls. Also, if we combine this with type checking then we can rely on the type checker to ensure that we are always using `set`s everywhere.

That said, `set`s have different methods (e.g., `add` instead of `append`; no `+` method), which may break some things. So **I think we should just leave this as is and fix it in a future PR**.

Dooruk: Sounds good, I was just testing it :)

Dooruk:
Thanks for your thorough review @ashiklom, I made most of the modifications you suggested and deferred some to a future PR. This is almost set to merge, except while I started fixing some issues with the CLI I encountered an error with the `iso-duration` type:

Screenshot 2024-06-12 at 9 54 11 AM

So `make_duration` in `question_and_answer_cli` is explicitly set to handle `PT12H` type entries but not `D` as I quickly looked at it, which should be an easy fix. I have a few meetings back-to-back now but if anyone has time to fix that that would be great.

ashiklom (approved):
Looks good to me!

Issue #325, opened 2024-04-23
SE: Flexible swell slurm settings


rtodling:
With the perspective that there will be a Software Engineering oversight of Swell, let me start list issues that make up good items for SE to work on. 

One such thing is the slurm connection (definition of partition and special queue). The way it is not, Swell wires a single option. The fact is that most users have multiple projects, under different accounts, different qos, different partitions, etc. A more flexible - experiment related (with perhaps a default set, say, as things are presently) will be necessary in not too long.

ashiklom:
Tagging myself here. I will take a look at this.

rtodling:
Thanks ... you might want to tag yourself to #308 also ... BTW: I have edited existing issue and placed an "SE:" identifier in front of them so we can more clearly know what's seen as system engineering issues. There might be at least one more, at this time ...

ashiklom:
Jotting down some notes here while they're fresh on my mind:

- SWELL SLURM configuration happens here. Looks like there's a `~/.swell/swell-slurm` file that can be used for user-specific SLURM configuration. There are also some hard-coded values in here that can be modified. https://github.com/GEOS-ESM/swell/blob/cebcb49b4008485bff0456ec346692bc9cea6367/src/swell/deployment/create_experiment.py#L437-L507
- Cycl allows SLURM directives to be passed via its config files ([link](https://cylc.github.io/cylc-doc/stable/html/user-guide/task-implementation/job-runner-handlers/cylc.flow.job_runner_handlers.slurm.html#module-cylc.flow.job_runner_handlers.slurm)). We can teach SWELL to pass directives from YAML into the cylc `[[directives]]` block.

Dooruk:
Could we move this discussion to #308 please. I don't see any difference so we can keep them in the same place? 🙂

ashiklom:
Closing issue as duplicate of #308.

PR #339, opened 2024-05-09
Flexible SLURM configuration

### LINKED ISSUE #308: SE: SLURM Nodes/Tasks wired in SWELL
This is to start the discussion for a generalization of the slurm settings in SWELL. These are the issues related to this:

1) All tasks are associated w/ a single slurm option.

2) all tasks are wired to use a single node - which not enough to run 3dvar_atoms with the full observing system at the small c90 case.

3) The wiring is such that preparing a test case at other resolutions can only be done by hand.

The above opens the can of worms of the need to generalize how these options are set. The slum defaults need to be (i) task dependent; (i) model resolution dependent; (ii) analysis resolution dependent; etc.

ashiklom:
- **Refactor SLURM directives**
- **Support model-specific SLURM directives**
- **Add SLURM unit tests**

## Description

Support passing arbitrary SLURM configurations via ~`experiment.yaml`.~ new option -- `-s <slurmfile.yaml>` (e.g., `swell create 3dvar -s myslurm.yaml`.

The general idea is that more _specific_ and _explicit_ configurations always override _generic_ and _implicit_ (hard-coded) values.
Any values set in the ~`experiment.yaml`~ `<slurmfile.yaml>` override corresponding hard-coded defaults in the source code, so the ~`experiment.yaml`~ `<slurmfile.yaml>` can be used to exert _complete_ control over _every possible_ SLURM directive.

The structure of the ~`experiment.yaml`~ `<slurmfile.yaml>` SLURM directives is as follows:

```yaml
# slurmfile.yaml

# Global user-specified default values. These apply to all tasks, and override
# all hard-coded values (including task-specific ones).
slurm_directives_global:
  account: x1234
  nodes: 1

# Task-specific settings. These always override the globals (above) and any
# hard-coded values.
slurm_directives_tasks:
  RunGEOSExecutable:
    # Values specific to this task for *all* models.
    all:
      nodes: 2   # <-- This implies nodes=2 for RunGEOSExecutable-geos_ocean
    # Overrides specifically for `RunGEOSExecutable-geos_atmosphere`
    geos_atmosphere:
      nodes: 4
```

Note also: This also allows _any valid (or invalid!) SLURM directive_ to be passed to the configuration.

Inheritance logic and default values are embedded in the `src/swell/utilities/slurm.py` file.
The current way to add new default configurations is to modify that file.

## Dependencies

None

## Impact

Closes #308 and #325.

### DIFF
--- a/src/swell/utilities/slurm.py
+++ b/src/swell/utilities/slurm.py
@@ -37,9 +37,9 @@ def prepare_scheduling_dict(logger, experiment_dict):
     # Global SLURM settings from experiment dict (questionary / overrides YAML)
     # ----------------------------------------------
     experiment_globals = {}
-    if "slurm_directives_common" in experiment_dict:
+    if "slurm_directives_global" in experiment_dict:
         logger.info(f"Loading additional SLURM globals from experiment dict")
-        experiment_globals = experiment_dict["slurm_directives_common"]
+        experiment_globals = experiment_dict["slurm_directives_global"]
 
     # Task-specific SLURM settings from experiment dict (questionary / overrides YAML)
     # ----------------------------------------------
--- a/src/swell/utilities/slurm.py
+++ b/src/swell/utilities/slurm.py
@@ -17,7 +17,6 @@ def prepare_scheduling_dict(logger, experiment_dict):
     global_defaults = {
         "account": "g0613",
         "qos": "allnccs",
-        "partition": None,
         "nodes": 1,
         "ntasks-per-node": 24,
         "constraint": "cas|sky"
--- a/src/swell/utilities/slurm.py
+++ b/src/swell/utilities/slurm.py
@@ -20,8 +20,8 @@ def prepare_scheduling_dict(logger, experiment_dict):
     }
 
     task_defaults = {
-        "RunJediVariationalExecutable": { "nodes": 3, "ntasks_per_node": 36},
-        "RunJediUfoTestsExecutable": {"ntasks_per_node": 1}
+        "RunJediVariationalExecutable": {"all": {"nodes": 3, "ntasks_per_node": 36}},
+        "RunJediUfoTestsExecutable": {"all": {"ntasks_per_node": 1}}
     }
 
     # Global SLURM settings stored in $HOME/.swell/swell-slurm.yaml
@@ -49,7 +49,7 @@ def prepare_scheduling_dict(logger, experiment_dict):
 
     # List of tasks using slurm
     # -------------------------
-    slurm_tasks = [
+    default_slurm_tasks = {
         'BuildJedi',
         'BuildGeos',
         'EvaObservations',
@@ -59,25 +59,52 @@ def prepare_scheduling_dict(logger, experiment_dict):
         'RunJediLocalEnsembleDaExecutable',
         'RunJediUfoTestsExecutable',
         'RunJediVariationalExecutable',
-        'RunGeosExecutable',
-        ]
+        'RunGeosExecutable'
+        }
 
-    # Ap
+    experiment_slurm_tasks = set(experiment_task_directives.keys())
+    slurm_tasks = default_slurm_tasks.union(experiment_slurm_tasks)
+
+    model_components = experiment_dict["model_components"]
+
     scheduling_dict = {}
     for slurm_task in slurm_tasks:
         # Priority order (first = highest priority)
-            # 1. Task-specific directives from experiment (experiment_task_directives)
-            # 2. Global directives from experiment (experiment_globals)
-            # 3. Directives from user config (user_globals)
-            # 4. Hard-coded task-specific defaults (task_defaults)
-            # 5. Hard-coded global defaults (global_defaults)
-        # NOTE: Hard-code "job-name" to SWELL task here...
+            # 1. Task- *and* model-specific directives from experiment
+            #    (experiment_task_directives[slurm_task][model_component])
+            # 2. Task-specific (model-generic) directives from experiment
+            #    (experiment_task_directives[slurm_task]["all"])
+            # 3. Global directives from experiment (experiment_globals)
+            # 4. Directives from user config (user_globals)
+            # 5. Hard-coded task-specific defaults (task_defaults)
+            # 6. Hard-coded global defaults (global_defaults)
+        # NOTE: Hard-code "job-name" to SWELL task here but it can be
+        # overwritten in task-specific directives.
         directives = {"job-name": slurm_task, **global_defaults, **user_globals, **experiment_globals}
         if slurm_task in task_defaults:
             directives = {**directives, **task_defaults[slurm_task]}
         if slurm_task in experiment_task_directives:
-            directives = {**directives, **experiment_task_directives[slurm_task]}
-        scheduling_dict[slurm_task]["directives"] = directives
+            if "all" in experiment_task_directives[slurm_task]:
+                directives = {
+                    **directives,
+                    **experiment_task_directives[slurm_task]["all"]
+                }
+        # Set model_agnostic directives
+        scheduling_dict[slurm_task] = {"directives": {"all": directives}}
+        # Now, for every model component, set the model-generic directives
+        # (`directives`) but overwrite with model-specific directives if
+        # present.
+        for model_component in model_components:
+            model_specific_directives = {}
+            if (slurm_task in experiment_task_directives) and \
+                    (model_component in experiment_task_directives[slurm_task]):
+                model_specific_directives = experiment_task_directives[slurm_task][model_component]
+            model_directives = {
+                "job-name": f"{slurm_task}-{model_component}",
+                **directives,
+                **model_specific_directives
+            }
+            scheduling_dict[slurm_task]["directives"][model_component] = model_directives
 
     return scheduling_dict
 
--- a/src/swell/utilities/slurm.py
+++ b/src/swell/utilities/slurm.py
@@ -72,7 +72,9 @@ def prepare_scheduling_dict(logger, experiment_dict):
     assert len(non_slurm_tasks) == 0, \
         f"The following tasks cannot use SLURM: {non_slurm_tasks}"
 
-    model_components = experiment_dict["model_components"]
+    model_components = experiment_dict["model_components"] \
+        if "model_components" in experiment_dict \
+        else []
 
     scheduling_dict = {}
     for slurm_task in slurm_tasks:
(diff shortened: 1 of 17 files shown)


Dooruk (dismissed):
Not sure if it was "swell create ready but" SLURM needs all of the original attributes it seems. After running `swell create 3dvar`, I received the following error:

`jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'execution_time_limit'`

I really like the utility here, this will be a major update to current setup!

Is the idea to host some of these default dictionaries under Swell/platforms (Discover SLES15 vs. SLES12, or even AWS etc.) or just host everything under `swell-slurm.yaml` externally and keep hard coded ones in the `slurm.py`?

ashiklom:
Sorry, this PR was a bit premature. There are still a few bugs I need to fix.

I'll drop the "Draft" tag once this is ready for formal review (though comments are welcome at any time!).

> Is the idea to host some of these default dictionaries under Swell/platforms (Discover SLES15 vs. SLES12, or even AWS etc.) or just host everything under swell-slurm.yaml externally and keep hard coded ones in the slurm.py?

We should think through this for future development. The most obvious next step is to add Questionary questions for SLURM customization, and then maybe to add some Questionary presets for common cases. Now that all of the SLURM logic is in Python, we can do lots of sophisticated stuff with hueristics based on the experiment configuration, automatic detection of the current system (and maybe even customization based on current system load), etc.

ashiklom:
OK, I think this is ready for review. NOTE that this initial pass is a little bit hacky in its implementation:

1. I added a new `-s / --slurm somefile.yaml` option to `swell create` that reads the SLURM config from a special YAML file. (This part of the implementation I'm OK with).
2. Currently, the SLURM information from (1) is _appended_ to the `experiment_dict` (and ultimately, to `experiment.yaml`) as a _separate_ step _after_ all the logic producing `experiment.yaml`. That might be OK, at least for now, but it feels kludgy. We should either
  a. Commit to storing SLURM stuff in `experiment_dict` / `experiment.yaml` and add some Questionary stuff around it; or...
  b. Commit to SLURM stuff being a completely separate file/object and pass it around separately in the SWELL code.

I'm slightly inclined towards (2) here, not least because it seems trying to formulate this in terms of Questionary prompts sounds really complicated, and because SLURM settings seem more like a user- and platform-specific configuration that people might want to manage separately (e.g., `gitignore`) from the more shareable `experiment.yaml/overrides.yaml`. But you all have to use this --- what do you think?

src/swell/utilities/slurm.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def prepare_scheduling_dict(logger, experiment_dict):
+    # Hard-coded defaults
+    # ----------------------------------------------
+    global_defaults = {
+        "account": "g0613",
+        "qos": "allnccs",
+        "partition": None,
+        "constraint": "cas|sky"
```
Dooruk, line 22: We need more of the default directives here unless we have more of the `global_defaults` to live in slurm.yaml located somewhere in Swell.

Dooruk (commented):
Thanks for working on this @ashiklom! This is a good step that would allow us to implement layout and model resolution dependencies and set defaults for them.

I'm voting for 2.b) Commit to SLURM stuff being a completely separate file/object and pass it around separately in the SWELL code.

I think Questionary (for JEDI config) and slurm.yamls require different logic.

Minor comments while testing, using `swell create` without the `-s` flag created  slurm directives missing some of the important components not defined in global defaults (nodes and ntasks per_node). So some slurm tasks failed as a result. I guess the fix will be contingent upon how we want to proceed with storing `slurm.yaml`.

another very minor comment, `--job-name` used to have model names,  e.g., `EvaObservations-geos_ocean`, it would be nice to keep that if possible. Not super critical though.

ashiklom:
> I'm voting for 2.b) Commit to SLURM stuff being a completely separate file/object and pass it around separately in the SWELL code.

Agreed!

> Minor comments while testing, using swell create without the -s flag created slurm directives missing some of the important components not defined in global defaults (nodes and ntasks per_node). So some slurm tasks failed as a result. I guess the fix will be contingent upon how we want to proceed with storing slurm.yaml.

Thanks for catching this. I'll look for where these were defined previously and try to add them back in. The `-s slurm.yaml` should be completely optional and this PR shouldn't break old workflows, so I'll need to fix that before this gets merged.

Can you point me to which specific tasks need this?

> another very minor comment, --job-name used to have model names, e.g., EvaObservations-geos_ocean, it would be nice to keep that if possible. Not super critical though.

This should work (https://github.com/GEOS-ESM/swell/pull/339/files#diff-68c8167035a32267716d03f8e6a0ce35fe12d2368732e07fabf202cc008a9e74R118-R120). Not sure why that's being ignored. I'll investigate.

Dooruk:
> 
> Can you point me to which specific tasks need this?
> 

As far as I'm aware, every SLURM task needs node information. Tasks per node might be optional but that goes hand to hand with the model layout. If MOM6 uses a 24x30 layout, I dedicate 20 notes and 36 tasks per node.

> This should work (https://github.com/GEOS-ESM/swell/pull/339/files#diff-68c8167035a32267716d03f8e6a0ce35fe12d2368732e07fabf202cc008a9e74R118-R120). Not sure why that's being ignored. I'll investigate.

Hmm, you are right. wonder if `-` has a specific meaning in templating?

src/swell/utilities/slurm.py
ashiklom: Ah, whoops --- I just accidentally skipped 2 lines in the original code specifying the nodes and tasks per node. Fixed in 69cb0a3.

ashiklom:
When I run `swell create 3dvar`, I get `flow.cylc` files that set the job-name correctly (with both task name and model component); for example:

```
    [[EvaObservations-geos_ocean]]
        script = "swell task EvaObservations $config -d $datetime -m geos_ocean"
        platform = nccs_discover
        execution time limit = PT30M
        [[[directives]]]
            --job-name = EvaObservations-geos_ocean
            --account = g0613
            --qos = allnccs
            --partition = None
            --nodes = 1
            --ntasks_per_node = 24
            --constraint = cas|sky
```

Can you confirm that you're not getting this, and if so, send me the command you're running so I can debug?

Dooruk:
Ok I tested it just now again, the slurm override functionality works. And the default produced:

```
    [[GenerateBClimatology-geos_ocean]]
        script = "swell task GenerateBClimatology $config -d $datetime -m geos_ocean"
        platform = nccs_discover
        execution time limit = PT1H
        [[[directives]]]
            --job-name = GenerateBClimatology-geos_ocean
            --account = g0613
            --qos = advda
            --partition = None
            --nodes = 1
            --ntasks_per_node = 24
            --constraint = cas|sky

    [[RunJediVariationalExecutable-geos_ocean]]
        script = "swell task RunJediVariationalExecutable $config -d $datetime -m geos_ocean"
        platform = nccs_discover
        execution time limit = PT1H
        [[[directives]]]
            --job-name = RunJediVariationalExecutable-geos_ocean
            --account = g0613
            --qos = advda
            --partition = None
            --nodes = 3
            --ntasks_per_node = 36
            --constraint = cas|sky
```

Dooruk (changes requested):
Ok, I caught two things:

1) It is `ntasks-per-node` not `ntasks_per_node`
2) So looks like `partition` needs special handling. My Slurm submission failed because `None` is invalid. Below is from the original code, `partition` is only written in the `flow.cylc` if it exists. Perhaps we can add a final check before returning `target_dict` and erase `partition` keys if they are `None`? Idk if there is a better solution.

```
{% if scheduling["BuildGeos"]["partition"]
--partition={{scheduling["BuildGeos"]["partition"]}}
{% endif %}
```

Dooruk:
I had to make changes to `ntasks-per-node` locally during my runs on another branch and I just noticed that original `swell_create_experiment.py` also has it as `ntasks_per_node`. Apparently, that gets accessed to `ntasks-per-node` through jinja2 handling.

From the original code:
```
            --ntasks-per-node={{scheduling["BuildJedi"]["ntasks_per_node"]}}

**Comment by ashiklom** [comment]
@Dooruk Thanks for catching all of that! Sorry for the sloppiness. I'll try to test things more thoroughly before submitting next time. You will have to show me your preferred approach to running some quick SWELL tests without launching massive GEOS simulations, so I can catch some of these things myself before submitting PRs.

`partition` issue is addressed in c670067 --- I just removed it from the global defaults. `nodes-per-task` fixed in 1890a54, and I added a validation step in a4d9a73.

**Comment by Dooruk** [comment]
> @Dooruk Thanks for catching all of that! 

Happy to help. I ran your branch (after making a local copy) via Tier 1 tests and some failed, swell code tests are also failing. Let's stay on after the CI meeting tomorrow and I can walk you through what I do briefly.

https://github.com/GEOS-ESM/swell/actions/runs/9118250195

**Comment by Dooruk** [comment]
Hi @ashiklom , I just ran this branch again and tier1 tests [worked this time](https://github.com/GEOS-ESM/swell/actions/runs/9275605961). I didn't touch anything so perhaps something was messed up with the github runners the previous time?

The code tests are failing though, so since this is on your personal repo you need to fix that, if you don't mind. Thanks!

**Comment by ashiklom** [comment]
> tier1 tests [worked this time](https://github.com/GEOS-ESM/swell/actions/runs/9275605961)

I think this is because the previous tests were broken by a JEDI update (hence #345) that has since been resolved.

> The code tests are failing though

My unit tests worked as expected! I just forgot to update them. But good to know they throw errors in situations when there should be an error! Should be fixed now.

File: src/swell/utilities/slurm.py [hunk 2]
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+    # Throw an error if a user tries to set SLURM directives for a task that
+    # doesn't use SLURM.
+    experiment_slurm_tasks = set(experiment_task_directives.keys())
+    non_slurm_tasks = experiment_slurm_tasks.difference(slurm_tasks)
+    assert len(non_slurm_tasks) == 0, \
+        f"The following tasks cannot use SLURM: {non_slurm_tasks}"
+
+    model_components = experiment_dict["model_components"]
```
Dooruk: For suites like `build_jedi` or `build_geos`, there are no model components in the exp. dictionary. Hence this is causing problem for those suites...

Dooruk: This following works:

```
    # Check if experiment_dict has model components
    # ----------------------------------------------
    model_components = []
    if "model_components" in experiment_dict:
        model_components = experiment_dict["model_components"]
```

Dooruk (commented):
Ok, tier 1 tests work, we are really close.

`model_components` is not in every experiment dict. so that is the final issue before we merge this, my mod should fix it.

src/swell/utilities/slurm.py
ashiklom: Fixed in 2db2741.

Dooruk (approved):

Dooruk:
Thanks, all looks good!

PR #348, opened 2024-06-05
add co2 to atmos background

rtodling:
## Description

This makes JEDI background a little more consistent w/ GEOS/GSI background - adding CO2 as part of it.

I don't expect significant changes from adding this, but some obs could be knocked in or out.

## Dependencies

None

## Impact

Should allow for geoval comps w/ GSI - w/ caution of level reversal.


gmao-jjin3 (approved):

Dooruk (approved):
Nothing broke in tier 1, merging..

PR #350, opened 2024-06-07
Use GEOS Experiment Backgrounds

Dooruk:
@rtodling wants to be able to tap into GEOS Experiment directories (x0048, x0049) for cycling 3DVar experiments. So one can request other experiments, depending on the background availability. The updated task allows that and it is used in the flow.cylc file as well. We can always revert it back to R2D2 at any point.

I tested with `3dvar_atmos` and `3dfgat_atmos` suites and there are some differences in the increment plots. Please take a look before we merge this. See `q_500` increment below:

inc_q_500


rtodling:
@Dooruk GEOS-FP!  Now you made me puzzled ... first let me ask you, where did the background for the picture above came from? 

Now, assuming all that is ok - I would say we should not worry too much about Q: if you recall, I have showed results about Q and I have pointed to all that the issue is related to a buggy interpolation of the background fields when saber takes cubed fields into gsibec. This is associated w/ the OOPS interpolations and their inconsistency w/ ATLAS interpolations - that's how I interpret the issue. 

But the increments of T and U/V and O3 should look decent ... 

To me the key question is: where did the background come from?

src/swell/deployment/platforms/nccs_discover/task_questions.yaml
```diff
@@ -13,6 +13,9 @@ existing_jedi_build_directory:
 existing_jedi_source_directory:
   default_value: /discover/nobackup/gmao_ci/swell/tier2/stable/build_jedi/jedi_bundle/source/
 
+geos_fp_background_directory:
```
rtodling: why fp?! this should just be geos_background_directory - no?

src/swell/deployment/platforms/generic/task_questions.yaml
```diff
@@ -13,6 +13,9 @@ existing_jedi_build_directory:
 existing_jedi_source_directory:
   default_value: This would need to be provided by user
 
+geos_fp_background_directory:
+  default_value: This would need to be provided by user
```
rtodling: I'm confused about this: this is a directory ... but it should be a templated filename ... right now, when I set up 3dvar_atmos (using this branch), I get in the experiment.yaml:

geos_fp_background_directory: /gpfsm/dnb05/projects/p139/dao_it/archive/

but this should really be:

geos_fp_background_directory: /gpfsm/dnb05/projects/p139/dao_it/archive//x0048/rs/Y%Y/M%M/x0048.bkgcrst.%Y%M%D_%H00z.tar

which also means, the variable name should be more like:

geos_exp_background_filename:

Dooruk:
Just FYI, this is a draft, I will make a few more changes.

Dooruk:
This essentially provides an option for directly accessing GEOS X experiments background located at:

`/gpfsm/dnb05/projects/p139/dao_it/archive/`

rtodling (approved):
This works like a charm.

PR #352, opened 2024-06-10
Refactor SLURM user global defaults for test

### LINKED ISSUE #351: SE: Minor code tests bug with SLURM config
`Test_slurm_config` code test works on Github actions, fails locally when `qos` is modified for a user on Discover. I guess that is fine until we move `swell-slurm.yaml` somewhere else?

Screenshot 2024-06-10 at 11 26 46 AM

ashiklom:
Use a mock (https://docs.python.org/3/library/unittest.mock.html) to "fake" the behavior of the new global user defaults funciton in unit tests to always produce consistent behavior.

Closes #351.

### DIFF
--- a/src/swell/test/code_tests/slurm_test.py
+++ b/src/swell/test/code_tests/slurm_test.py
@@ -11,6 +11,8 @@
 import unittest
 import logging
 
+from unittest.mock import patch
+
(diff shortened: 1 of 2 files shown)


Dooruk (approved):
Looks good, thanks!

PR #354, opened 2024-06-13
Add SLES15 as a platform

### LINKED ISSUE #355: Add SLES15 as a platform
- [x] Code changes and home directory updates
- [x] #356

Dooruk:
It is finally happening....

... this will require some changes in the `.swell` and `.cylc` folders.


src/swell/utilities/slurm.py
```diff
@@ -7,25 +7,31 @@
...
-        "nodes": 1,
-        "ntasks-per-node": 24,
-        "constraint": "cas|sky"
-    }
+
+    # Obtain platform-specific SLURM directives and set them as global defaults
+    # -----------------------------------------------------------------------
+    swell_lib_path = get_swell_path()
+    platform_path = os.path.join(swell_lib_path, 'deployment', 'platforms', platform)
+    platform_slurm = os.path.join(platform_path, "slurm.yaml")
```
ashiklom: This is reading YAML files from inside the currently loaded SWELL repository, right? If so, the "correct" and probably more reliable approach here is to use `importlib.resources` (https://importlib-resources.readthedocs.io/en/latest/using.html).

That said, since we use this pattern elsewhere in SWELL and presumably this does work, I'm OK leaving this as is for now and just opening an issue to replace this construct with `importlib.resources`. I'll leave that up to you.

Dooruk: Ok, I used `importlib.resources` here but not sure if that's what you had in mind.

Dooruk:
I'm proposing not using `.swell/slurm-swell.yaml` and moving it under `platforms` (which is already done but I forgot to explain my rationale), so the user would need to use the `-s` option unless we set the `account` etc. NCCS Discover generic.

Just need to fix the mock test next..

Dooruk:
@ashiklom this is ready for a review. I commented out `user_globals`, so I'm proposing having `global_defaults` in `platforms`, nothing in the `$HOME` directory (now that we have them in `platforms`, no need for `user_globals`), with the `-s` option for the user during `swell create`. Hence, I had to comment out lines from the `code_tests` please take a look when you have a chance.

I had to introduce `platform` in both `code_test` and the `slurm` utility but we might need a more elegant solution..

src/swell/utilities/slurm.py
```diff
@@ -15,18 +17,25 @@
...
+    # -------------------------------------------
+    platform_path = f"swell.deployment.platforms.{platform}"
+
+    # Import the path dynamically
+    # ------------------------------
+    path_import = __import__(platform_path, fromlist=[''])
+
+    logger.info(f'Loading SLURM user configuration for the "{platform}" platform')
+    with importlib.resources.open_text(path_import, 'slurm.yaml') as yaml_file:
+        global_defaults = yaml.safe_load(yaml_file)
```
ashiklom, line 41: Design question: Is internal YAML the right place to store this information? Personally, I am inclined to store hard-coded defaults directly in Python, which simplifies the code and reduces the number of places people have to look. E.g., If I'm coming into this cold and trying to understand the platform argument, I have to follow the following jumps: `platform` argument --> `platform_path` --> `path_import` --> "oh, I guess that means `swell/deployment/platforms/<platform>/slurm.yaml`" --> navigate to a separate directory and read the YAML file.

I would be inclined to just have a big dict that lives directly in this file:

```python
platforms = {
  "nccs_discover_sles15": {
    "account":"xxxx",
    <...>,
  "nccs_discover_sles14": {
    "account": "yyyy",
    <...>
  }
}
```

Counterpoint: The main advantage of the current design is that all the platform information --- including SLURM --- lives in the same directory. That makes more sense if you are working from a platform-based standpoint ("I need to make SWELL work on a new system") rather than a functionality-based standpoint ("I need to improve the way that SWELL does SLURM stuff"). (Even then, I still think that it would be generally better design to have fewer YAML files and more logic coded directly in Python via classes, which can also handle much more sophisticated logic, are easier to validate, can do inheritance, etc.).

Food for thought. The approach here works, so we can leave this as is for now and revisit later.

src/swell/utilities/slurm.py
```diff
@@ -6,7 +6,9 @@
 
 # --------------------------------------------------------------------------------------------------
 
-import os
+import importlib.resources
```
ashiklom: To expose some additional `importlib` functionality:

```suggestion
import importlib
```

src/swell/utilities/slurm.py
```diff
@@ -15,18 +17,25 @@
...
-    }
 
+    # Obtain platform-specific SLURM directives and set them as global defaults
+    # Start by constructing the full platforms path
+    # -------------------------------------------
+    platform_path = f"swell.deployment.platforms.{platform}"
+
+    # Import the path dynamically
+    # ------------------------------
+    path_import = __import__(platform_path, fromlist=[''])
```
ashiklom: [Best practice](https://docs.python.org/3/reference/import.html#importlib) is to use `importlib.import_module` instead of `__import__` for this.

Also, good opportunity for better error handling.

```suggestion
    try:
        path_import = importlib.import_module(platform_path)
    except ModuleNotFoundError:
        raise Exception(f"Platform '{platform}' has not been configured in SWELL")
    except Exception as err:
        raise err
```

src/swell/utilities/slurm.py
```diff
@@ -185,17 +200,17 @@ def validate_directives(directive_dict):
...
+#     logger: Logger,
+#     yaml_path: str = "~/.swell/swell-slurm.yaml"
+# ) -> dict:
+#     yaml_path = os.path.expanduser(yaml_path)
+#     user_globals = {}
+#     if os.path.exists(yaml_path):
+#         logger.info(f"Loading SLURM user configuration from {yaml_path}")
+#         with open(yaml_path, "r") as yaml_file:
+#             user_globals = yaml.safe_load(yaml_file)
+#     return user_globals
```
ashiklom: No need to keep this around if we're not using it. That's what git is for!

```suggestion
```

src/swell/test/code_tests/slurm_test.py
```diff
@@ -8,10 +8,11 @@
 # --------------------------------------------------------------------------------------------------
 
 
-import unittest
 import logging
+import platform as pltfrm
+import unittest
 
-from unittest.mock import patch
+# from unittest.mock import patch
```
ashiklom: I would delete all the commented code in here (and elsewhere). That's what git is for!

src/swell/test/code_tests/slurm_test.py
```diff
@@ -50,13 +51,22 @@ def test_slurm_config(self, mock_global_defaults):
...
+        # prepare_scheduling_dict requires consistent platform input
+        # This is a temporary solution until we have a more robust way
+        # ----------------------------------------------------------
+        current_platform = "nccs_discover"
+
+        # Check if platform contains Linux-5.14.21, which indicates platform is SLES15
+        if 'Linux-5.14.21' in pltfrm.platform():
+            current_platform = "nccs_discover_sles15"
+
+        sd = prepare_scheduling_dict(logger, experiment_dict, platform=current_platform)
```
ashiklom: This makes the unit tests platform specific, which is not how this should work for multiple reasons.

What I would really do here is define (simple) unit tests for every supported platform; e.g.,

```python
# Platform-specific definitions and tests
sd_nccs_discover = prepare_scheduling_dict(logger, experiment_dict, platform = "nccs_discover")
assert sd_nccs_discover[...]["directives"]["account"] = "XXX"
sd_sles15 = prepare_scheduling_dict(logger, experiment_dict, platform="nccs_discover_sles15")
assert sd_sles15[...]["directives"]["account"] = "YYYY"

# Platform-agnostic unit tests
for sd in (sd_nccs_discover, sd_sles15):
    for mc in [...]:
        assert sd[...]

# Finally, check that passing an invalid platform throws an error
with self.assertRaises(ModuleNotFoundError):
    prepare_scheduling_dict(logger, experiment_dict, platform = "not_a_platform"
```

ashiklom: Also, now that I'm actually reading the `unittest` docs, it seems like we should be using `self.assertEqual` and similar methods rather than built-in Python assertions (probably for better error reporting).

E.g.:

```python
# Replace...
assert sd_nccs_discover[...]["directives"]["account"] = "XXX"
# ...with...
self.assertEqual(sd_nccs_discover[...]["directives"]["account"], "XXX")
```

src/swell/test/code_tests/slurm_test.py
```diff
@@ -22,11 +23,11 @@ class SLURMConfigTest(unittest.TestCase):
...
     # Mock the `slurm_global_directives` function to ignore "real"
     # configuration.
-    @patch("swell.utilities.slurm.slurm_global_defaults")
-    def test_slurm_config(self, mock_global_defaults):
+    # @patch("swell.utilities.slurm.slurm_global_defaults")
+    def test_slurm_config(self):
 
         # Fake user-specified global values (for consistent unit tests)
-        mock_global_defaults.return_value = {"qos": "dastest"}
+        # mock_global_defaults.return_value = {"qos": "dastest"}
```
ashiklom: Ditto -- delete commented code.

src/swell/test/code_tests/slurm_test.py
```diff
@@ -50,13 +51,22 @@ def test_slurm_config(self, mock_global_defaults):
...
+        sd = prepare_scheduling_dict(logger, experiment_dict, platform=current_platform)
 
         for mc in ["all", "geos_atmosphere", "geos_ocean"]:
-            # Hard-coded global defaults
-            assert sd["EvaObservations"]["directives"][mc]["constraint"] == "cas|sky"
+            # Hard-coded global defaults (constraint is platform-specific)
+            # assert sd["EvaObservations"]["directives"][mc]["constraint"] == "cas|sky"
             # Global user-specific defaults (NOTE: mocked above!)
-            assert sd["EvaObservations"]["directives"][mc]["qos"] == "dastest"
+            # assert sd["EvaObservations"]["directives"][mc]["qos"] == "dastest"
```
ashiklom: Delete commented code.

src/swell/utilities/slurm.py
```diff
@@ -36,7 +45,13 @@ def prepare_scheduling_dict(
     # ----------------------------------------------
     # NOTE: Separate function to allow it to be mocked in unit tests.
     # See https://github.com/GEOS-ESM/swell/issues/351
-    user_globals = slurm_global_defaults(logger)
+    # user_globals = slurm_global_defaults(logger)
```
ashiklom, line 48: Delete comment and code.

ashiklom: Delete comment and code.

src/swell/utilities/slurm.py
```diff
@@ -92,7 +107,7 @@ def prepare_scheduling_dict(
         directives = {
             "job-name": slurm_task,
             **global_defaults,
-            **user_globals,
+            # **user_globals,
```
ashiklom:
```suggestion
```

src/swell/utilities/slurm.py
```diff
@@ -139,7 +154,7 @@ def prepare_scheduling_dict(
                 )
             model_directives = {
                 **model_directives,
-                **user_globals,
+                # **user_globals,
```
ashiklom:
```suggestion
```

src/swell/utilities/slurm.py
```diff
@@ -36,7 +45,13 @@ def prepare_scheduling_dict(
...
     # NOTE: Separate function to allow it to be mocked in unit tests.
     # See https://github.com/GEOS-ESM/swell/issues/351
-    user_globals = slurm_global_defaults(logger)
+    # user_globals = slurm_global_defaults(logger)
+
+    # Check if platform contains Linux-5.14.21, which indicates platform is SLES15
+    if 'Linux-5.14.21' in pltfrm.platform():
+        current_platform = "nccs_discover_sles15"
+        assert platform == current_platform, \
+            f"Inconsistent platfrom choice: {platform} vs. {current_platform}"
```
ashiklom: Suggest a more informative error message:

```suggestion
        assert platform == "nccs_discover_sles15", (
            "'Linux-5.14.21' detected, which implies platform 'nccs_discover_sles15. " +
            f"That is inconsistent with user-specified platform '{platform}'."
        )
```

src/swell/utilities/slurm.py
Dooruk: Wanted to hear your take before I take stuff out, it's the scientist in me :)

src/swell/utilities/slurm.py
Dooruk: It makes intuitive sense to me to keep that information at the platform level but I'm not adamant about it.

In the next iteration, we should also define defaults for different resolutions somewhere, to be used for different tier tests.

src/swell/test/code_tests/slurm_test.py
Dooruk: ok, I had to pass an extra bool in `prepare_scheduling_dict` for this to work

Dooruk:
I had to add a `unit_test` switch to `prepare_scheduling_dict` to allow platform catch conditional to work but looks good on my end.

src/swell/utilities/slurm.py
```diff
@@ -6,19 +6,20 @@
...
 
 from swell.utilities.logger import Logger
+from importlib import resources
 
 
 def prepare_scheduling_dict(
     logger: Logger,
     experiment_dict: dict,
     platform: str,
+    unit_test: bool = False
```
ashiklom: The correct way to do this is to `mock` the output of `pltform.platform()` in the unit tests.

First, remove the `unit_test` from this file.

Then, in the unit test file, add the following:

```python
from unittest.mock import patch
```

Decorate and change the definition of `test_slurm_config` as follows:

```python
    @patch("platform.platform")
    def test_slurm_config(self, platform_mocked):
```

This combination says that the new argument to this function -- `platform_mocked` is a mocked version of the function `platform.platform`, which allows us to control how downstream code uses this function. What that means is that we can now fully control the behavior of `platform_mocked` and, by extension, what the code _thinks_ is `platform.platform`.

Further down in the unit test, make the following change:

```python
        platform_mocked.return_value = "Linux-4.12.14"  # Old Discover version
        sd_discover = prepare_scheduling_dict(logger, experiment_dict,
                                              platform="nccs_discover")
        self.assertEqual(sd_discover["RunJediVariationalExecutable"]["directives"]["all"]
                         ["constraint"], "cas|sky")
```

This is a confusing design pattern that took me some time to figure out (in my last PR), but what's happening here is: `platform_mocked.return_value = "Linux-4.12.14"` (combined with the `@patch("platform.platform")` decorator above) makes it so that when any code used by this function sees `platform.platform()`, instead of _actually_ calling that function, it calls a "mocked" version that always returns `"Linux-4.12.14".

We repeat this pattern in the next block to again modify the fake (mocked) version of `platform.platform` to _now_ return `"Linux-5.14.21"`.

```python
        platform_mocked.return_value = "Linux-5.14.21"  # Discover milan version
        sd_discover_sles15 = prepare_scheduling_dict(logger, experiment_dict,
                                                     platform="nccs_discover_sles15")
        self.assertEqual(sd_discover_sles15["RunJediVariationalExecutable"]["directives"]
                         ["all"]["constraint"], "mil")

        with self.assertRaises(AssertionError):
            prepare_scheduling_dict(logger, experiment_dict,
                                    platform="nccs_discover")

```

The new block _after_ that tests for the error we expect --- while `platform.platform()` is still "mocked" to always spit out a specific version of Linux, we can now test that, under those conditions, we get the right error.

Mocking provides a really effective way to test how code like this would run under different circumstances. This is a really common design pattern for, e.g., web APIs --- e.g., we can also "mock" web functions to return a particular kind of error or status code to check that our code handles those errors the way we want. In my earlier code, it also provided a good way to test the behavior of our code assuming different global configuration files (without even having to create those files).

Dooruk: OK, `patch` to `platform_mocked` connection didn't make intuitive sense to me, which I should've searched online, but it's much more clear now. Really appreciate the long comment and explanation 👍 

Thanks for being "assertive" for the mock test!

Dooruk:
Can I get that approval now 😄

Dooruk:
Hmm, since we are not using the global `swell-slurm.yaml` anymore, `gmao_ci` automatically uses the default settings. That is one minor issue but that's related to how our Github actions are setup currently.

But thinking about it further, this will impact other users too, unless they include this in their `-s slurm-override.yaml`.

ashiklom (approved):

ashiklom:
> Hmm, since we are not using the global `swell-slurm.yaml` anymore, `gmao_ci` automatically uses the default settings. That is one minor issue but that's related to how our Github actions are setup currently.
> 
> But thinking about it further, this will impact other users too, unless they include this in their `-s slurm-override.yaml`.

Yes, removing the `user_globals` / `~/.swell/swell-slurm.yaml` is a breaking change, and not necessarily a necessary one for this feature. I think keeping `user_globals` around is safer --- there are platform-agnostic but still useful ways to set the user-specific defaults (most importantly, setting your user account), and it can always be overridden by `-s slurm.yaml`, so it's easy to soft-deprecate it gradually (rather than ripping it out abruptly). It's also consistent with how other SLURM-related software does things; e.g., [Dask jobqueue](https://jobqueue.dask.org/en/latest/configuration.html).

Personally, my vote would be to bring `user_globals` back. But if you decide to leave it out, a simple fix to the CI would be to add `-s /home/gmao_ci/.swell/swell-slurm.yaml` to every `swell create` command in the CI.

Dooruk:
> Personally, my vote would be to bring `user_globals` back. But if you decide to leave it out, a simple fix to the CI would be to add `-s /home/gmao_ci/.swell/swell-slurm.yaml` to every `swell create` command in the CI.

Ok, I revived `user_globals` and now we have not just one but two mock tests!

PR #370, opened 2024-06-24
Towards WCDA: Introduce `geos_marine`, keep `geos_ocean`

### LINKED ISSUE #264: GEOS version update and iced.nc handling
title says it all

### LINKED ISSUE #265: Modifying JEDI files to handle CICE files
In terms of marine DA, there could be different DA configurations such as ocean only and ocean+sea-ice. This requires handling JEDI files differently according to the marine DA type.

For instance, in this [yaml file](https://github.com/GEOS-ESM/swell/blob/develop/src/swell/configuration/jedi/interfaces/geos_ocean/model/background.yaml) state variables would change to handle `cicen`, `hicen` etc. and `ice_filename` has to be defined on top of `ocn_filename`.

This involves observation yamls as well.

- [ ] `R2D2.yaml` and storing cice and MOM6 files

@danholdaway the time has come to have this discussion 😟

### LINKED ISSUE #388: Add `--no-requeue` flag to SLURM submissions
Received a yellow card from NCCS:

> "RunJediVariationalExecutable" job on Discover today has been running compute nodes out of memory, causing them to crash. The job re-queues and re-runs on new nodes after each failure, with the same result each time - this can be prevented with the Slurm job option: #SBATCH --no-requeue).

This happened to me while running `3dvar_cycle` on SLES15. While NCCS is figuring out what went wrong, we just need to add the flag hard-coded while submitting SLURM jobs.

..maybe a memory leak?

Dooruk:
This introduces `geos_marine` as a 3rd model interface and allows assimilating ocean and sea-ice data (first proper Weakly-Coupled DA with SWELL!). `geos_marine` will eventually replace `geos_ocean` but we can keep it around for now while we figure out ocean-only DA vs. marine DA changes. So that creates a lot of code duplication in terms of observation YAMLs under `geos_ocean` and `geos_marine` but that is temporary.

Some highlights:

- GEOSgcm version is bumped to v11.6.0, which required some changes in terms of Boundary Conditions linking. 
- We switched from using CICE4 to CICE6. As far as SWELL is concerned, this is a one line change in `run_geos_executable` as we point to the external GEOS builds and experiment directories.
- Introduced the new task SOCA2CICE variable change, which should only be active in `flow.cylc` while the sea-ice DA is active.
- `SOCA` is designed to handle different model types (ocean, sea-ice, BGC, waves), hence some of the configuration creation will depend on if these models are turned on or off. This is a longer term consideration but an important one as combinations of different model types are concerned.
- Some minor EVA related changes, mainly concerning projections for sea-ice related plots.
- Introduces `--no-requeue` `#SBATCH` option to prevent Cylc tasks from resubmitting jobs and "spreading" to other compute nodes.

Closes the following:
- https://github.com/GEOS-ESM/swell/issues/265
- https://github.com/GEOS-ESM/swell/issues/264
- https://github.com/GEOS-ESM/swell/issues/388


src/swell/tasks/eva_observations.py
```diff
@@ -112,17 +116,39 @@ def execute(self):
...
+                    eva_override['map_projection'] = 'spstere'
+
+            # # Check if the "passivate" condition exists within the "obs filters" list
+            passivate_exists = any(
+                filter_item.get('action', {}).get('name') == 'passivate'
+                for filter_item in observation_dict.get('obs filters', [])
+            )
+
+            if passivate_exists:
+                print("Condition 'passivate' exists in 'obs filters'")
```
ashiklom:
```suggestion
                self.logger.info("Condition 'passivate' exists in 'obs filters'")
```

src/swell/tasks/link_geos_output.py
```diff
@@ -80,5 +95,53 @@ def link_restart(self):
...
+        nj = np.shape(ds['aicen'])[1]
+        ni = np.shape(ds['aicen'])[2]
+
+        # populate xarray with aggregated quantities
+        # ------------------------------------------
+        aggds = xr.merge([xr.DataArray(
+                        name=varname,
+                        data=np.reshape(np.sum(ds[soca2cice_vars[varname]].values, axis=0),
+                                        (1, nj, ni)),
+                        dims=['time', 'yaxis_1', 'xaxis_1']) for varname in soca2cice_vars.keys()])
```
ashiklom: FYI: Strictly speaking, `xarray.merge` accepts any iterable as an input; it doesn't have to be a list. So, if you wanted to be slightly more pythonic, you could just drop the `[ ]` around this expression and it will still work (and may be _slightly_ more efficient because you don't have to create a list fist). This is called a [generator expression](https://docs.python.org/3/reference/expressions.html#generator-expressions). In practice, here, it should make practically no difference, but it's good practice to use them because there are other situations in which you can carve out _major_ efficiencies by skipping the eager coercion to `list`.

```python
aggds = xr.merge(xr.DataArray(...) for thing in stuff)
```

src/swell/tasks/link_geos_output.py
```diff
@@ -80,5 +95,53 @@ def link_restart(self):
 
         return src, dst
 
+    # ----------------------------------------------------------------------------------------------
+
+    def prepare_cice6(self):
```
ashiklom, line 100: Type hint (this syntax should be supported in Python 3.9+; if we need 3.8 or lower, we will need to do `from typing import Tuple` above and, here, `Tuple(int, int)`.

```suggestion
    def prepare_cice6(self) -> tuple[str, str]:
```

_Strictly_ speaking, it might be good to take advantage of Python 3's built-in `pathlib` for paths --- then, this would be more clearly `tuple[Path, Path]`, and we would be able to replace a bunch of `os.path.join(somepath, someotherpath)`, `os.path.exists(somepath)` with syntactic sugar like `somepath / someotherpath` (to join paths) and `somepath.exists()`. But I'm not going to make you do that here 😉  --- just flagging as food for thought for the future.

src/swell/tasks/move_da_restart.py
```diff
@@ -93,8 +93,15 @@ def cycling_restarts(self):
...
-        move_files(self.logger, self.forecast_dir('tile.bin'),
-                   self.at_next_fcst_dir('tile.bin'))
+        # Create a dictionary of src/dst for the single files
+        # ---------------------------------------------------
+        src_dst = {'tile.bin': '',
+                   'RESTART/iced.nc': 'INPUT',
+                   }
+
+        for src, dst in src_dst.items():
+            dst = os.path.join(dst, src.split('/')[-1])
```
ashiklom:
```suggestion
            dst = os.path.join(dst, os.path.basename(src))
```

Same thing, but communicates intent more clearly.

(Note that _both_ methods will return `''` if `src` has a trailing slash, like `src = "some/path/").

src/swell/tasks/move_forecast_restart.py
```diff
@@ -72,7 +72,15 @@ def cycling_restarts(self):
...
 
-        move_files(self.logger, self.forecast_dir('tile.bin'), self.at_next_fcst_dir('tile.bin'))
+        # Create a dictionary of src/dst for the single files
+        # ---------------------------------------------------
+        src_dst = {'tile.bin': '',
+                   'RESTART/iced.nc': 'INPUT',
+                   }
+
+        for src, dst in src_dst.items():
+            dst = os.path.join(dst, src.split('/')[-1])
```
ashiklom:
```suggestion
            dst = os.path.join(dst, os.path.basename(src))
```

src/swell/tasks/run_jedi_convert_state_soca2cice_executable.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+class RunJediConvertStateSoca2ciceExecutable(taskBase):
+
+    # ----------------------------------------------------------------------------------------------
+
+    def execute(self):
```
ashiklom:
```suggestion
    def execute(self) -> None:
```

src/swell/tasks/run_jedi_convert_state_soca2cice_executable.py
```diff
@@ -0,0 +1,117 @@
...
+
+            # Get the JEDI interface metadata
+            # -------------------------------
+            model_component_meta = self.jedi_rendering.render_interface_meta()
+
+            # Compute number of processors (only requires a single node and fails
+            # for multiple nodes, so we set it to 36 processors for now)
+            # ----------------------------------------------------------
+            np = 24
+            # np = eval(str(model_component_meta['total_processors']))
```
ashiklom: A few things:

1. The comment here is mismatched with the value (24 vs 36)
2. The value is a magic number that is buried in the middle of this function. If we have to hard-code it for now, I would follow PEP 8 and definite it as a clearly named constant in all caps (e.g., `N_PROCESSORS_DEFAULT=24`) near the top of the function, or maybe even near the top of the file.
3. Not strictly necessary for this PR, but it would be nice to define this in a way that it could be modified by a user. I.e., Something like `np = int(model_component_meta['total_processors']) if model_component_meta['total_processors'] else N_PROCESSORS_DEFAULT`
4. `eval(str(...))` is dangerous --- you risk running arbitrary python code. If we know to expect an int, just coerce to int --- `int(model_component_meta['total_processors'])`.

ashiklom (commented):
I had a few minor comments on the Python implementation, but I'm not equipped (yet) to comment on the YAML configs that are the bulk of this PR or the science behind them. I'll have to defer to somebody with a better understanding of JEDI for that.

src/swell/tasks/run_jedi_convert_state_soca2cice_executable.py
Dooruk: Thanks for catching this, must have slipped in the commits!

I will set a hard override for single node in `slurm.py`. Currently we don't have a mechanism to define processors for individual tasks so I still have to define a constant value here.

Hmm, good heads up on eval, I used that in other parts of the code.

src/swell/tasks/link_geos_output.py
Dooruk: Sounds good!

Dooruk:
> I had a few minor comments on the Python implementation, but I'm not equipped (yet) to comment on the YAML configs that are the bulk of this PR or the science behind them. I'll have to defer to somebody with a better understanding of JEDI for that.

Thanks for reviewing this PR. I'm the only one exercising the cyling GEOS+SOCA capability currently so I added `geos_marine` as a first step. Hopefully in the second step (another PR) when I take `geos_ocean` out David can test it to make sure ocean-only DA still works.

src/swell/configuration/jedi/interfaces/geos_marine/task_questions.yaml
```diff
@@ -0,0 +1,111 @@
+analysis_forecast_window_offset:
```
asewnath, line 1: This variable is typed as `string-check-list` and is expecting `options` to be defined.

src/swell/configuration/jedi/interfaces/geos_marine/task_questions.yaml
```diff
@@ -0,0 +1,111 @@
...
+  default_value: -PT6H
+
+analysis_variables:
```
asewnath, line 6: This variable is typed as `string-check-list` and is expecting `options` to be defined.

src/swell/configuration/jedi/interfaces/geos_marine/task_questions.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+cice6_domains:
+  default_value:
+  - arctic
+  - antarctic
+  options:
+  - arctic
+  - antarctic
+
+clean_patterns:
```
asewnath, line 40: This variable is typed as `string-check-list` and is expecting `options` to be defined.

asewnath (commented):
Just some comments related to a bug found when building swell experiments with the CLI

Dooruk:
I am going to merge this to move forward with the FMS updates.

PR #377, opened 2024-06-28
Add option to cycle VarBC

### LINKED ISSUE #375: Introduce `cycling_varbc` as a conditional in flow.cylc

Dooruk:
The linking from the previous cycle works, I'm still receiving errors with the variational task but @rtodling you can test this and see what the problem could be.

`cycling_varbc` is introduced as a bool option that impacts the suite behavior. I also added `first_cycle_dto` as an object.

Cylc graph outputs to explain `flow.cylc` changes:

- ## Cycling turned off
All cycles start with the lack of inter-cycle dependencies, except the tasks dependecies on initial cycle (R1) tasks.
no_cycling

- ## Cycling turned on
`get_observations.py` will wait until previous cycle Variational task is complete.
yes_cycling

### DIFF
--- a/pycodestyle.cfg
+++ b/pycodestyle.cfg
@@ -8,4 +8,4 @@
 max-line-length = 100
 indent-size = 4
 statistics = True
-#ignore = build
+exclude = ._*, build
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ click>=8.0.0
 jinja2>=3.0.3
 pyyaml>=6.0
 pycodestyle>=2.11.0
+numpy<2
 pandas>=1.4.0
 isodate>=0.5.4
 f90nml>=1.4.3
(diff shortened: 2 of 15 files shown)


rtodling:
This seems to do what we expect - cool diagram of the cycling one. I tried two cycles last night; I am now trying two experiment  for about 3 days of analyses one w/o cycled varBC and another cycled. I have already been cycled two similar experiments in the context of the DAS ... I see virtually no difference in the increments which is very encouraging; it means the JEDI VarBC is estimating newly the same biases as GSI's.

rtodling (approved):
I think this is working rather nicely. I have two cases running within the SWELL context and similar cases running within the GEOS workflow and both compare rather nicely. 

What I see when I compare increments from non-cycled and cycled VarBC with increments from GSI, I see very little differences in the increment differences. This to me indicates the VarBC in the cycled case is basically reproducing the VarBC from GSI.

The mechanism introduced in this PR by @Dooruk to cycle VarBC is working as expected. The only thing (which is unrelated to this PR) is that the runahead_limit in the experiment.yaml cannot be larger than P3- anything larger than that causes SWELL to get an error in some process (typically the RunVar) that must be re-triggered by hand within cylc. 

I'm good w/ what's in here.

PR #396, opened 2024-07-30
Parsing ozinfo.db from GEOS_mksi and removing old observing system yamls

### LINKED ISSUE #369: Add handing of OZONE database
This is similar to the satellite database handling, but for ozone.

asewnath:
## Description

- Modifying `GenerateObservingSystemRecords` to process both `sidb` and `ozinfo.db` from `GEOS_mksi`. All resulting yamls are found in the `observing_system_records` directory inside the run directory. Note that there is currently no logic doing more with the ozinfo yamls besides saving them to the run directory.

- Removing observing system yaml files since it is there for reasons that are no longer pursued and can be confusing to new users.

Resolves #369


rtodling (approved):
This seems to work for me.

Thank you.

Dooruk (approved):
Less YAMLs in SWELL, @ashiklom will be happy 🎉 

Thanks @asewnath.

Issue #362, opened 2024-06-20
Wired observations in geos-atmosphere: needs to be dynamic


rtodling:
I changed changed the period in which a test is ran w/ SWELL and found (something I knew but forgot) that the set of observing system used in a run w/ GEOS-ATMOSPHERE is wired ... by that I mean that the experiment.yaml always lists the same ops-types regardless of period it's dealing w/ ... for example, in the present list we find aura_omi ... which works for the period of Dec 2021 ... but fails for the period Aug 2023 when these data are not available for a certain time. 

The fact is that this list could be as general as possible, but the code should all fail is something is in the database for a certain period and not found ... otherwise, the code should simply disregard not finding a certain obs type ... whether that obs type should be in the observers part of the yaml is debatable ... as long as JEDI understands the zero-ob case, there should be not issue - but I think JEDI only understand zero-ob when the file is actually present ... which means we'd have to have zero obs IODA files created ... perhaps better to simply not include missing data-type in the final yaml ...

rtodling:
I will add a little to this: part of the issue here is the disconnect between GetObservations and GetObservingSystemsRecord - the first should depend on the latter ... presently that dependency does not exist. What needs to happen is there needs to be a check of consistency of the ops-types listed in the experiment.yaml and what is actually present in the db ... if a particular ops-type is requested by found not to be available at the current cycle, the swell machinery should simply ignore that type as opposed to leave w/ an error saying the obs file corresponding to that type has not been found.

rtodling:
I have a suggestion for how this could be handled in swell.

The experiment.yaml could have as default all possible observation types (what GEOS calls it classes). 

SWELL could introduce a variable similar to that in GEOS, named **strict**. The default for this would be **false**.  A check could be placed in get_observations that would compare the mksi findings for what applies for the current period and the actual observations found in, say r2d2 for the cycle in question. If an obs-type is request by mksi but the corresponding file is missing in r2d2, the job would go on merrily if strict=false, and would halt when strict=true. 

The variable strict would be available for the user to set in experiment.yaml.

Dooruk:
I would like to revisit this discussion as I'm now encountering similar problems with the GMAO/ODAS observations in terms of missing files. As I understand, the UFO `missing_file_action` option is not reliable at the moment (doesn't work when certain filters are on) and that doesn't help with our workflow (`experiment.yaml`, `get_observations`, etc.) anyway.

[...]

PR #400, opened 2024-08-01
allow for missing obs file w/o a JEDI crash

rtodling:
## Description

This allows for ops-class (type) to be in yaml even when input file is missing ... JEDI will warn of missing file, but it won't crash - in OPS these would need to be set more restrictively. 

See documentation for specifics on options for this: 
https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/ioda/file-formats.html#missing-file-action

## Dependencies

None

## Impact

None

## Issue

This is part of what's needed to address issue https://github.com/GEOS-ESM/swell/issues/362  - in this case, wiring the obs-classes would be ok ... but r2d2 (or swell) must be less strict and allow for job to continue when obs file is not found - this is to be done and therefore this is still a DRAFT.


rtodling:
The knob placed here will no fully work as they should. They only work when some filters are not present. This knob is supposed to create an zero-ob observer object corresponding to the missing instrument and create allow for things like satbias and satbias_cov (for radiance) to be propagated even in the absence of the instrument - which is what we want. Unfortunately, some of the filters don't work properly in the absence of observations (zero-ops) so the quest is now one of finding the offending filters and get them fixed; a non-trivial task.

gmao-jjin3 (approved):

gmao-yzhu:
The changes made are necessary. At the same time, in terms of the JEDI code workflow, when data are missing, the filters should all be skipped.

gmao-jjin3:
I made a few minor updates based on this PR.  So far, my PR https://github.com/GEOS-ESM/swell/pull/415 works for missing convectional, ozone, and microwave radiance data. It crashes swell-3dfgat_atmos run when any of AIRS, IASI, CRIS-FSR, and AVHRR input files is missing.

PR #402, opened 2024-08-02
Add to docs a procedure note on code_tests

metdyn:
## Description

Adding `test codes before pull request` to docs to save the procedure we used in the group. Especially,
set os.environ["LOG_INFO"] = "1"  for ctest


Dooruk (approved):
Thanks for contributing to the docs! Let's wait for the ensemble changes first then merge this.

Issue #404, opened 2024-08-06
run_jedi_hofx_ensemble_executable.py shows errors on discover


metdyn:
In `localensembleda` suite, the `run_jedi_hofx_ensemble_executable` is  a local step.  It could not run on discover currently.   The issues include:
- some file directory has been outdated
- yaml related to HofX3D is missing
- the flow.cylc which uses this enHX with packets was outdated: not using the current GenerateObservingSystemRecords. 

I am working on this.

PR #397, opened 2024-07-31
Add task to calculate ensemble mean and variance from background state

metdyn:
## Description

A step towards LETKF is to calculate ensemble mean and variance:
- https://github.com/GEOS-ESM/swell/issues/365

Alternatively, this functionality may be used as a simple, independent tool.

Tested on SCU17:
```
swell create localensembleda  -s slurmfile.yaml -p nccs_discover_sles15

slurm_directives_tasks:
  RunJediLocalEnsembleDaExecutable:
    geos_atmosphere:
      nodes: 1
      ntasks-per-node: 36
  GetEnsembleMeanVariance:
    all:
      nodes: 1
      ntasks-per-node: 36
```

## Dependencies

None

## Impact

It is part of the localensembleda suite.

Issue #314, opened 2024-03-21
(SE) `task_question.yaml`: a design issue


rtodling:
I am finding more and more that a single task_question file to handle all question that can ever be asked of swell for all possible tasks is a true bottle neck.

1) many questions for a particular task have nothing to do w/ another. Keep them in the same file causes confusion

2)  the way things are also prevents generality; for example:  the LETKF uses a cubed ensemble; the Variational analysis uses a lat-lon ensemble; the GetEnsemble task refers to a variable path_to_ensemble which means only one default can be set right - preventing us from being able to set defaults for two completely independent suites LETKF and Var.  Sure there are ways around it be they become entwined w/ more variables needing to be defined w/o need.

3) default npc and npy can only be given as general settings  - equal to all applications - but that will not be the case. 

There're other examples of the inconvenience of this single all-purpose file. It would be nice to rethink this.

Dooruk:
I will add some thoughts/feedback here somewhat regarding this issue.

In #377, I'm adding a new configuration key `cycling_varbc`, which would impact the `3dvar_atmos` and `3dfgat_atmos` suites' `flow.cylc` file. So my first thought was to add this into `suite_questions.yaml`, but then I wouldn't be able to access `cycling_varbc` inside a task, `get_observations.py` in this case. Hence, I added it to `task_question.yaml` instead and not in `suite_questions.yaml`. So I wasn't sure if the best approach is to have them in both places or just keep it in the `task_questions.yaml`. Moving forward with the `geos_marine`, which handles multiple models,  I will include a `turn_on_icemodel` switch and will encounter the same issue.

To be able to use the templated version`{{cycling_varbc}}` in `flow.cylc`, I had to add the exception part in `create_experiment.py` where it takes a model level `experiment.yaml` dictionary setting, and copies it to the first level, along with `cycle_times` `experiment_id` etc. The same approach was taken for the `ensemble_*` configuration keys.

I overall like the `task_questions.yaml` structure, makes it possible to track what config keys are used for which tasks and simplifies questionary creation, so this is more of a feedback.

ashiklom:
[...]

PR #405, opened 2024-08-07
Fix  run_jedi_hofx_ensemble_executable.py  related to LocalEnsembleDA

metdyn:
## Description

See details in
- https://github.com/GEOS-ESM/swell/issues/404

## This is a prerequisiste for
- https://github.com/GEOS-ESM/swell/pull/397

## Unresolved issue

- swell test code_tests failed to detect  polymorphism:
class RunJediHofxEnsembleExecutable(RunJediHofxExecutable):
class RunJediHofxExecutable(taskBase):

so that code_tests tries to regenerate  task_questions.yaml by removing `RunJediHofxEnsembleExecutable and RunJediLocalEnsembleDaExecutable`, because it is not smart enough. 
```
tasks/task_questions.yaml
save_geovals:
  ask_question: false
...
  tasks:
  - RunJediHofxExecutable
  - RunJediHofxEnsembleExecutable
  - RunJediLocalEnsembleDaExecutable
  type: boolean
```

This is a dilemma:
If I use my own  questions.yaml,  code_tests will fail. 
If I use the regenerated  questions.yaml,  I got the following error in the end of the text.

We have several choices:
- use tricks to by pass  code_tests, because it does not see polymorphism
- fix the bug in code_tests
- eliminate  super().execute(ensemble_members=packet_ensemble_members)  in enHX, stop using this polymorphism, and just transplant the needed code from HX to enHX. 

What is the previous experience to overcome the defects in test_codes?

│ File "/gpfsm/dnb07/projects/p234/yyu11/br_jedi_scu17/install_swell/src/swell/swell_lenda/src/swell/tasks/run_jedi_hofx_ensemb\
le_executable.py", line 141, in execute
│ │ super().execute(ensemble_members=packet_ensemble_members)

│ File "/gpfsm/dnb07/projects/p234/yyu11/br_jedi_scu17/install_swell/src/swell/swell_lenda/src/swell/tasks/run_jedi_hofx_execut\
able.py", line 42, in execute
│ │ save_geovals = self.config.save_geovals()

│ File "/gpfsm/dnb07/projects/p234/yyu11/br_jedi_scu17/install_swell/src/swell/swell_lenda/src/swell/utilities/config.py", line\
 135, in variable_not_found
│ │ self.__logger__.abort(f'In config class, trying to get variable \'{name}\' but ' +


Dooruk:
@metdyn, can you please merge `develop` into this and resolve conflicts, aftrerwards I will test it. Thanks!

src/swell/suites/localensembleda/flow.cylc
```diff
@@ -42,8 +47,19 @@
         {% if cycle_time[model_component] %}
             # Task triggers for: {{model_component}}
             # ------------------
+            # Generate satellite channel records
+            CloneGeosMksi-{{model_component}}[^] => GenerateObservingSystemRecords-{{model_component}}
+
             # Perform staging that is cycle dependent
-            CloneJedi[^] => StageJediCycle-{{model_component}}
+            #BuildJediByLinking[^]? | BuildJedi[^] => StageJediCycle-{{model_component}}
+            StageJediCycle-{{model_component}}
```
Dooruk, line 55: StageJedi has a dependency here, uses some files from the linked JEDI build directory in the experiment folder. So we should maintain that.

src/swell/test/suite_tests/localensembleda-tier1.yaml
```diff
@@ -1,19 +1,19 @@
 final_cycle_point: '2021-12-12T00:00:00Z'
 jedi_build_method: use_existing
+save_geovals: true
```
Dooruk, line 3: I don't think this is needed for ensemble, could be set to false.

src/swell/test/suite_tests/localensembleda-tier1.yaml
```diff
@@ -1,19 +1,19 @@
...
 model_components: ['geos_atmosphere']
 models:
   geos_atmosphere:
-    horizontal_resolution: '4'
-    npx_proc: 2
-    npy_proc: 2
+    npx_proc: 1
+    npy_proc: 1
```
Dooruk, line 8: Particular reasons to decrease processor counts?

Dooruk (commented):
Looks like there is an issue with window_type. Ensmeanvariance works with 3D but not with the 4D window type. On the contrary, enshofx requires 4D window type. I'm pretty sure we would like 4D window type for enshofx (@rtodling could correct me in today's meeting) so we just need to figure out to run Ensmeanvariance that way?

Your second comment pertains the inheritance rather than polymorphism perhaps? Child classes obtaining parent inputs. With the way `task_questions.yaml` is setup, that is not quite possible. We used inheritance originally but then abandoned it, I see the benefit of it for ensemble tasks, so we could revisit that.

Some potential strategies were discussed here:
https://github.com/GEOS-ESM/swell/issues/314

src/swell/suites/localensembleda/flow.cylc
metdyn: Thanks, that makes sense.  I had a miscocept that R1 will run first before the Tij group, which is incorrect.   Code corrected.

src/swell/test/suite_tests/localensembleda-tier1.yaml
metdyn: Removing this key helps to pass `swell test code_tests`.  The default in task_quetions.yaml is false.

src/swell/test/suite_tests/localensembleda-tier1.yaml
metdyn: oops, it is for simple test only and will be restore once I am done with the development.

metdyn:
I am closing the PR to work offline to solve many issues.  Thanks, @Dooruk!

PR #408, opened 2024-08-09
CLI for Tier 1 tests

ashiklom:
This adds the ability to run tier 1 tests directly from the command line, bypassing the automated GitHub actions-based CI. The basic interface is similar to the standard Swell CLI:

```
swell t1test hofx
```

I've added an additional optional configuration file --- `~/.swell/swell-test.yaml` that (1) defines the `test_root` --- the directory where test outputs will be put (if absent, defaults to a temporary directory that will be deleted after the test completes) --- and (2) adds the ability to set user-specific test overrides. Here's an example I'm currently using:

```yaml
test_root: "/home/ashiklom/projects/SwellExperiments/tier1"
override:
  models:
    geos_atmosphere:
      observing_system_records_mksi_path: "/home/ashiklom/projects/GEOS_mksi"
```

All of this is documented in `docs/code_tests/suite_tests.md`.

For now, the tests only run the jobs --- there's no validation of config files, outputs, etc. That will come later. The goal here was to put in place a basic set of local integration tests to support new Swell developers (like me! And our forthcoming Swell engineer).


docs/code_tests/suite_tests.md
```diff
@@ -0,0 +1,59 @@
...
+One recommended override, per the example below, is to request a local copy of `GEOS_mksi`, since compute nodes will not be able to clone the repo from GitHub (per the default Swell behavior)
+    ```yaml
+    test_root: ~/projects/SwellExperiments/tier1
+    override:
+      models:
+        geos_atmosphere:
+          observing_system_records_mksi_path: "~/projects/GEOS_mksi"
+    ```
+
+2. Set up your compute SWELL environment interactively (following the [Platform specific instructions](platforms/platforms.md) for your system).
```
asewnath, line 34: @ashiklom Can you explain this instruction a bit more? I'm testing the tier1 tool (on Milan node) and I'm running into this issue: `[jobs-submit err] No matching platform "nccs_discover_sles15" found`. I don't usually run on sles15, so I'm not sure what I'm missing.

asewnath: Maybe this is a question for @Dooruk as well?

Dooruk: @asewnath, you might be missing the global Cylc file change mentioned here:

https://geos-esm.github.io/swell/#/platforms/discover/configuring_cylc_discover

And `swell create` takes `-p` argument (the default is `nccs_discover`, that should be `swell create --platform nccs_discover_sles15`. If you are using overrides, make sure your platform key is `nccs_discover_sles15`.

asewnath: Thanks @Dooruk!

asewnath: I am able to use this tool now, thanks!

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -0,0 +1,107 @@
...
+        if k is None:
+            raise KeyError("observing_system_records_mksi_path is set but is None")
+    except KeyError:
+        print(
+            "WARNING: 'mksi_path' test config is unset. " +
+            "By default, Swell will try to clone GEOS_mksi from GitHub, " +
+            "which will fail on compute nodes with no internet access."
+        )
+
+    experiment_dir = testdir / experiment_id
```
Dooruk, line 71: Just curious, do you prefer not to use `os.path.join`? Is it for code brevity?

Dooruk (approved):
This worked for me. I successfully ran `ufo_testing` on a login node (sorry Discover users) before reading the documentation.

Thanks for the documentation!

src/swell/test/suite_tests/suite_tests.py
ashiklom: Partially, brevity. Partially, it's because I think `pathlib.Path` is more idiomatic Python 3. And partially, it's because `pathlib.Path` objects come with a bunch common file-related methods built in so there's a bit less to import and to remember.

E.g., Construct a path, list all the files, and copy them to a new location.

```python
import os
import glob
import shutil
rootdir = "my-root"
destdir = "my-dest"
os.makedirs(destdir, exist_ok=True)
subdir = os.path.join("my-root", "subdir")
files = glob.glob(os.path.join(subdir, "*"))
for f in files:
    shutil.copy(f, os.path.join(destdir, os.path.basename(f)))

# With pathlib
from pathlib import Path
rootdir = Path("my-root")
destdir = Path("my-dest")
destdir.mkdir(exist_ok=True, parents=True)
subdir = rootdir / "subdir"
files = subdir.glob("*")
for f in files:
    shutil.copy(f, destdir / f.name)
```

The `pathlib` version is slightly shorter, has one less import, and, in my opinion, looks a bit cleaner.

PR #412, opened 2024-08-26
Add command line option to set LOG_INFO  in code_tests.py

metdyn:
## Description

This modification is introduced to allow users to use command line to turn on LOG_INFO in code_tests.py without modifying code_tests.py itself. 
```
export LOG_INFO=1
swell test  code_tests
```
Then compare, e.g., /tmp/geos_atmosphere_task_questions_Ki1d65V3.yaml  with task_questions.py for details.

## Dependencies

None

## Impact

Make unit test for user easier

### DIFF
--- a/src/swell/test/code_tests/code_tests.py
+++ b/src/swell/test/code_tests/code_tests.py
@@ -25,8 +25,15 @@ def code_tests():
     logger = Logger('TestSuite')
     logger.test('Running Swell Test Suite')
 
-    # Turn off the regular info testing
-    os.environ["LOG_INFO"] = "0"  # Set this to 1 when errors are being debugged
+    # Check for environment variable
+    log_env = os.environ.get('LOG_INFO')
+
+    # If found set element to environment variable
+    if log_env is None:
+        # Turn off the regular info testing (default)
+        os.environ["LOG_INFO"] = "0"  # Set this to 1 when errors are being debugged
+    else:
+        os.environ["LOG_INFO"] = log_env
 
     # Create a test suite
     test_suite = unittest.TestSuite()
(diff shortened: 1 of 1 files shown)


src/swell/test/code_tests/code_tests.py
```diff
@@ -25,8 +25,15 @@ def code_tests():
...
-    os.environ["LOG_INFO"] = "0"  # Set this to 1 when errors are being debugged
+    # Check for environment variable
+    log_env = os.environ.get('LOG_INFO')
+
+    # If found set element to environment variable
+    if log_env is None:
+        # Turn off the regular info testing (default)
+        os.environ["LOG_INFO"] = "0"  # Set this to 1 when errors are being debugged
+    else:
+        os.environ["LOG_INFO"] = log_env
```
ashiklom: Good feature, but implementation can be simplified with `os.environ.setdefault`

```suggestion
    # Logging level
    os.environ.setdefault("LOG_INFO", "0")
```

If `LOG_INFO` is set, this will not do anything -- `os.environ["LOG_INFO"]` will be whatever value was passed via command line. If `LOG_INFO` is unset, this effectively does `os.environ["LOG_INFO"] = "0"`.

An example:

```
python -c "import os; os.environ.setdefault('NOTAVAR', '3'); print(os.environ['NOTAVAR'])"
# 3
NOTAVAR=5 python -c "import os; os.environ.setdefault('NOTAVAR', '3'); print(os.environ['NOTAVAR'])"
# 5
```

ashiklom (changes requested):
Good feature! But can be implemented with one line -- see suggestion.

src/swell/test/code_tests/code_tests.py
metdyn: Thank you so much, @ashiklom. That is so nice!   And I have made the modifications!

ashiklom (approved):

PR #413, opened 2024-08-26
LocalEnsembleDA: use_linear_observer in GETKF

metdyn:
## Description

In Task local_ensemble_da, we use the linear_observer option for GETFK to run the suite.  For this task, we skip the separate HofX task, then run the same executable `fv3jedi_letkf.x`.

Keywords:
```
    local_ensemble_solver: GETKF
    skip_ensemble_hofx: true                 # knob to skip the old external HX design
    local_ensemble_use_linear_observer: true
```

slurmfile.yaml
```
slurm_directives_tasks:
    geos_atmosphere:
      nodes: 1
      ntasks-per-node:  N
  RunJediEnsembleMeanVariance:
    all:
      nodes: 1
      ntasks-per-node: N
```
choosing appropriate N for scu17 or scu16. 
(I use 120 for scu17 and I think N .ge. 24 for scu16)

output dir = /discover/nobackup/projects/gmao/aist-nr/yyu11/br_jedi_scu17/run_swell/getkf/run/20211212T000000Z/geos_atmosphere

## Dependencies

done


Dooruk (approved):
This is all good, I tested it and was able to run it on my end, thank you @metdyn . I like the `sync_point` feature, creates code brevity. I will merge this as no Tier tests are currently testing this.

Regarding EVA, as you mentioned some groups are missing from the processed observation files, namely `EffectiveQC`. Since they are required for some of the EVA transformations the task is failing. We just need to edit the `eva-observations.yaml` file to create different diag plots for the lgetkf case. We can discuss that with @rtodling.

metdyn:
Right, I used `sync_point` to do a relay, before connecting to the long common name `RunJediHofxEnsembleExecutable-{{model_component}}`, which makes the flow chart look simpler.

We can modify the EVA file (there seems to be a significantly number of variables that won't be needed).

> This is all good, I tested it and was able to run it on my end, thank you @metdyn . I like the `sync_point` feature, creates code brevity. I will merge this as no Tier tests are currently testing this.
> 
> Regarding EVA, as you mentioned some groups are missing from the processed observation files, namely `EffectiveQC`. Since they are required for some of the EVA transformations the task is failing. We just need to edit the `eva-observations.yaml` file to create different diag plots for the lgetkf case. We can discuss that with @rtodling.

PR #414, opened 2024-08-28
Error reporting for empty channel lists

### LINKED ISSUE #411: `get_channels` utilities are fragile and throw uninformative errors
When running Tier 1 tests, I frequently hit an error like this:

```
"/gpfsm/dnb33/ashiklom/swell/src/swell/swell_develop/src/swell/tasks/run_jedi_hofx_executable.py", line 117, in execute
    jedi_dictionary_iterator(jedi_config_dict, self.jedi_rendering, window_type,
  File "/gpfsm/dnb33/ashiklom/swell/src/swell/swell_develop/src/swell/utilities/run_jedi_executables.py", line 48, in jedi_dictionary_iterator
    jedi_dictionary_iterator(value, jedi_rendering, window_type, obs,
  File "/gpfsm/dnb33/ashiklom/swell/src/swell/swell_develop/src/swell/utilities/run_jedi_executables.py", line 75, in jedi_dictionary_iterator
    obs_dict = jedi_rendering.render_interface_observations(ob)
  File "/gpfsm/dnb33/ashiklom/swell/src/swell/swell_develop/src/swell/utilities/render_jedi_interface_files.py", line 206, in render_interface_observations
    avail_channels, active_channels = get_channels(self.observing_system_records_path,
  File "/gpfsm/dnb33/ashiklom/swell/src/swell/swell_develop/src/swell/utilities/get_channels.py", line 91, in get_channels
    active_channels_list = process_channel_lists(active_channels)
  File "/gpfsm/dnb33/ashiklom/swell/src/swell/swell_develop/src/swell/utilities/get_channels.py", line 28, in process_channel_lists
    if '-' in element:
TypeError: argument of type 'NoneType' is not iterable
```

Here's the relevant source:

https://github.com/GEOS-ESM/swell/blob/develop/src/swell/utilities/get_channels.py#L24-L33

This is because one or more entries in `channel_list` is `None`. But the underlying issue here is missing observations/channel files (which might be caused by issues in GEOSmksi). (For reference, the root cause of this particular error was that I was using the `main` branch of GEOSmksi` rather than the `develop` branch.)

I think we need to:

1. Catch and report this error more accurately upstream (e.g., `"Missing channels for variable X. Confirm that you are using the right version of GEOSmksi"`).
2. Write some tests that check these capabilities in isolation (or maybe just start running the existing `missing_obs_test.py` that @asewnath already wrote as part of our `code_tests` suite).

asewnath:
## Description

- Adding descriptive logger aborts to `get_channels` for empty channel lists.
- Adds/modifies code testing for generating observing system yaml files

Resolves #411.


asewnath:
@ashiklom ready for review

src/swell/test/code_tests/test_generate_observing_system.py
```diff
@@ -0,0 +1,90 @@
...
+    # Clone repo if not already cloned
+    if not os.path.exists("GEOS_mksi"):
+        git_clone_cmd = ["git", "clone", "-b", branch, url, "GEOS_mksi"]
+        subprocess.run(git_clone_cmd, stderr=subprocess.DEVNULL)
+    else:
+        cwd = os.getcwd()
+        os.chdir("GEOS_mksi")
+        git_checkout_cmd = ['git', 'checkout', branch]
+        subprocess.run(git_checkout_cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        os.chdir(cwd)
```
ashiklom: `os.chdir` is a bit fragile and also harder to reason about (you have to look back to the most recent call to remember exactly what directory you're in).

A more reliable approach is to tell `subprocess.run` exactly where the command should be running.

```suggestion
        git_checkout_cmd = ['git', 'checkout', branch]
        subprocess.run(git_checkout_cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, cwd="./GEOS_mksi")
```

src/swell/utilities/observing_system_records.py
```diff
@@ -69,6 +69,9 @@ def read_sat_db(path_to_sat_db, column_names):
                 # convert channel list to string
                 df.loc[idx, 'channels'] = str(channel_list)
                 idx += 1
+
+    file.close()
```
ashiklom: A more reliable (and pythonic!) way to do this is to use a context manager, which will automatically handle closing the file for you; i.e., earlier, on line 31:

```python
with open(filename, "r") as file:
  lines = file.readlines()

# Read blindly into an array...
```

ashiklom:
Two small suggestions, but otherwise, this looks good! Special thanks for putting together and enabling the code test!

src/swell/test/code_tests/test_generate_observing_system.py
asewnath: Cool! Might also change `chdir` usage in `jedi_bundle` as well

ashiklom (approved):

PR #419, opened 2024-09-12
Cycling FGAT for `geos_marine` and towards erasing `geos_ocean`

Dooruk:
This is a step towards getting rid of `geos_ocean` model section and sticking with `geos_marine` but that will require further testing and verification beyond the scope of this PR.

For context, SOCA (marine component of JEDI) is designed to handle different model types such as ocean(mom6), sea-ice (cice6), waves(ww3), and BGC. This means, there will be optional tasks (such as SOCA2CICE) that not need triggered while CICE model is not active. Additionally, there are optional configuration entries for the input config YAML, for instance `ice_filename` and `wav_filename` below in the `state` section are optional entries. More detailed information [here](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/soca/classes.html).

``` yaml
state:
  read_from_file: 1
  basename: ./INPUT/
  ocn_filename: MOM.res.nc
  ice_filename: cice.res.nc
  wav_filename: wav.res.nc
  date: 2018-04-15T00:00:00Z
  state variables: [cicen, hicen, socn, tocn, ssh, hocn, uocn, vocn, swh]
```
To handle this, this PR introduces a new key called `marine_models` to be used in `flow.cylc` and various tasks. Jinja is able to utilize this key using if statements, I tested it with both `flow.cylc` and EVA YAMLs. 

Second, I added a `dynamic_keys` feature, (see `utilities/render_jedi_interface_files.py`). These are entries that are dynamic, for instance `states` in `PseudoModel` would be dependent on `window_length` and `background_frequency` as well as `marine_models`, see below for it's output, it has two entries because `background_frequency` is 3H and `window_length` is 6H:

```yaml
name: PseudoModel
states:
- basename: ./
    date: '2021-07-02T06:00:00Z'
    ice_filename: ice.fc.2021-07-02T03:00:00Z.PT3H.nc
    ocn_filename: ocn.fc.2021-07-02T03:00:00Z.PT3H.nc
    read_from_file: 1
- basename: ./
    date: '2021-07-02T09:00:00Z'
    ice_filename: ice.fc.2021-07-02T03:00:00Z.PT6H.nc
    ocn_filename: ocn.fc.2021-07-02T03:00:00Z.PT6H.nc
    read_from_file: 1
tstep: PT3H
```


src/swell/configuration/jedi/interfaces/geos_marine/model/background.yaml
```diff
@@ -2,5 +2,7 @@ date: '{{local_background_time_iso}}'
 read_from_file: 1
 basename: './'
 ocn_filename: 'MOM6.res.{{local_background_time}}.nc'
+{% if 'cice6' in marine_models %}
```
ashiklom, line 5: So, for reference, where I _personally_ want to see us get to is removing these YAML files altogether and replacing them with a Python dictionary that writes out YAMLs at runtime. E.g.,

```python
background = {
  "date": local_background_time,
  "read_from_file": 1,
  "basename": pathlib.Path('.'),
  #...
}

if 'cice6' in config.models.marine_models:
  background["ice_filename"]: f"cise.res.{local_background_time}"
  
# Some function to test that the background is valid
# Can have more complex logic about
validate_jedi_background(background)

# ...
# When we're ready to process the experiment
# ...
with open("path/to/experiment/jedi/geos_marine/background.yaml", "w") as f:
  yaml.dump(background, f)
```

This approach gives you a lot more power and flexibility in writing these files, _and_ means that you only have to understand Python syntax (as opposed to Python _and_ YAML _and_ Jinja).

ashiklom: That said, as a stopgap, what you have here is fine --- it's clear enough what's going on.

src/swell/tasks/generate_b_climatology.py
```diff
@@ -199,19 +201,22 @@ def parameters_diffusion_vt(self):
...
+        if not self.generate_yaml_and_exit:
+            self.logger.info('Running '+jedi_executable_path+' with '+str(self.np)+' processors.')
+            command = ['mpirun', '-np', str(self.np), jedi_executable_path, jedi_config_file]
 
-        command = ['mpirun', '-np', str(self.np), jedi_executable_path, jedi_config_file]
+            # Move to the cycle directory
+            # ---------------------------
+            os.chdir(self.cycle_dir())
+            if not os.path.exists('background_error_model'):
+                os.mkdir('background_error_model')
```
ashiklom: `os.chdir` is really fragile, and makes the code harder to reason about because you have to remember what the last time you called `os.chdir` was.

I strongly recommend avoiding it and just doing something like this here:

```python
background_error_model_dir = os.path.join(self.cycle_dir(), "background_error_model")
if not os.path.exists(background_error_model_dir):
  os.mkdir(background_error_model_dir)
```

For running subprocess commands, `subprocess` supports a `cwd` argument. I.e., Instead of:

```python
os.chdir("somewhere")
subprocess.run("some_command", ...)
```

...do...

```python
subprocess.run("somecommand", ..., cwd="somewhere")
```

I'm not sure if `run_track_log_subprocess` is set up to accept `cwd` as an (optional) argument. But if not, it should be! (More generally, all our `run_..._subprocess` commands should just accept `**kwargs` as an argument that gets propagated through to `subprocess.run(..., **kwargs)`.

src/swell/tasks/link_geos_output.py
```diff
@@ -27,73 +30,87 @@ def execute(self):
...
+        self.bkgr_time_iso, self.bkgr_time_dto = self.da_window_params.local_background_time(
+            self.window_offset,
+            self.window_type,
+            dto=True)
+
+        # Create source and destination files for linking model output to cycle directories
+        # -----------------------------------------------------------------------------------
+        self.src_dst_dict: Dict[str, Tuple[str, ...]] = {}
+
+        if self.get_model() == 'geos_ocean' or self.get_model() == 'geos_marine':
```
ashiklom: Stylistically, I have a slight preference for the following. It's shorter, and IMHO, the intent is clearer.

```suggestion
        if self.get_model() in ('geos_ocean', 'geos_marine'):
```

src/swell/suites/3dfgat_cycle/eva/increment-geos_marine.yaml
ashiklom, line 1: @asewnath Per my comment above, one thing I'd like you to explore (when you have the time) is defining a Python class, or set of Python classes composed together, (better yet: a [dataclass](https://docs.python.org/3/library/dataclasses.html) or [pydantic data model](https://docs.pydantic.dev/latest/)) that might ultimately replace YAML files like this.

I'm _inclined_ to think that using Python classes would be better, but it's hard to know exactly until we actually see a prototype implementation and try to do some stuff with it.

src/swell/tasks/link_geos_output.py
```diff
@@ -27,73 +30,87 @@ def execute(self):
...
-
-        dst = 'MOM6.res.' + self.current_cycle + '.nc'
-
-        return src, dst
+    def link_mom6_history(self, src_dst_dict):
+
+        # Create links to GEOS history for SOCA inputs
+        # Depending on the DA type, there could be multiple state files to link
+        # -----------------------------------------------------------------------
+        if self.window_type == '4D' or 'fgat' in self.suite_name():
```
ashiklom: This conditional spans this entire function, so I would suggest refactoring into distinct functions like `link_mom6_history_fgat` vs. `link_mom6_history_default` (or something), and then call them as:

```python
        if self.get_model() == 'geos_ocean' or self.get_model() == 'geos_marine':
            if `self.window_type == "4D" or 'fgat' in self.suite_name():
              self.link_mom6_history_fgat(self.src_dst_dict)
            else:
              self.link_mom6_history_default(self.src_dst_dict)
```

src/swell/tasks/link_geos_output.py
```diff
@@ -27,73 +30,87 @@ def execute(self):
...
+            self.background_frequency = self.config.background_frequency()
+
+        self.bkgr_time_iso, self.bkgr_time_dto = self.da_window_params.local_background_time(
+            self.window_offset,
+            self.window_type,
+            dto=True)
+
+        # Create source and destination files for linking model output to cycle directories
+        # -----------------------------------------------------------------------------------
+        self.src_dst_dict: Dict[str, Tuple[str, ...]] = {}
```
ashiklom: Do we expect this attribute to be used anywhere else in the class. If yes, then we can keep it as an attribute, but then, it probably makes sense to change the MOM6 method below to just take `self` as an argument and grab this attribute from `self`.

If not, then we can just make it a local variable and pass it to the method without the `self`.

src/swell/utilities/data_assimilation_window_params.py
```diff
@@ -110,9 +110,15 @@ def local_background_time_iso(self, window_offset, window_type):
...
-    def local_background_time(self, window_offset, window_type):
+    def local_background_time(self, window_offset, window_type, dto=False):
 
         local_background_time = self.__get_local_background_time__(window_type, window_offset)
+
+        # Return datetime object if asked
+        if dto:
+            return local_background_time.strftime(datetime_formats['directory_format']), \
+                local_background_time
+
```
ashiklom: First, the behavior of `dto=True/False` here is inconsistent with other functions that do this (like `window_start/end_iso`. Those functions return _either_ a datetime object or the string representation. Here, `dto=True` returns both.

Second, having a function return different kinds of things based on its arguments is called "type instability" and is a pattern that should be avoided. It's much better to either have these be either separate dedicated functions (e.g., `local_background_time` returns a datetime, `local_background_time_dir` returns a string) or---my recommendation---use function composition:

```python
def date2dir(dt: datetime.datetime):
  return dt.strftime(datetime_formats["directory_format"])

def local_background_time(self, window_offset, window_type) -> datetime.datetime:
  return self.__get_local_background_time__(window_type, window_offset)

#  ...
some_directory = date2dir(local_background_time(...))
```

src/swell/utilities/geos.py
```diff
@@ -375,4 +375,62 @@ def resub(self, filename, pattern, replacement):
...
+                         background_frequency: str,
+                         window_length: str,
+                         window_begin_iso: str,
+                         model: str = 'geos_marine',
+                         marine_models: list = []
+                         ) -> list:
+
+        states = []
+        self.logger.info('Generating states for model: '+model)
+        if model == 'geos_ocean' or model == 'geos_marine':
```
ashiklom:
```suggestion
        if model in ("geos_ocean", "geos_marine"):
```

src/swell/utilities/geos.py
```diff
@@ -375,4 +375,62 @@ def resub(self, filename, pattern, replacement):
...
+
+        # Generate the list of states dictionary with date and marine filename entries.
+        # The date is calculated by adding the background frequency to the window begin date.
+        # The ocn_filename is calculated by adding the background frequency to the window begin date
+        states = []
+
+        # For FGAT and 4D-Var, the first state is the background state, hence we need to
+        # skip the first state in the loop by adding "-1" to the range function.
+        for i in range(number_of_states-1):
+            i += 1
```
ashiklom: Range can take a `start` and `stop` argument. So, if you want to skip `i=0`, you can just do:

```suggestion
        for i in range(1, number_of_states):
```

src/swell/tasks/link_geos_output.py
asewnath: `self.src_dst_dict: Dict[str, Tuple[str, ...]] = {}` -> You can also instantiate a dictionary of lists using `defaultdict(list)`. This way, every time you have a new key, there will be an empty list associated with it that you can append to.

```
self.src_dst_dict = defaultdict(list)
self.src_dst_dict[src].append(dst)
```

This way, you don't have to force `dst` to be a tuple in order to add it to the `src` list

src/swell/tasks/link_geos_output.py
```diff
@@ -27,73 +30,87 @@ def execute(self):
...
+
+        if self.get_model() == 'geos_ocean' or self.get_model() == 'geos_marine':
+            self.link_mom6_history(self.src_dst_dict)
+
+            if 'cice6' in marine_models:
+                # Link CICE6 restart (iced.nc) and create SOCA input file (cice.res.nc)
+                src, dst = self.prepare_cice6()
+                self.src_dst_dict[src] = dst
```
asewnath, line 75: `self.src_dst_dict` is typed as `Dict[str, Tuple[str, ...]]` but value is set to `dst`. Is `dst` a tuple in this case?

Dooruk:
Thank you for the reviews! Before this one, the following needs to be merged:
https://github.com/GEOS-ESM/swell/pull/430

Then I will work on your suggestions.

src/swell/tasks/generate_b_climatology.py
Dooruk: Thanks, good suggestion! I will add `kwargs` to `shell_commands` utility.

src/swell/tasks/link_geos_output.py
Dooruk: Thanks, great suggestion!

src/swell/configuration/jedi/interfaces/geos_marine/model/background.yaml
Dooruk: ok, marking this as resolved for now.

src/swell/tasks/link_geos_output.py
Dooruk: I will stick with `self.` approach as this task may be handling different model outputs

Dooruk: I found out using Tuple was unnecessary, is there an advantage of using `defaultdict(list)` vs. `{}` if that's the case?

ashiklom: With `defaultdict(list)`, the following will _always_ work, for any value of `key`, regardless of whether or not the corresponding key exists in the dict. That's because `x[key]` is automatically initialized to a blank list for any missing value of `key`.

```python
x = defaultdict(list)
x[key].append(something)
```

Without `defaultdict`, the above would almost certainly throw an error, and you would need to write much uglier and more brittle code along the lines of:

```python
x = {}
x[key] = x[key] + [something] if key in x else [something]
```

Dooruk:
I'm still struggling with `kwargs` typehinting. I tried @mranst's suggestion `Union[int, str, None] = None` but that didn't work unfortunately. For now I commented it out but I will test the rest of this PR and come back to it.

Also, following @ashiklom's suggestion on using `cwd` for `subprocess`, I will not providing a default so it will be `cwd: str`. Hence, it will always throw an error if it's not defined, I think that makes sense?. This is the part I'm talking about: 

https://github.com/GEOS-ESM/swell/blob/597bbbe9c87867178a130a10c6d07418d1a212d8/src/swell/utilities/shell_commands.py#L21-L25

src/swell/utilities/data_assimilation_window_params.py
Dooruk: Good suggestion. This impacts quite a few tasks and suites so it won't be straightforward. Let me address this in a separate issue.

ashiklom:
`**kwargs` don't need to be type hinted. The main situation in which you _would_ type hint them is if all the arguments had the same type (e.g., every `**kwargs` was an integer). If that's not the case, then the type hint is confusing and, more importantly, defeats a lot of the flexibility of using `kwargs` in the first place.

Unfortunately, type hinters don't seem to be smart enough to propagate type checking through `kwargs`, but that's OK --- we shouldn't let the perfect be the enemy of the good!

If you use `**kwargs`, you don't need to propagate the `cwd` argument directly. However, if you want to propagate it anyway, I would just follow the semantics of the parent function [subprocess.run](https://docs.python.org/3/library/subprocess.html#subprocess.run) and default to `cwd: Optional(str) = None`, where `cwd = None` means use the current working directory.

src/swell/tasks/link_geos_output.py
asewnath: > I found out using Tuple was unnecessary, is there an advantage of using `defaultdict(list)` vs. `{}` if that's the case?

It depends on why it is unnecessary. If the tuple is unnecessary because you only expect one value for every key, then you don't need to use defaultdict, and you can use {} instead.

Dooruk: Unnecessary because I thought I needed to have different values sharing the same key but that turned out not to be the case..

Dooruk:
So I have one issue (more of a design flaw 🥲  ) left related to low-level design of task/suite/experiment config creation but this is good to go. For now I will set a hard-coded value in the task and move on but I can also create an issue depending on the comments.

The issue arises a key in `task_questions.yaml` or `suite_questions.yaml` needs to be used by a conditional task in `flow.cylc`.  Following is the entry in my `task_questions.yaml`. I don't see `cice6_domains` in my `experiment.yaml` after `swell create` step:

```yaml
cice6_domains:
  ask_question: true
  default_value: defer_to_model
  models:
  - geos_marine
  options: defer_to_model
  prompt: Which CICE6 domains do you wish to run DA for?
  tasks:
  - RunJediConvertStateSoca2ciceExecutable
  type: string-check-list
```

Only the `RunJediConvertStateSoca2ciceExecutable` task requires this. In `flow.cylc`, this task is used only when `cice6` is active:

```cylc
    {% if 'cice6' in marine_models %}

    [[RunJediConvertStateSoca2ciceExecutable-{{model_component}}]]
        script = "swell task RunJediConvertStateSoca2ciceExecutable $config -d $datetime -m {{model_component}}"
        platform = {{platform}}
        execution time limit = {{scheduling["RunJediConvertStateSoca2ciceExecutable"]["execution_time_limit"]}}
        [[[directives]]]
        {%- for key, value in scheduling["RunJediConvertStateSoca2ciceExecutable"]["directives"][model_component].items() %}
            --{{key}} = {{value}}
        {%- endfor %}

    {% endif %}
```
While creating `experiment.yaml`, Swell scans `flow.cylc` `[runtime]` section to find unique tasks and hence to define associated `questions (keys)` in `task_questions.yaml`. `marine_models` was not defined during this non-exhaustive search step, hence this task inside the conditional is essentially not registered. I'm talking about at this stage of the code:

https://github.com/GEOS-ESM/swell/blob/597bbbe9c87867178a130a10c6d07418d1a212d8/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L382

Here, `self.experiment_dict` contains `model_components` so tasks inside those conditionals work:

```cylc

    [[LinkGeosOutput-{{model_component}}]]
        script = "swell task LinkGeosOutput $config -d $datetime -m {{model_component}}"
    {% endfor %}
```

Anyways, this was really long explanation but happy to explain even further, I had to do some digging to find this issue.

src/swell/tasks/link_geos_output.py
```diff
@@ -53,72 +51,161 @@ def execute(self) -> None:
...
-        # Create links to GEOS history for SOCA inputs
+        # Convert background frequency to 02d format
+        duration = isodate.parse_duration(background_frequency)
+        hist_hours, remainder = divmod(duration.total_seconds(), 60*60)
+
+        return f'{int(hist_hours):02d}h'
+
+    # ----------------------------------------------------------------------------------------------
+
+    def link_cice6_history_4d(self) -> Tuple[str, str]:
```
ashiklom: This has no `return` statement, so it returns `None`, _not_ a Tuple.

```suggestion
    def link_cice6_history_4d(self) -> None:
```

src/swell/tasks/link_geos_output.py
```diff
@@ -53,72 +51,161 @@ def execute(self) -> None:
...
+
+        # Using a 4D window; CICE6 background is at the beginning of the window
+        src_history = self.cice6_history_formatter(self.bkgr_time_dto, hour_prefix)
+        dst_history = os.path.join(self.cycle_dir(), 'cice.res.' + self.bkgr_time_iso + '.nc')
+
+        self.prepare_cice6_history(src_history, dst_history)
+
+    # ----------------------------------------------------------------------------------------------
+
+    def link_cice6_history_3d(self) -> Tuple[str, str]:
```
ashiklom: Ditto here.

```suggestion
    def link_cice6_history_3d(self) -> None:
```

src/swell/tasks/link_geos_output.py
```diff
@@ -53,72 +51,161 @@ def execute(self) -> None:
...
+
+        self.prepare_cice6_history(src_history, dst_history)
 
     # ----------------------------------------------------------------------------------------------
 
-    def prepare_cice6(self) -> Tuple[str, str]:
+    def prepare_cice6_history(self,
+                              src_history: str,
+                              dst_history: str,
+                              ) -> Tuple[str, str]:
```
ashiklom: Ditto here.

```suggestion
                              ) -> None:
```

src/swell/tasks/link_geos_output.py
```diff
@@ -53,72 +51,161 @@ def execute(self) -> None:
...
+        # Add the background (beginning of the window) file
+        src = self.forecast_dir('his_' + self.bkgr_time_dto.strftime('%Y_%m_%d_%H') + '.nc')
+        dst = 'MOM6.res.' + self.bkgr_time_iso + '.nc'
+
+        # Append dst to the dictionary
+        self.src_dst_dict[src] = dst
+
+    # ----------------------------------------------------------------------------------------------
+
+    def link_mom6_history_3d(self) -> Tuple[str, str]:
```
ashiklom:
```suggestion
    def link_mom6_history_3d(self) -> None:
```

src/swell/tasks/link_geos_output.py
```diff
@@ -53,72 +51,161 @@ def execute(self) -> None:
...
-            if isinstance(dst, tuple):
-                for d in dst:
-                    self.geos.linker(src, d, self.cycle_dir())
-            else:
-                self.geos.linker(src, dst, self.cycle_dir())
+            self.geos.linker(src, dst, self.cycle_dir())
+
+    # ----------------------------------------------------------------------------------------------
+
+    def link_mom6_history_4d(self) -> Tuple[str, str]:
```
ashiklom:
```suggestion
    def link_mom6_history_4d(self) -> None:
```

src/swell/utilities/data_assimilation_window_params.py
```diff
@@ -116,7 +116,7 @@ def local_background_time_iso(self, window_offset: str, window_type: str) -> str
 
     # ----------------------------------------------------------------------------------------------
 
-    def local_background_time(self, window_offset, window_type, dto=False):
+    def local_background_time(self, window_offset, window_type, dto=False) -> str:
```
ashiklom: This can return either a string or a tuple of a string and a datetime. So, the correct type hint should be something like:

```suggestion
    def local_background_time(self, window_offset, window_type, dto=False) -> Union[str, Tuple[str, datetime]]:
```

(If we're using Python >= 3.10, then this can be written more concisely as: `str | Tuple[str, datetime]`).

While I'm pontificating, in the future, I would just drop the `dto` argument, change this function to _always_ return the background time as the second argument, and then just use tuple unpacking to absorb the unused argument whenever we don't need it.

```python
mystring, _ = self.local_background_time(window_offset, window_type)
# Python will put the datetime into the `_`, which is then silently discarded.
```

ashiklom (changes requested):

src/swell/tasks/link_geos_output.py
Dooruk: oops..

src/swell/utilities/data_assimilation_window_params.py
Dooruk: I will revisit the `dto` with the issue I created. I agree it's confusing in it's current form.

ashiklom (approved):
Someone else should review the DA aspects, but code looks good. Appreciate the documentation!

Issue #418, opened 2024-09-11
Add QR/QL/QI in linear operator of GMI


rtodling:
This is a place holder to be acted on when [PR#1257](https://github.com/JCSDA-internal/fv3-jedi/pull/1257) is approved ... 

Perhaps other MW data need similar revisions (certainly AMSR2)

rtodling:
I am close this in view of: https://github.com/GEOS-ESM/swell/pull/422

PR #422, opened 2024-09-12
Add hydrometeors

rtodling:
## Description

This adds hydrometeors and changes to the linear obs operators necessary to account for sensitivity of all-sky MW to such fields.

## Dependencies

- [https://github.com/JCSDA-internal/fv3-jedi/pull/1257 ] waiting on fv3-jedi 
- [ https://github.com/JCSDA-internal/saber/pull/939] waiting on saber

## Impact

Enhanced features and corrections and consistency w/ GSI

## Issues

[#418]  this is associated w/ the issue marked here.

## Required Change in Static Files

This will require a change to the cli_* and hyb_* gsibec 'nml' : all qr/qs references in these files must be uncommented.


rtodling:
Look around to try to understand some - unrelated to this - I came across the fact that by default, the hydrometeors radii is set to use GFDL's numbers ... the GSI numbers are an option ... The yaml's for GMI and AMRS2 in swell are using the fv3-jedi default which is not what we want - I believe. 

I ran a little test w/ GMI obs only to see what difference we get from these two schemes in comparison to GSI: I find that the difference between the two schemes is considerable smaller than the difference of either scheme w/ the actual GSI. Here is an illustration of the differences: top: GSI-minus-fv3-jedi using GSI radii; bottom: diff of JEDI increments when using either GSI or GFDF radii ...

qr_radii_schemes

The differences in the temperature are unnoticeable either way - GSI and JEDI are quite good match to each other regardless of which scheme is used - no need to show.

In any case, the consistent choice of radii is also being introduced in the changes presented here.

rtodling:
The PRs from JEDI mentioned above have been merged. I believe the changes here could now make it into SWELL

Dooruk:
@rtodling Can you set the environment `LOG_INFO` variable to 1 and run `swell test code_tests` with this branch please? There is a code testing issue.

https://geos-esm.github.io/swell/#/code_tests/code_tests

Dooruk:
This will be on-hold until a JEDI build post ~September 20th.

rtodling:
@Dooruk All tests seem to have passed for this - can we take this in? If not, can we chat?

Dooruk:
> @Dooruk All tests seem to have passed for this - can we take this in? If not, can we chat?

Did you run t1tests? I will fire it up in github actions now

rtodling:
> > @Dooruk All tests seem to have passed for this - can we take this in? If not, can we chat?
> 
> Did you run t1tests? I will fire it up in github actions now

I was looking in the doc how as I user I can fire up the tier-1 test ... but it's not quite clear to me.

Dooruk:
Doesn't this require the new JEDI build by the way?

src/swell/configuration/jedi/oops/LocalEnsembleDA.yaml
```diff
@@ -12,6 +12,9 @@ background:
   TASKFILLbackground_ensemble
 
 observations:
+  get values:
```
Dooruk, line 15: When you put anything in the OOPS part of SWELL this will impact the marine side as well if it's a generic application. So for this, we don't have a corresponding `get_values.yaml` on the `geos_ocean` side and `3dvar` is complaining. One option is that I can create an dummy `get_values.yaml` which is fine.

However when I searched for this `get values` key in the SWELL source code I see sections that modifies that behavior, see below, so we have to be careful with this:

https://github.com/GEOS-ESM/swell/blob/2b0f03bf30e2782b8441657f9be86558e63e5436/src/swell/tasks/run_jedi_ufo_tests_executable.py#L87-L93

src/swell/configuration/jedi/oops/hofx3D.yaml
```diff
@@ -10,5 +10,7 @@ state:
   TASKFILLbackground
 
 observations:
+  get values:
```
Dooruk, line 13: Same above^

src/swell/configuration/jedi/oops/variational3D.yaml
```diff
@@ -13,6 +13,8 @@ cost function:
   background error:
     TASKFILLbackground_error
   observations:
+    get values:
```
Dooruk, line 16: Same above^

src/swell/configuration/jedi/oops/variational4D.yaml
```diff
@@ -18,6 +18,8 @@ cost function:
   background error:
     TASKFILLbackground_error
   observations:
+    get values:
```
Dooruk, line 21: Same above^

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry.yaml
Dooruk, line 1: Did you accidentally changed the file permissions here?

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background.yaml
Dooruk, line 1: permission issue

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.yaml
Dooruk, line 1: Did you accidentally changed the file permissions here?

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry_inner.yaml
Dooruk, line 1: permission

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/pseudo-model.yaml
Dooruk, line 1: permission

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/r2d2.yaml
Dooruk, line 1: permission

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/stage.yaml
Dooruk, line 1: permission

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/stage_cycle.yaml
Dooruk, line 1: permission

Dooruk (changes requested):
So overall, there are issues regarding `get_values` and file permissions were changed. These changes are breaking ocean 3dvar and hofx.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry.yaml
rtodling: Ok - 1st to the file permissions: not accidental - I don't see any reason for yaml files to be executable files, unless I am missing something, am I?

src/swell/configuration/jedi/oops/hofx3D.yaml
rtodling: What's the correct way of handling this?  I am completely surprised and puzzled as to why  the UFO tasks are removing  the "get values" entries ... but then again, I don't care too much about those tests. 

So the important thing for me is to have "get values" properly appear in the Var cases ... which as far as I know they do. Now, I didn't think of that upsetting the marine apps ... so I agree that something else would need to be done to avoid that - perhaps as you say, implementing an empty "get values"

src/swell/configuration/jedi/oops/LocalEnsembleDA.yaml
rtodling: As I say below, I personally see no reason why an entry that's used in the Var should not be present in the UFO test; that makes the UFO test inconsistent w/ the Var ... but also, as I say below, I don't particularly care too much about what's being done in these tests. What I care about is for the Var opts/settings to come out correctly - and believe they do .... as look at the final yaml file in my Var runs.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry.yaml
Dooruk: Sounds good, wanted to make sure nothing was changed unintentionally.

Dooruk (approved):
all good

Issue #345, opened 2024-05-17
SE: Pin SWELL to specific JEDI commit hashes


ashiklom:
Per the last SWELL team meeting, we want to be able to "pin" SWELL to specific versions of JEDI. This will prevent SWELL and its workflows from breaking _unexpectedly_ when where are updates to JEDI.

(Note that this concept of pinning and manually updating pins is exactly how most package managers already work).

The desired end state is one where we regularly (weekly? Several times per week? Daily??) submit pull requests _only to update the JEDI commit hash_ and nothing else and test whether something new has broken _specifically because of JEDI_.

The biggest disadvantage to this is that we will, by definition, be slightly behind JEDI development (and will start to lag further behind if JEDI introduces breaking changes). But, the advantage of this is that we will have a system that is more stable and resilient to breakage from external sources, which should especially be useful for more operational systems.

ashiklom:
@asewnath Per our discussion yesterday, the proposed high-level steps for this are:

1. Modify https://github.com/GEOS-ESM/jedi_bundle/ to allow the option of building a particular commit hash of JEDI (but still default to `develop`) (pseudocode: `bundle_jedi --commit-id=a123456`)
    - Ideally, it would be great to have a single interface that can work with commit IDs, branches, or tags, resolving them automatically. I think a lot of git commands can already do this (e.g., `git checkout` works with commits, tags, and branches).
2. Modify the SWELL JEDI build task and questionary YAML (CC @CRWayman @Dooruk) to optionally take a specific JEDI commit hash to build.
3. (Pending discussion with SWELL team; CC @rtodling @rgelaro) "Flip the switch" to make building against a particular JEDI commit hash / reference the _default_ for SWELL (while retaining the user-specified option to build against the `develop` branch of JEDI).

Each of these should neatly fall into a separate pull request. NOTE: 1 and 2 above _should not affect default SWELL behavior_ (except maybe that users face one additional question with a preset default) and therefore should not affect users. The big breaking change comes in 3 when we switch the default from `develop` to a pinned commit / reference.

ashiklom:
4. Develop a simple script in SWELL for updating the JEDI commit hash to the current target of the JEDI `develop` branch. (This will make it much easier to quickly update the commit hash and submit a corresponding PR).

Dooruk:
> The desired end state is one where we regularly (weekly? Several times per week? Daily??)

[...]

PR #429, opened 2024-09-17
Pinned versions utilities and abort suppression

### LINKED ISSUE #420: Suppress abort messages in code tests
The abort messages from code tests that expect failure are alarmingly red and confusing. Need to suppress these `logger.abort` messages to better communicate that the code tests have passed.

asewnath:
## Description

- Adds `check_hash.py` utility to evaluate whether hashes in `jedi_bundle` directory align with those in `pinned_versions.yaml`
- Adds testing for `check_hash.py`
- Adds abort suppression so that swell's logger output is not shown during code tests

Note that `check_hash.py` contains a logger abort. Since most people using swell (myself included) do not have permissions in `advda` to update `jedi_bundle`, it would not make sense for a routine in swell to try to perform git checkouts to the correct hashes. It _can_ ask swell to create a local directory for `jedi_bundle`, but that defeats the purpose of having managed `jedi_bundles` in `advda`. 

Resolves #420
Addresses some concerns in #345

### DIFF
--- a/src/swell/utilities/pinned_versions/check_hashes.py
+++ b/src/swell/utilities/pinned_versions/check_hashes.py
@@ -1,0 +1,41 @@
+import os
+import yaml
+import subprocess
+from swell.utilities.logger import Logger
+
+
+def get_pinned_vers_path() -> str:
+    return os.path.split(__file__)[0]
+
+
+def check_hashes(jedi_bundle_loc: str, logger: Logger) -> None:
+
+    # Get list of directories in jedi_bundle_loc
+    dirs = os.listdir(jedi_bundle_loc)
+
+    pinned_vers_file = get_pinned_vers_path() + "/pinned_versions.yaml"
+    # Loaded pinned_versions into dict
+    with open(pinned_vers_file) as stream:
+        pinned_vers = yaml.safe_load(stream)
+
+    incorrect_hash = []
+    for repo_dict in pinned_vers:
+        repo_name = next(iter(repo_dict))
+        expected_hash = repo_dict[repo_name]["branch"]
+
+        # Get hash or branch of repo in jedi bundle
+        if repo_name in dirs:
+            cmd = ["git", "rev-parse", "HEAD"]
+            with subprocess.Popen(cmd, cwd=jedi_bundle_loc+repo_name,
+                                  stdout=subprocess.PIPE) as proc:
+                curr_hash = proc.stdout.read()
+                proc.kill()
+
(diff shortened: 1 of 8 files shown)


src/swell/utilities/pinned_versions/check_hashes.py
```diff
@@ -0,0 +1,41 @@
+import os
+import yaml
+import subprocess
+from swell.utilities.logger import Logger
+
+
+def get_pinned_vers_path() -> str:
+    return os.path.split(__file__)[0]
```
ashiklom: [Strictly speaking](https://stackoverflow.com/questions/6028000/how-to-read-a-static-file-from-inside-a-python-package), in Python >=3.7, it's better to use `importlib.resources` for stuff like this:

```python
import importlib
def get_pinned_vers_path() -> str:
  return str(importlib.resources("swell.utilities.pinned_versions") / "pinned_versions.yaml")
```

(Note that `importlib.resources` returns a `pathlib.Path` object rather than a string, hence the need for the `str` call. _Eventually_, I think it might make sense for us to use `Path` objects for working with paths, but for now, we can stick with our current approach of using strings).

...and then, further down, you can just do:

```python
pinned_vers_file = get_pinned_vers_path()
```

src/swell/utilities/pinned_versions/check_hashes.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # Get hash or branch of repo in jedi bundle
+        if repo_name in dirs:
+            cmd = ["git", "rev-parse", "HEAD"]
+            with subprocess.Popen(cmd, cwd=jedi_bundle_loc+repo_name,
+                                  stdout=subprocess.PIPE) as proc:
+                curr_hash = proc.stdout.read()
+                proc.kill()
+
+            curr_hash = curr_hash.decode("utf-8").replace('\n', '')
```
ashiklom: `rstrip` method strips all whitespace from the end (`lstrip` does it from the beginning, and `strip` does it from both sides).

```suggestion
            curr_hash = curr_hash.decode("utf-8").rstrip()
```

ashiklom (commented):
Two minor suggestions, but otherwise, looks good!

src/swell/utilities/pinned_versions/check_hashes.py
asewnath: Thanks for the link, this is good to know

ashiklom (approved):

PR #430, opened 2024-09-18
Updates for the SOCA config for updating pinned JEDI build

Dooruk:
Following some SABER updates, configuration for SOCA needs to be updated to use JEDI builds past Aug 3rd.

`Explicit_diffusion` became generic (i.e., it can be shared with other models). For SOCA, it has to use the model interface and that introduces some quirks. The correlation scales need to be named as ocean model variables (instead of `hz` or `vt`, `tocn` and `ssh`). And the `state_variable` in `background_diffusion_vt.yaml` has to only contain the correlation scale variable (`tocn` in this case). Hopefully this will be fixed in a future update.

## Dependencies:

- [x] ~Requires a new build, which I already did for August 31st and works for that. So we should first wait for https://github.com/GEOS-ESM/swell/pull/429~

@rtodling tested with an August 31st JEDI build and results are practically identical.

- [x] ~Requires Static file changes which are currently located in my local folder.~

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_ocean/model/background_diffusion_vt.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_ocean/model/background_diffusion_vt.yaml
@@ -1,0 +1,5 @@
+date: '{{local_background_time_iso}}'
+read_from_file: 1
+basename: './'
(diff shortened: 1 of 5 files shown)


DavidSRussell (approved):
Looks good to me.

PR #433, opened 2024-09-23
Using Pinned Hashes for JEDI Repositories 

### LINKED ISSUE #423: (SE) "Edit documents" link is broken in docs
It still assumes we're using the `gh-pages` branch structure, rather than having the docs live in `/docs`, so the link doesn't work.

asewnath:
## Description
Adds two options for using pinned versions in swell experiments:

1. Adds `pinned_create` jedi build method. Calling this method will clone jedi repositories with hashes in `pinned_versions.yaml`. Then it will build jedi and proceed as normal.
2. Adds `use_pinned_existing` jedi build method. This will point to `build` and `source` directories in `/discover/nobackup/projects/gmao/advda/jedi_bundles` or `/discover/nobackup/projects/gmao/advda/jedi_bundles_sles15` for milan nodes. This build is currently pinned to August 31st. The `check_hashes` script is called to check if commit hashes in this build lines up with `pinned_versions.yaml`. If it does not, it will abort. If it does, it will proceed as normal.
3. Quick change for edit documents link

Note that none of these options are default and they must be indicated through the CLI.

### Notes
- Automatically building jedi in local directory if `use_pinned_existing` fails will be addressed in another PR
- Github action will be addressed in different PR

## Issues
Resolves #423
Addresses some concerns in #345

### DIFF
--- a/src/swell/tasks/build_jedi.py
+++ b/src/swell/tasks/build_jedi.py
@@ -33,8 +33,7 @@ class BuildJedi(taskBase):
 
         # Choice to link to existing build or build JEDI using jedi_bundle
         # ----------------------------------------------------------------
-        if self.config.jedi_build_method() == 'create' or \
-           self.config.jedi_build_method() == 'pinned_create':
+        if self.config.jedi_build_method() in ('create', 'pinned_create'):
 
             # Determine which bundles need to be build
             model_components = self.get_model_components()
--- a/src/swell/tasks/build_jedi.py
+++ b/src/swell/tasks/build_jedi.py
@@ -33,44 +33,41 @@ class BuildJedi(taskBase):
 
         # Choice to link to existing build or build JEDI using jedi_bundle
         # ----------------------------------------------------------------
-        if self.config.jedi_build_method() in ('create', 'pinned_create'):
-
-            # Determine which bundles need to be build
-            model_components = self.get_model_components()
-            if model_components is not None:
-                bundles = []
-                for model_component in model_components:
-                    # Open the metadata config for interface
-                    self.jedi_rendering.add_key('npx_proc', 1)
-                    self.jedi_rendering.add_key('npy_proc', 1)
-                    self.jedi_rendering.add_key('total_processors', 1)
-                    meta = self.jedi_rendering.render_interface_meta(model_component)
-                    bundles.append(meta['jedi_interface'])
-            else:
-                bundles = get_bundles()
-
-            use_pinned = False
-            if self.config.jedi_build_method() == 'pinned_create':
-                use_pinned = True
-
-            # Generate the build dictionary
-            jedi_bundle_dict = set_jedi_bundle_config(self.config.bundles(bundles),
-                                                      jedi_bundle_source_path,
-                                                      jedi_bundle_build_path, self.platform(),
-                                                      use_pinned, 24)
-
-            # Perform the clone of JEDI repos
-            execute_tasks(['configure', 'make'], jedi_bundle_dict)
-            try:
-                execute_tasks(['configure', 'make'], jedi_bundle_dict)
-            except Exception:
-                self.logger.abort(f'A failure occurred in jedi_bundle.execute_tasks')
-
-        else:
-
+        if self.config.jedi_build_method() not in ('create', 'pinned_create'):
             self.logger.abort(f'Found \'{self.config.jedi_build_method()}\' for ' +
                               f'jedi_build_method in the  experiment dictionary. Must be ' +
                               f'\'create\' or \'pinned_create\'.')
+
+        # Determine which bundles need to be build
+        model_components = self.get_model_components()
+        if model_components is not None:
+            bundles = []
+            for model_component in model_components:
+                # Open the metadata config for interface
+                self.jedi_rendering.add_key('npx_proc', 1)
+                self.jedi_rendering.add_key('npy_proc', 1)
+                self.jedi_rendering.add_key('total_processors', 1)
+                meta = self.jedi_rendering.render_interface_meta(model_component)
+                bundles.append(meta['jedi_interface'])
+        else:
+            bundles = get_bundles()
+
+        use_pinned = False
+        if self.config.jedi_build_method() == 'pinned_create':
+            use_pinned = True
+
+        # Generate the build dictionary
+        jedi_bundle_dict = set_jedi_bundle_config(self.config.bundles(bundles),
+                                                  jedi_bundle_source_path,
+                                                  jedi_bundle_build_path, self.platform(),
+                                                  use_pinned, 24)
+
+        # Perform the clone of JEDI repos
+        execute_tasks(['configure', 'make'], jedi_bundle_dict)
+        try:
+            execute_tasks(['configure', 'make'], jedi_bundle_dict)
+        except Exception:
+            self.logger.abort(f'A failure occurred in jedi_bundle.execute_tasks')
 
 
 # --------------------------------------------------------------------------------------------------
--- a/src/swell/tasks/build_jedi.py
+++ b/src/swell/tasks/build_jedi.py
@@ -33,7 +33,8 @@ class BuildJedi(taskBase):
 
         # Choice to link to existing build or build JEDI using jedi_bundle
         # ----------------------------------------------------------------
-        if self.config.jedi_build_method() == 'create':
+        if self.config.jedi_build_method() == 'create' or \
+           self.config.jedi_build_method() == 'pinned_create':
 
             # Determine which bundles need to be build
             model_components = self.get_model_components()
@@ -49,10 +50,15 @@ class BuildJedi(taskBase):
             else:
                 bundles = get_bundles()
 
+            use_pinned = False
+            if self.config.jedi_build_method() == 'pinned_create':
+                use_pinned = True
+
             # Generate the build dictionary
             jedi_bundle_dict = set_jedi_bundle_config(self.config.bundles(bundles),
                                                       jedi_bundle_source_path,
-                                                      jedi_bundle_build_path, self.platform(), 24)
+                                                      jedi_bundle_build_path, self.platform(),
+                                                      use_pinned, 24)
 
             # Perform the clone of JEDI repos
             execute_tasks(['configure', 'make'], jedi_bundle_dict)
--- a/src/swell/tasks/build_jedi.py
+++ b/src/swell/tasks/build_jedi.py
@@ -20,7 +20,7 @@ from swell.utilities.build import set_jedi_bundle_config, build_and_source_dirs
 
 class BuildJedi(taskBase):
 
-    def execute(self):
+    def execute(self) -> None:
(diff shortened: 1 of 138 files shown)


src/swell/tasks/build_jedi.py
```diff
@@ -33,7 +33,8 @@ def execute(self) -> None:
 
         # Choice to link to existing build or build JEDI using jedi_bundle
         # ----------------------------------------------------------------
-        if self.config.jedi_build_method() == 'create':
+        if self.config.jedi_build_method() == 'create' or \
+           self.config.jedi_build_method() == 'pinned_create':
```
asewnath:
```suggestion
           self.config.jedi_build_method() in ('create', 'pinned_create')
```

src/swell/tasks/build_jedi.py
```diff
@@ -65,7 +70,7 @@ def execute(self) -> None:
 
             self.logger.abort(f'Found \'{self.config.jedi_build_method()}\' for ' +
                               f'jedi_build_method in the  experiment dictionary. Must be ' +
-                              f'\'create\'.')
+                              f'\'create\' or \'pinned_create\'.')
```
ashiklom, line 39: Align the happy path to the left edge! A.k.a., Return early!

In general, rather than really large if-else blocks like this...

```python
if good_condition:
  # do...
  # a...
  # bunch...
  # of...
  # stuff...
else: 
  raise SomeError(...)
```

a much better approach is something like:

```python
if not good_condition:
  raise SomeError()

# Do...
# a...
# bunch...
# of...
# stuff.
# ^^ Note: This block does *not* have to be indented anymore!
```

Here, I would flip the conditional to `if jedi_build_method not in ('create', 'pinned_create')`, put the error there, and then unindent the rest of the code (the "happy path").

(I know this isn't your code, but can't hurt to make cheap style improvements when we can 😄 ).

src/swell/test/platform_tests/__init__.py
```diff
@@ -0,0 +1,9 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+import os
+
+repo_directory = os.path.dirname(__file__)
```
ashiklom: An `__init__.py` file is required for Python to pick this up as a module. But you don't actually need anything in the `__init__.py` file.

I did a quick scan and we never actually use the `repo_directory` variable anywhere (and, even if we did, the more correct thing to do would be to use `importlib.resources` anyway).

At some point, we should go through and remove all this unnecessary code. But, in the meantime, we can at least stop adding new instances of it.

```suggestion
```

src/swell/utilities/build.py
```diff
@@ -57,6 +60,7 @@ def set_jedi_bundle_config(
     path_to_source: str,
     path_to_build: str,
     platform: str,
+    use_pinned: bool,
```
ashiklom, line 63: I don't necessarily disagree that we should include this as a required argument --- it has a major impact on what Swell does, so it makes sense to force users to specify this.

However, just noting that introducing new _required_ arguments to things is a breaking change, and, in general, should be considered carefully and communicated accordingly.

We can keep this as is, but just flagging it for future consideration.

src/swell/tasks/clone_jedi.py
```diff
@@ -40,7 +41,15 @@ def execute(self) -> None:
             # Link the source code directory
             link_path(self.config.existing_jedi_source_directory(), jedi_bundle_source_path)
 
-        elif self.config.jedi_build_method() == 'create':
+        elif self.config.jedi_build_method() == 'use_pinned_existing':
```
ashiklom, line 43: Somewhat related to my earlier comment about the "happy path", it may be a good idea to put the check for valid values of the `jedi_build_method` early and then, separately, to go through the conditional execution. I.e., Something like:

```python
build_method = self.config.jedi_build_method()

if not build_method in ("use_existing", "use_pinned_existing", "create", "pinned_create"):
  logger.abort("...")

# Now, proceed with the code:
if build_method == "use_existing":
  ...
elif ... # and so on
```

This makes it very easy to quickly for other developers to see what all the valid values of `build_method` are.

Since this pattern occurs in many places, an even better thing to do would be to define a `JEDI_VALID_BUILD_METHODS` constant in _one_ place somewhere and then check `jedi_build_method` against that.

Even better is to do this check as part of the `self.config` initialization. Define which values of `jedi_build_method` are valid, and then throw an error immediately if a user tries to set that to anything else. Then, we don't have to keep redundantly checking that this is valid everywhere else in the Swell code.

For inspiration, I suggest reading this: [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). The language there is Haskell, but the underlying idea --- use the most precise data type you can --- is universal.

ashiklom (changes requested):
Nice work!

In addition to the stuff in my review, one more requested change (I couldn't add a suggestion because it's too far from any code you touched): In `src/swell/tasks/task_questions.yaml` line 466 (definition of `jedi_build_method`), we need to add the new valid options.

Also, if we do that, is there a way to _enforce_ a particular set of options at runtime? I.e., When you go to set `self.config.jedi_build_method`, if it's set to something invalid, it'll fail immediately? Not sure if the `string-drop-list` or some other `type` will do that for us. In any case, if there's a way to easily do that, we should do it, because that will allow us to remove lots of `if self.config.jedi_build_method() in (...)` code blocks.

src/swell/tasks/clone_jedi.py
ashiklom: This idea is also why I think Python's dataclasses (and Pydantic data models) are such a good idea. With dataclasses, we can specify valid values of `self.config.jedi_build_method()` through an `Enum` type annotation (though that won't do any runtime checking). With Pydantic, we do that too, but we _also_ get runtime checking and clear errors built in (though we have to eat the cost of having another external dependency).

src/swell/utilities/build.py
asewnath: Sure, I can make changes like this more obvious in the PR description. The flag communicates to `set_jedi_bundle_config` that it needs to create it's jedi bundle build config with the pinned versions that swell tracks. If there's a better way of doing this please let me know

src/swell/tasks/build_jedi.py
asewnath: TIL what happy path is 🥲

Dooruk (dismissed):
Thanks, looks good!

We just need to ensure the consistency of `pinned_versions.yaml` and the following locations:

https://github.com/GEOS-ESM/swell/blob/597bbbe9c87867178a130a10c6d07418d1a212d8/src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml#L10-L14

src/swell/utilities/build.py
ashiklom: TLDR: No changes requested here; this can be resolved 😄 

I think this is fine. It's also internal code, so it's somewhat less important than a change to the exposed user interface (though, note that unlike some other languages like Java, Python has no formal concept of public vs. private methods/attributes and manages these things purely by convention and best practices. There's no such thing as a truly private Python method in the same way as there is in Java, so you never know who is relying on your internal implementation details!).

I wouldn't handle this any differently than you did. I was just making a general comment that changing the _required_ part of a function signature is a common kind of breaking change. Breaking changes are fine, especially in relatively early phases of software development. Enforcing strict backwards compatibility too early can lock us into suboptimal design choices.

That said, what we need to do at _some_ point with Swell is adopt proper [Semantic Versioning](https://semver.org/), possibly paired with something like [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). Since we squash-merge everything anyway, we would only have to enforce something like Conventional Commits at the PR level rather than at the individual commit level.

asewnath:
> Thanks, looks good!
> 
> We just need to ensure the consistency of `pinned_versions.yaml` and the following locations:
> 
> https://github.com/GEOS-ESM/swell/blob/597bbbe9c87867178a130a10c6d07418d1a212d8/src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml#L10-L14

I modified the platform task_question.yaml files to point to pinned version jedi bundles for SLES12 and SLES15.

src/swell/utilities/build.py
asewnath: Thanks for the information!

asewnath:
> Also, if we do that, is there a way to _enforce_ a particular set of options at runtime? I.e., When you go to set `self.config.jedi_build_method`, if it's set to something invalid, it'll fail immediately? Not sure if the `string-drop-list` or some other `type` will do that for us. In any case, if there's a way to easily do that, we should do it, because that will allow us to remove lots of `if self.config.jedi_build_method() in (...)` code blocks.

In `task_questions.yaml`, I modified `jedi_bundle_method` to include the new valid options,  and then I changed it to `string-drop-list` which works well for constraining the user to only valid build methods. If there are any remaining `if self.config.jedi_build_method() in (...)` code blocks, it's because the task itself is checking for specific build methods and aborts if other valid methods were provided. For example, `BuildJediByLinking` aborts if the user has selected either the `create` or `pinned_create` build method.

src/swell/tasks/clone_jedi.py
```diff
@@ -37,20 +37,16 @@ def execute(self) -> None:
         # Choice to link to existing build or build JEDI using jedi_bundle
         # ----------------------------------------------------------------
         if self.config.jedi_build_method() == 'use_existing':
-
             # Link the source code directory
             link_path(self.config.existing_jedi_source_directory(), jedi_bundle_source_path)
 
-        elif self.config.jedi_build_method() == 'use_pinned_existing':
-
+        if self.config.jedi_build_method() == 'use_pinned_existing':
```
ashiklom: This one should probably be `elif`, right? Since we're _either_ linking `jedi_bundle_source_path` to the pinned source directory _or_ the existing source directory.

asewnath: `CloneJedi` also handles the other jedi build methods: `if self.config.jedi_build_method() in ('create', 'pinned_create')`. I'm not sure what the `else` statement would be after all the `elif` statements since we are assuming the user passes a valid jedi build method

ashiklom: So I think this is a matter of style. `if + elif` explicitly communicates that the statements are related and the underlying clauses are mutually exclusive --- "either X or Y or Z".

https://stackoverflow.com/questions/22782785/when-to-use-if-vs-elif-in-python

My recommendation would be to use `elif` here to clearly communicate that `jedi_build_method` can only be one of those things and these are mutually exclusive clauses.

You don't need a final `else` statement --- it's perfectly valid to have a bunch of `elif` statements without a concluding `else`.

ashiklom (approved):

PR #437, opened 2024-10-02
LGETKF: add function to tune horizontal localization for each observer

metdyn:
## Description

This PR adds a mechanism to customize horizontal localization (e.g.,  Gaspari-Cohn) for obs in LGETKF.

- add a directory /jedi/interfaces/geos_atmosphere/observations/localization
- each yaml in that directory is used to update localization for the corresponding observer


metdyn:
Offline for improvement

Dooruk:
A quick question, did we confirm having `localization` in observation YAMLs break the Variational applications?

metdyn:
> A quick question, did we confirm having `localization` in observation YAMLs break the Variational applications?
@Dooruk, the added directory `jedi/interfaces/geos_atmosphere/observations/localization` is used by local ensemble da only.   And other modifications are in 
```
M       src/swell/tasks/run_jedi_local_ensemble_da_executable.py
M       src/swell/test/suite_tests/localensembleda-tier1.yaml
```
This PR does not affect variational runs.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -171,9 +167,24 @@ def execute(self) -> None:
...
         # Include ensemble localizations and halo types with each observation
         # -------------------------------------------------------------------
-        for observer in jedi_config_dict['observations']['observers']:
-            observer.update({'obs localizations': localizations})
-            observer['obs space'].update({'distribution': {'name': 'Halo', 'halo size': 5000e3}})
+
+        swell_path = get_swell_path()
+        localization_path = os.path.join(swell_path,
+                                         f'configuration/jedi/interfaces/geos_atmosphere'
+                                         f'/observations/localization')
```
Dooruk, line 174: Could we use the `importlib` to define `path` here?

You can see Alexey's justification here:

https://github.com/GEOS-ESM/swell/pull/429#discussion_r1763612715

Dooruk (changes requested):
This is good to go with two minor changes. One is in the review the other is a simple change in this branch here, that needs to be added to run `3dvar_atmos` on Milan nodes:

`feature/tasks_per_node`

Could you make that line change in this branch so we could merge this PR? @rtodling do you have any comments?

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: possible, but we can do it more systematically later on.

Dooruk (approved):

Dooruk:
I don't want to hold back this any longer, thanks @metdyn for working on this.

@mranst just FYI the `3dvar` suite doesn't work on SLES15 in the `gmao_ci` context so tier1 test will break, but `3dvar` suite works locally on my end with defaults. I'm a bit puzzled. I will merge this PR and look into the issue, I suspect something to do with the node allocations.

PR #444, opened 2024-10-21
Introducing GMAO ODAS obs sources

Dooruk:
Added currently available ODAS observations to SWELL. Reminder that in the long term `geos_ocean` will go away and I will take out inactive obs. YAMLs. SOCA naming sprint changes are included.

## Important changes:
Two changes regarding `get_observations.py` that are important to emphasize (FYI @rtodling @gmao-wgu @mranst @ashiklom ):

- R2D2 `fetch` method now includes a `ignore missing: True` flag, so it just creates a warning if the observation file is missing. This is necessary because we want Variational Bias Coeff.s to propagate regardless of the missing obs. This will need further testing by ADAS (@rtodling)

- Since there could be multiple providers for observations (`odas` and `gdas_marine` (EMC) for marine) I switched `obs_provider` key to be a check-list so `get_observations.py` can browse multiple locations. There could be better handling of this in the future with R2D2v3 but this is where we are for now.

### Some Quick Plots:
Showing scatter plots for ARGO data jedi vs h(x)  salinity and temperature (black dots failed the QC).

| Salinity | Temperature |
| --- | --- |
| Image 1 |  |

FYI @DavidSRussell, you should be able to test with these YAMLs (by setting `odas` as the obs provider) and in theory it should be easy to add SWOT altimeter data to the list.

- Filters in YAMLs need to be modified according to our needs, @lren20 can help with this part.


rtodling:
Quite nice!!!

DavidSRussell:
I'm getting a `cannot find field sea_ice_area_fraction` error in job.out. I noticed there is a variable called `sea_area_fraction` elsewhere; maybe that's a conflict?

Dooruk:
> I'm getting a `cannot find field sea_ice_area_fraction` error in job.out. I noticed there is a variable called `sea_area_fraction` elsewhere; maybe that's a conflict?

Hmm, can you check your `fields_metadata.yaml` in the `geos_ocean` folder? There should be `sea_area_fraction`

DavidSRussell:
> > I'm getting a `cannot find field sea_ice_area_fraction` error in job.out. I noticed there is a variable called `sea_area_fraction` elsewhere; maybe that's a conflict?
> 
> Hmm, can you check your `fields_metadata.yaml` in the `geos_ocean` folder? There should be `sea_area_fraction`

There is `sea_ice_area_fraction` under `ice state variables with no categories` and `sea_area_fraction` under `built-in derived variables that you don't need to worry about`. By the way, I'm pretty sure I got a similar error before whenever `sea_ice_area_fraction` wasn't commented out in the observation yamls.

Dooruk:
> There is `sea_ice_area_fraction` under `ice state variables with no categories` and `sea_area_fraction` under `built-in derived variables that you don't need to worry about`. By the way, I'm pretty sure I got a similar error before whenever `sea_ice_area_fraction` wasn't commented out in the observation yamls.

Derived variables one is the GeoVals, and `sea_area_fraction` in particular is something that we have been using quite a bit. Just to make sure, you are not pointing to your own SwellStatic files correct?

Dooruk:
Ok, on a second look, that GeoVal (`sea_ice_are_fraction`) is used by `sst_merra2` and that would require a `sea_ice` state input. So you are right, if you turn that filter off it should run.

DavidSRussell:
OK, after commenting out those lines and changing the analysis variable names in the override yaml, it seems to be working fine.

DavidSRussell (approved):

PR #446, opened 2024-10-22
Adding Instructions on how to use `uv` and `venv` for Swell installation

Dooruk:
I used @ashiklom's instructions and included them in the documentation. I think `py_lmod_installer` might be confusing for some users so hope this is useful!

Also added a few Github links, feel free to suggest more sources.

### DIFF
--- a/docs/platforms/discover/installing_swell_uv_venv.md
+++ b/docs/platforms/discover/installing_swell_uv_venv.md
@@ -23,7 +23,7 @@ mod_swell() {
   elif [[ $KERNEL_VERSION =~ "5.14" ]]; then
     source /discover/nobackup/projects/gmao/advda/swell/jedi_modules/modules-intel-sles15
   else
-    echo "Unkonwn platform: $PLATFORM"
+    echo "No matching platform for kernel version: $KERNEL_VERSION"
   fi
 }
 ```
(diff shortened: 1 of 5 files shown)


mranst (commented):
Nice! This is the method I used when installing swell for the first time, I did find it more straightforward as a new user than the other method.

docs/platforms/discover/installing_swell_uv_venv.md
```diff
@@ -0,0 +1,56 @@
...
+```bash
+mod_swell() {
+  module purge
+  KERNEL_VERSION=$(uname -r)
+  if [[ $KERNEL_VERSION =~ "4.12" ]]; then
+    source /discover/nobackup/projects/gmao/advda/swell/jedi_modules/modules-intel
+  elif [[ $KERNEL_VERSION =~ "5.14" ]]; then
+    source /discover/nobackup/projects/gmao/advda/swell/jedi_modules/modules-intel-sles15
+  else
+    echo "Unkonwn platform: $PLATFORM"
```
ashiklom: Bug! That's my bad.

```suggestion
    echo "No matching platform for kernel version: $KERNEL_VERSION"
```

ashiklom (dismissed):
One minor fix (which I think is my bad for giving bad instructions), but otherwise, looks good -- thanks for submitting this!

Dooruk:
This still needs an approval. I can't merge without approval 🙅

asewnath (approved):

rtodling:
@ashiklom I'd like a chance to try this out before this gets in.  I don't understand why something that seemed to have worked in the past no long seems to work - I insulated a version yesterday - the traditional way of installing - following the direction in SWELL doc. Then, I close my windows, the next day when I came and reloaded the modules - nothing worked - swell is not in the path. 

I will try th uv way to see if any better.

ashiklom:
@rtodling Let me know if you run into trouble! There's a lot of ways these different things can go wrong, depending on exactly what's in your `.bashrc`, etc. If it doesn't work, I could hop on a Teams call to try to debug interactively.

PR #448, opened 2024-10-24
Missing optional parameters in experiment.yaml for 3dfgat_cycle

### LINKED ISSUE #440: (SE) Optional tasks in Cylc suite do not trigger keys to be added to `experiment.yaml`
The issue arises when a key in `task_questions.yaml` or `suite_questions.yaml` needs to be used by a conditional task in `flow.cylc`, example below.

Following is the entry in my `task_questions.yaml`. I don't see `cice6_domains` in my `experiment.yaml` after `swell create` step:

```yaml
cice6_domains:
  ask_question: true
  default_value: defer_to_model
  models:
  - geos_marine
  options: defer_to_model
  prompt: Which CICE6 domains do you wish to run DA for?
  tasks:
  - RunJediConvertStateSoca2ciceExecutable
  type: string-check-list
```

Because only the `RunJediConvertStateSoca2ciceExecutable` task requires this. In `flow.cylc`, this task is used only when `cice6` is active:

```cylc
    {% if 'cice6' in marine_models %}

    [[RunJediConvertStateSoca2ciceExecutable-{{model_component}}]]
        script = "swell task RunJediConvertStateSoca2ciceExecutable $config -d $datetime -m {{model_component}}"
        platform = {{platform}}
        execution time limit = {{scheduling["RunJediConvertStateSoca2ciceExecutable"]["execution_time_limit"]}}
        [[[directives]]]
        {%- for key, value in scheduling["RunJediConvertStateSoca2ciceExecutable"]["directives"][model_component].items() %}
            --{{key}} = {{value}}
        {%- endfor %}

    {% endif %}
```

While creating `experiment.yaml`, Swell scans `flow.cylc` `[runtime]` section to find unique tasks and hence to define associated `questions (keys)` in `task_questions.yaml`. `marine_models` was not defined during this non-exhaustive search step, hence this task inside the conditional is essentially not registered. I'm talking about at this stage of the code:

https://github.com/GEOS-ESM/swell/blob/597bbbe9c87867178a130a10c6d07418d1a212d8/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L382

Here, `self.experiment_dict` contains `model_components` so tasks inside those conditionals work:

```cylc

    [[LinkGeosOutput-{{model_component}}]]
        script = "swell task LinkGeosOutput $config -d $datetime -m {{model_component}}"
    {% endfor %}
```

Anyways, this was really long explanation but happy to explain even further, I had to do some low-level digging to find this issue.

_Originally posted by @Dooruk in https://github.com/GEOS-ESM/swell/issues/419#issuecomment-2405472031_

mranst:
## Description

This (maybe just partially) addresses issue #440 . In the case of `3dfgat_cycle`, which seems to be the example mentioned in the issue, the consideration of `cice6_domain` in the experiment file is conditional on "cice6" being present in `marine_models`, as specified in that suite's `flow.cylc`.

https://github.com/GEOS-ESM/swell/blob/2b0f03bf30e2782b8441657f9be86558e63e5436/src/swell/suites/3dfgat_cycle/flow.cylc#L221-L232

In setting up the experiment, two sweeps of `flow.cylc` are, done using the function `template_string_jinja2`, which compares `flow.cylc` and a dictionary set up by suite and task questions. `marine_models` is a model-dependent variable, so it's actually addressed by the second, more exhaustive sweep.

https://github.com/GEOS-ESM/swell/blob/2b0f03bf30e2782b8441657f9be86558e63e5436/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L431-L439

The issue here is that the `marine_models` specified in the question dictionary are actually part of a sub-dictionary, so it's not resolved in the jinja2 sweep. This change just points `marine_models` to the right place, so. `cice6_domains` is triggered to add to `experiment.yaml`. I think this issue is linked to a broader discussion of the configuration files, so I'm not sure if it completely addresses the issue, but the behavior in this case seems to be fixed


Dooruk (approved):
Thanks @mranst! Sorry this was caused by a tiny bug on my part. I agree with your comment on the more advanced configuration changes in the future but this resolves #440, since there is already #314.

On another note, is there a practical reason to work on your fork?

mranst:
> Thanks @mranst! Sorry this was caused by a tiny bug on my part. I agree with your comment on the more advanced configuration changes in the future but this resolves #440, since there is already #314.
> 
> On another note, is there a practical reason to work on your fork?

Why have a fork of swell that I contribute to rather than create a branch on the main repository? There's not really a reason other than that's how I've done things in the past. I've never really thought much about it, admittedly I don't yet have the strongest confidence on how things are done in GitHub

ashiklom:
> On another note, is there a practical reason to work on your fork?

Forking to your own GitHub account and then submitting pull requests from that is generally the "default" open source contribution workflow, because it doesn't require adding people as collaborators to a GitHub repository, and liberates folks to experiment more with branch naming, repo structure, etc. (because you don't have to worry about repository-specific things like branch protection, CODEOWNERS, etc.).

I think the main reason we prefer to work directly off the `GEOS-ESM` repo is that we don't have the ability to trigger Tier 1 tests via GH Actions for forks. This would make sense if our Tier 1 tests ran automatically when PRs were submitted. But, since our system only allows a small number of specific people to trigger those workflows, and only by hand, I think that policy is probably security overkill (because you should do a code review before triggering that action anyway, especially for unfamiliar contributors). But that's a question for @jardizzo.

Dooruk:
Tier 1 actions is the main reason, but my question is mostly practical. If we were to collaborate on a branch I can't push changes directly to remote personal forks. I don't think I can even suggest as a PR (?).

Huh, I was able to merge develop to this fork on Github web. I didn't know I could do that, I thought it was going to reject.

jardizzo:
Regarding forks and workflows: I'm not an expert on what can be done with the forks except to say that it makes sense that workflows are not going to launch once the repo is forked. I'm a little concerned that I have not adequately conveyed the workflow mechanism however. There is no limit on the range of triggers for the workflows. We can change the workflows to automatically run no matter who makes the PR request. Manual triggers are an option but not at the exclusion of automatic triggers. Dan had the idea that we could make the workflows require a label assignment as part of the credentials for triggering a workflow. We can remove that requirement and give permission to all or a large group of users to execute the on-prem workflows. I don't think we want too many Tier2+ workflows executing however. Just the cheaper Tier-1 workflows. I'm happy to discuss more. I will review the trigger configuration now.

jardizzo:
Here are the current triggers allowed:

on:
  push:
    branches:
      - develop
  workflow_dispatch:

The "workflow_dispatch" is the manual trigger. We can add pull_request and we can add constraints on what branches are allowed for each trigger (or branches to exclude). I can then add users to the NAMS check that we implemented. That will enable more users to execute the workflow.

Issue #428, opened 2024-09-17
(SE) Cleaning up Logger implementation


mranst:
In adding type hinting for swell, we discovered that the implementation of `Logger` could be improved:

- Implement `Logger` (from `swell.utilities.logger`) as an inherited class of `logging.Logger`. This would simplify annotations and make it possible for swell code to simply use `logging.Logger` where sufficient

This would resolve possible confusion in cases such as in `slurm_test.py`, which passes a `logging.Logger` object to `SWELLError`, where most other functions specifically expect a swell `logger`.

Dooruk:
Pull #450 closed this right?

ashiklom:
Yes! Good catch.

PR #450, opened 2024-10-28
Swell logger extends logging.Logger

mranst:
## Description

#428 – This PR implements the swell `logger` as an extension of the `logging` library. This means that Swell loggers are now handled in a similar way to the base library.

The Logger class in the `logging` library is invoked through the `getLogger` method, rather than directly. This method goes through a manager class, which handles initialized logger objects. This PR introduces the `get_logger` method, which overrides the base `logging.Manager` object so it uses the swell Logger. Functions from `Logger` such as `info()` override the base functions, so the behavior doesn't change. Invoking `Logger` directly will still work, but if accepted, logger instances should ideally be invoked using `get_logger` in the future.


src/swell/utilities/logger.py
```diff
@@ -33,11 +35,12 @@
 # --------------------------------------------------------------------------------------------------
 
 
-class Logger:
+class Logger(logging.Logger):
 
     def __init__(self, task_name: str) -> None:
+        super().__init__(task_name)
```
ashiklom: Strictly speaking, if you are doing inheritance, you should obey the [Liskov Substitution Principle](https://stackoverflow.blog/2021/11/01/why-solid-principles-are-still-the-foundation-for-modern-software-architecture/#h2-35aa7e2297150); i.e., if a `swell.Logger` object is a subclass of `logging.Logger`, then you should be able to use a `swell.Logger` _anywhere_ where you can use ` logging.Logger` object.

In practice, the specific complaint here is that the `__init__` methods for `logging.Logger` and `swell.logging.Logger` have incompatible argument signatures. Something like this, `swell.logging.Logger(name="this", level="DEBUG")`, will throw an error, but under a strict inheritance hierarchy, this should work.

A simple fix is to (1) change `task_name` to just `name` (for consistency with `logging.Logger` and (2) accept `**kwargs` as an optional argument that gets passed on via `super().__init__(name, **kwargs)`.

src/swell/utilities/logger.py
```diff
@@ -33,11 +35,12 @@
 # --------------------------------------------------------------------------------------------------
 
 
-class Logger:
+class Logger(logging.Logger):
```
ashiklom, line 38: Now that we are using `get_logger` everywhere, do we actually need our own class at all? I.e., Can we just use `get_logger` along with appropriate `logging.Logger` internals to create `logging.Logger` objects that do close to what we want? If so, that would significantly simplify things.

I think you should be able to use the built in `logging.Logger` to customize the formatting of messages for specific levels.

The `get_logger` method would also be responsible for setting the appropriate log level based on environment variables or wherever else we're currently getting that information.

src/swell/utilities/logger.py
```diff
@@ -158,6 +161,29 @@ def assert_abort(self, condition: bool, message: str, wrap: bool = True) -> None
 
     def input(self, message: str) -> None:
 
-        input(' '+self.task_name+': '+message + ". Press any key to continue...")
+        input(' '+self.name+': '+message + ". Press any key to continue...")
```
ashiklom: I think this is the only logger method that doesn't have a direct analog to a `logging.Logger` method. Everything else should "just work" with `logging.Logger` objects.

But, IMHO, this is also an inappropriate use of `logger` and should really be its own function. Worth checking how much we actually use this and, if we never use it, I would just silently deprecate it. If we _do_ use it, then I suggest a separate function like, `prompt_confirmation(...)`.

ashiklom (commented):
Nice work!

Though, now that you've gone through and replaced `Logger(...)` with `get_logger(...)` throughout (which I approve!), I think we might as well consider deprecating our custom `Logger` class altogether and just teaching `get_logger` to create `logging.Logger` objects with the semantics we expect.

mranst:
> Nice work!
> 
> Though, now that you've gone through and replaced `Logger(...)` with `get_logger(...)` throughout (which I approve!), I think we might as well consider deprecating our custom `Logger` class altogether and just teaching `get_logger` to create `logging.Logger` objects with the semantics we expect.

I think this makes a lot of sense, pretty much everything should be able to go through `logging` internals. The one exception might be the `abort` function, which seems very convenient and doesn't seem to have an obvious alternative in `logging` (though I'm still looking into it). At the very least, I think I can accomplish the same thing with a much simpler extension of `Logger` than I currently have.

mranst:
I tried to implement some more elegant solutions to this problem, but ended up sort of going in a circle. Logging actually doesn't seem to have very complex formatting for message levels. The implementation now at least makes better use of inheritence for the logging levels. 

The customization for detail is simpler now. Setting the LOGLEVEL environment variable will control the minimum priority logging level, and by default it will include messages at INFO or above. This is in line with how logging usually works.

Dooruk:
Is this still active and pending for a review? Figure it might have been forgotten in other numerous PRs 😃

mranst:
> Is this still active and pending for a review? Figure it might have been forgotten in other numerous PRs 😃

Yeah, I had sort of forgotten about it too. I think I had it where I wanted it on the latest commit

src/swell/test/code_tests/test_pinned_versions.py
```diff
@@ -25,5 +25,9 @@ def test_wrong_hash(self) -> None:
         abort_message = "Wrong commit hashes found for these repositories in jedi_bundle: [oops]"
         # Run check hash (expect abort)
         with self.assertRaises(SystemExit) as abort, suppress_stdout():
+            # Suppress logging output for test
+            log_level = logger.level
+            logger.setLevel(60)
```
ashiklom, line 33: Magic number. Can you add a comment to what `60` means here?

src/swell/utilities/logger.py
ashiklom: My main outstanding comment, which still stands. This is a great opportunity to purge a chunk of code if we can...but if it's too much work (for this PR), we can punt it to a future PR.

ashiklom: Never mind --- I just saw your note addressing this:

https://github.com/GEOS-ESM/swell/pull/450#issuecomment-2476452192

So, we can resolve this and potentially revisit later (once we are ready to eliminate `logger.abort` in favor of proper Pythonic exception handling, which I think we should do...but at a later date).

ashiklom (dismissed):
Pending one tiny fix (add a comment about what `loglevel(60)` means), and assuming this passes all our tests, we can get this merged.

Eventually, I still would like us to remove our own `logging.Logger` class in favor of using Python's builtin logger only where appropriate (for logging; *not* for error handling or taking user input) and using more appropriate tools for managing other things. But for now, let's get this in and tackle the rest of the problem later.

mranst:
> Pending one tiny fix (add a comment about what `loglevel(60)` means), and assuming this passes all our tests, we can get this merged.
> 
> Eventually, I still would like us to remove our own `logging.Logger` class in favor of using Python's builtin logger only where appropriate (for logging; _not_ for error handling or taking user input) and using more appropriate tools for managing other things. But for now, let's get this in and tackle the rest of the problem later.

Thanks, I've added a comment clarifying the meaning

ashiklom (approved):

PR #463, opened 2024-11-22
Add missing surface type to linear obs op of MW

### LINKED ISSUE #462: Linear obs operator for MW needs to be given all surface types
This has been found while looking at the SSMI/S issue #460 - the Tskin increment due to SSMI is (obviously) insignificant small in GSI ... and it should be such in JEDI, but in the latter the increment is way too small - and indeed not quite as it should look.

here is a fig:

tskin-ssmis-gs-Xjedi

notice the increments have been blown up by a huge factor (1e5) just so something shows up.

This is what happens when all surfaces are passed to the linear observer --- this is a fix that came in when the T-skin analysis was added to JEDI/GSIBEC - these changes should have been made in when the T-skin was added to  SWELL, but they fell thru the cracks.

This is what happens w/ the fix:

tskin-ssmis-gsiXjedi-fix

The T-skin increments in JEDI are smaller that those from GSI - but the change makes the two results consistent.

rtodling:
## Description

This  add the missing surface types to the linear obs operators of the  MW instruments.

See issue #462 

## Dependencies

NONE

## Impact

NONE

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_aqua.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_aqua.yaml
@@ -20,7 +20,7 @@ obs operator:
     CoefficientPath: '{{crtm_coeff_dir}}/'
   linear obs operator:
     Absorbers: [H2O,O3]
-    Surfaces: [Water_Temperature]
+    Surfaces: [Water_Temperature,Land_Temperature,Ice_Temperature,Snow_Temperature]
 
 obs bias:
   input file: '{{cycle_dir}}/amsua_aqua.{{background_time}}.satbias.nc4'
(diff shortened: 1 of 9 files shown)


gmao-wgu (approved):

rtodling:
@Dooruk Could you please run tier1 test on this and see if we can get this in? Thanks

Dooruk:
> @Dooruk Could you please run tier1 test on this and see if we can get this in? Thanks

Hi @rtodling, two things:

1) I triggered it but you can **locally** test Tier1 with @ashiklom and @mranst's recent changes. Otherwise, if it's just a matter of running tier1 tests with a certain branch connected to a PR @mranst can help as well.
2) Can you link the relevant issue to this PR so that gets automatically closed? If you click on the "Development" on the right hand side of this page it will allow you to browse active issues

Dooruk (approved):
Note: `3dvar` fill fail due to a known issue.

PR #464, opened 2024-11-22
Set tropopause pressure opt consistent w/ GSI

rtodling:
## Description

This has been found sometime ago to be missing setting in SWELL. To be consistent w/ GSI we need to the Model2GeoVal to the formulation of GSI to calculate tropopause pressure.

## Dependencies

None

## Impact

None


rtodling:
@Dooruk Sorry to be asking for these ... but can we work to get this one in as well?

Dooruk (approved):
**Note:** `3dvar` fill fail due to a known issue.

Dooruk:
@rtodling which issues would this close?

rtodling:
> @rtodling which issues would this close?

@Dooruk there isn't an issue associated w/ this ... this is simply fixing an error in the setup and configurations used in the UFO. The default in UFO is to use a version of the tropopause pressure calculation that is not what we presently use.

PR #465, opened 2024-11-23
Revise emis coeffs in light of troubled JEDI

### LINKED ISSUE #426: Something is amiss in IR radiance assimilation
As I introduce T-skin as a control variable, I am re-doing some of the exercise I've done before of looking at each of the main instruments increments and how they compare w/ GSI's. IASI (and CrIS-FSR) has always been in the suspicious list. I want to have a record of this here so we can work together to address whatever issue affects hyperspectral IR.

The temperature increment for microwave instruments such as ATMS (clear sky) and GMI (all sky) are quite consistent between GSI and JEDI. The corresponding T-skin increments are also quite consistent - with still a puzzle over Antartica (but not a show stopper - see e.g., https://github.com/JCSDA-internal/fv3-jedi/pull/1257.

For IASI, the story is quite alarmingly different. For starts, GSI does not fully converge its minimization when given up to 100 iteration to crank; but so be it. JEDI, on the other hand, seems to converge in about 40 something iterations, which is suspicious in itself. But the real issue is how different the increments from both systems look (perhaps not a surprise given the convergence differences - chicken and egg).

Interestingly enough, the increments on specific humidity between the two systems are quite reasonable - here is an example at 850 hPa - GSI (top) JEDI (bottom)

q_inc_due_to_iasib

Unfortunately, this is what temperature looks like at the same level:

t850_inc_due_to_iasib

notice T increments are multiplied by a factor of 2 for plotting enhancement purposes; still -diff in increments are larger than increments themselves.

Now here is what T-skin increments looks like:

ts_due_to_iasib

It looks as though the weight given to whatever induces the sensitivity of Tb to T-skin is overweighted.

rtodling:
## Description

After considerable looking I find that a temporary solution to the issue associated with the very large increments of T-skin obtained from IR instruments in JEDI can be alleviated by rescaling the emissivity coeffs in the surface Jacobian filter as indicated here.

I have long discussions in the SWELL issue:

- issue #426 

and also - even more extensive here: https://github.com/orgs/JCSDA-internal/discussions/156

The palliative being put forward here meant to allow GMAO to run with a JEDI configuration with known issues minimized as much as possible. 

It is still necessary to look more carefully at this and figure try to figure out why JEDI gets  such large IR emissivities to begin with.   

## Dependencies

None

## Impact

Much more reasonable results.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/airs_aqua.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/airs_aqua.yaml
@@ -241,7 +241,9 @@ obs post filters:
       options:
         channels: *airs_aqua_channels
         sensor: *Sensor_ID
-        obserr_demisf: [0.01, 0.02, 0.03, 0.02, 0.03]
+#       obserr_demisf: [0.01, 0.02, 0.03, 0.02, 0.03]
+#   Todling: return to compensate for unknown IR emis issue
+        obserr_demisf: [0.20, 0.40, 0.45, 0.40, 0.45]
         obserr_dtempf: [0.50, 2.00, 4.00, 2.00, 4.00]
 #  Gross check
 - filter: Background Check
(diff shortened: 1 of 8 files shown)


rtodling:
I think this is an issue that relates to other problems - simply revising the emiss won't do it.

PR #474, opened 2024-11-25
MW: Revised coeffs scalling emissivity term in surf Jacobian

rtodling:
## Description

Though much less that for IR, the emissivity terms for the surface Jacobian for AMSUA and ATMS also seem to benefit from a revision -  all, expect mixed surfaces are multiplied by a factor of 5.

here is an illustration of 2 JEDI swell configurations (default; top)) and the one proposed here (bot) compared agains GSI - the figs are for Tskin increment differences w/ GSI.

 
ts-dueAllMW-diff-gsiX2jedi

## Dependencies

None

## Impact

None

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_aqua.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_aqua.yaml
@@ -147,7 +147,7 @@ obs post filters:
           sensor: *Sensor_ID
           use_biasterm: true
           test_biasterm: ObsBiasTerm
-          obserr_demisf: [0.010, 0.020, 0.015, 0.020, 0.200]
+          obserr_demisf: [0.050, 0.100, 0.075, 0.100, 0.200]
           obserr_dtempf: [0.500, 2.000, 1.000, 2.000, 4.500]
 # Gross check
   - filter: Background Check
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_metop-b.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_metop-b.yaml
@@ -147,7 +147,7 @@ obs post filters:
           sensor: *Sensor_ID
           use_biasterm: true
           test_biasterm: ObsBiasTerm
-          obserr_demisf: [0.010, 0.020, 0.015, 0.020, 0.200]
+          obserr_demisf: [0.050, 0.100, 0.075, 0.100, 0.200]
           obserr_dtempf: [0.500, 2.000, 1.000, 2.000, 4.500]
 # Gross check
   - filter: Background Check
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_metop-c.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_metop-c.yaml
@@ -147,7 +147,7 @@ obs post filters:
           sensor: *Sensor_ID
           use_biasterm: true
           test_biasterm: ObsBiasTerm
-          obserr_demisf: [0.010, 0.020, 0.015, 0.020, 0.200]
+          obserr_demisf: [0.050, 0.100, 0.075, 0.100, 0.200]
           obserr_dtempf: [0.500, 2.000, 1.000, 2.000, 4.500]
 # Gross check
   - filter: Background Check
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_n15.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_n15.yaml
@@ -147,7 +147,7 @@ obs post filters:
           sensor: *Sensor_ID
           use_biasterm: true
           test_biasterm: ObsBiasTerm
-          obserr_demisf: [0.010, 0.020, 0.015, 0.020, 0.200]
+          obserr_demisf: [0.050, 0.100, 0.075, 0.100, 0.200]
           obserr_dtempf: [0.500, 2.000, 1.000, 2.000, 4.500]
 # Gross check
   - filter: Background Check
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_n18.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsua_n18.yaml
@@ -147,7 +147,7 @@ obs post filters:
           sensor: *Sensor_ID
           use_biasterm: true
           test_biasterm: ObsBiasTerm
-          obserr_demisf: [0.010, 0.020, 0.015, 0.020, 0.200]
+          obserr_demisf: [0.050, 0.100, 0.075, 0.100, 0.200]
           obserr_dtempf: [0.500, 2.000, 1.000, 2.000, 4.500]
 # Gross check
   - filter: Background Check
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/atms_n20.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/atms_n20.yaml
@@ -149,7 +149,7 @@ obs post filters:
           use_biasterm: true
           test_biasterm: ObsBiasTerm
           sensor: *Sensor_ID
-          obserr_demisf: [0.010, 0.020, 0.015, 0.020, 0.200]
+          obserr_demisf: [0.050, 0.100, 0.075, 0.100, 0.200]
           obserr_dtempf: [0.500, 2.000, 1.000, 2.000, 4.500]
 # Gross check
   - filter: Background Check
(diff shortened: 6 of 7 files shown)


gmao-wgu:
Are the large differences in the Ts and emissivity sensitivity between GSI and JEDI due to background interpolation differences(GeoVals) or potential bugs in the CRTM interface in JEDI?

gmao-yzhu:
We are not using surface-sensitive AMSU-A and ATMS radiances. I'm not sure why Ts increment is sensitive to their emissivity jacobians? I probably missed something here? 

@gmao-yzhu You are missing that I am matching what GSI is doing ... so whatever question is this respect must be answered in the GSI sense.

mjkagnes123:
> We are not using surface-sensitive AMSU-A and ATMS radiances. I'm not sure why Ts increment is sensitive to their emissivity jacobians? I probably missed something here?
> 
> @gmao-yzhu You are missing that I am matching what GSI is doing ... so whatever question is this respect must be answered in the GSI sense.

Observations of some AMSUA channels (4 and 5)  and ATMS oxygen channels ( 5, 6, 17, 18), located at the outer edges of the absorption band, do have  sensitivity to surface. 
Picture1

mjkagnes123:
I looked that Ts increment from GSI (x0048) assimilated only each MW observation type.The majority of the Ts increments generated by microwave sensors come from AMSUA-A and ATMS, with only a small contribution from MHS. (double checked and confirmed that no Ts increments from GMI, AMSR2, and SSMIS)

x0048_mhs_ts_inc
x0048_atms_ts_inc
x0048_amsua_ts_inc
me from AMSUA-A and ATMS, with only a small contribution from MHS.

Here is the Ts increment from x0050 modified to assimilate only MW radiance observations.
x0050_allmw_ts_inc

mjkagnes123:
My understanding is that increasing the values in 'obserr_demisf' in this PR will penalize the observations more, particularly when the first-guess Tb sensitivity to emissivity in CRTM is significant. I am totally okay with this, and I think that's how it should have been even in GSI. However, what still puzzles me is why the Ts increment in the original JEDI run (shown in Ricardo's first plot) was that high, and what could cause that huge difference shown in the 1st plot.

gmao-wgu:
This change will cause our Tier 1 and Tier 2 tests to fail unless we intentionally ignore them.

mjkagnes123 (approved):
My understanding is that increasing the values in 'obserr_demisf' in this PR will penalize the observations more, particularly when the first-guess Tb sensitivity to emissivity in CRTM is significant. I am totally okay with this, and I think that's how it should have been even in GSI.  (i.e. "Green" light for this review and this PR.).   

However, what still puzzles me is why the Ts increment in the original JEDI run (shown in Ricardo's first plot) was that high, and what could cause that huge difference shown in the 1st plot.

gmao-yzhu:
I agree with both what Ming-Jeong and Wei said. On one hand, to make Ts increments comparable between GSI and JEDI, increasing obs error inflation would help to acheive that; on the other hand, it will make our UFO tests fail, and we haven't done any tests to verify the impact of doing such. You both mentioned one question -- why they are so different given that the same CRTM is used. Could you compare the background used in the CRTM Tb calculation?

rtodling:
All - it's okay that you don't agree w/ my proposed changes. But I also don't think changing observations errors would be justifiable either - simply for the same reasons: JEDI should take the configuration of GSI. 

Now clearly, JEDI has a problem handling the emissivities. So, I need to put the work of figuring out why is it that JEDI handles emissivities so differently than GSI, on you.

I hope you can and will take the time to figure this out. I, for my part, will no longer look at this - but I will also not cycle w/ the values that we have since I will be wasting my time.

rtodling:
> I looked that Ts increment from GSI (x0048) assimilated only each MW observation type.The majority of the Ts increments generated by microwave sensors come from AMSUA-A and ATMS, with only a small contribution from MHS. (double checked and confirmed that no Ts increments from GMI, AMSR2, and SSMIS)
> 
> x0048_mhs_ts_inc x0048_atms_ts_inc x0048_amsua_ts_inc me from AMSUA-A and ATMS, with only a small contribution from MHS.
> 
> Here is the Ts increment from x0050 modified to assimilate only MW radiance observations. x0050_allmw_ts_inc

@mjkagnes123 Perhaps it is not clear to you: but all that you are showing here, I have already produce on my own. I exactly what the Ts increment looks like for every single combination of the radiance observing system. Thank you though.

rtodling:
> Are the large differences in the Ts and emissivity sensitivity between GSI and JEDI due to background interpolation differences(GeoVals) or potential bugs in the CRTM interface in JEDI?

@gmao-wgu why would the differences you refer to produce a factor of 5 or 10 difference in increments when the difference in temperature is of a few percents? Also, why the differences you refer to would produce different magnitude of Tskin increments for different instruments (MW vs IR)?  

These are indeed rhetorical question, I am sorry to say, I don't think B has anything to do w/ this. But if you can prove me wrong ... all good.

gmao-wgu:
@rtodling  I was confused here. Looking back, I don’t recall seeing any plots directly showing significant differences in Ts and surface emissivity sensitivity, which are used to inflate observation errors in both MW and IR, between GSI and JEDI. Perhaps I missed something.

gmao-wgu:
Secondly, when you increase those coefficients by a factor of 5, it does not necessary translate to the observation errors being 5 times larger.

gmao-wgu:
We know that the background errors used in JEDI are larger compared to those in GSI. If you reduce the background errors slightly, you might achieve similar results to those obtained with the inflated coefficients by a factor of 5.

rtodling:
> We know that the background errors used in JEDI are larger compared to those in GSI. If you reduce the background errors slightly, you might achieve similar results to those obtained with the inflated coefficients by a factor of 5."

The same way you guys are not willing to make the changes here, I am not wiling to make the change in the background errors - the whole reason I introduced GSIBEC w/ the help of Dan was to avoid tuning the background errors.  Sorry.

gmao-wgu:
@rtodling It is entirely your call. I support whatever decision you make to make it work.

gmao-wgu (approved):

gmao-yzhu:
I see both Wei and Min-Jeong approved this PR. I’m fine with this, which might be the easiest way to adjust, we can take a closer look later once we have the system ready. We will need to remember that after this change we should not expect UFO tests for ATMS and AMSUA to pass.

From: Wei Gu ***@***.***>
Date: Monday, December 16, 2024 at 7:52 PM
To: GEOS-ESM/swell ***@***.***>
Cc: Zhu, Yanqiu (GSFC-6101) ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] [BULK] Re: [GEOS-ESM/swell] MW: Revised coeffs scalling emissivity term in surf Jacobian (PR #474)
CAUTION: This email originated from outside of NASA.  Please take care when clicking links or opening attachments.  Use the "Report Message" button to report suspicious messages to the NASA SOC.



@gmao-wgu approved this pull request.

rtodling:
Ok, as we talked about in the meeting this AM, I looked at the UFO ... I am looking at channel 5 from ATMS-N20. And this is what I am finding (an example), the swell code/settings in develop produce the following scatter for Final Obs Errors JEDI vs GSI:

orig

The following is unfortunately what I get with the proposed changes in this PR:

prop

The above seem clearly to be against the change in ATMS (I see the same in the ATMS on NPP).

But oddly enough, I don't see this in AMSUA-N19 channel 5 and 4, but I do see a little off-centered scatter for AMSUA-N18, chan 4.

So, given these ... I am going to close this issue ... and look for a solution to the increment results elsewhere.

gmao-yzhu:
Ricardo, Thanks very much for doing tests.


From: rtodling ***@***.***>
Date: Wednesday, December 18, 2024 at 1:33 PM
To: GEOS-ESM/swell ***@***.***>
Cc: Zhu, Yanqiu (GSFC-6101) ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] [BULK] Re: [GEOS-ESM/swell] MW: Revised coeffs scalling emissivity term in surf Jacobian (PR #474)
CAUTION: This email originated from outside of NASA.  Please take care when clicking links or opening attachments.  Use the "Report Message" button to report suspicious messages to the NASA SOC.



Ok, as we talked about in the meeting this AM, I looked at the UFO ... I am looking at channel 5 from ATMS-N20. And this is what I am finding (an example), the swell code/settings in develop produce the following scatter for Final Obs Errors JEDI vs GSI:
orig.png (view on web)<https://github.com/user-attachments/assets/b951ba50-f31f-4201-a137-e66e5ac0b8f7>

The following is unfortunately what I get with the proposed changes in this PR:
prop.png (view on web)<https://github.com/user-attachments/assets/c3033172-af0b-4b1b-bef1-d3f9a5425d26>

The above seem clearly to be against the change in ATMS (I see the same in the ATMS on NPP).

But oddly enough, I don't see this in AMSUA-N19 channel 5 and 4, but I do see a little off-centered scatter for AMSUA-N18, chan 4.

So, given these ... I am going to close this issue ... and look for a solution to the increment results elsewhere.

PR #477, opened 2024-12-04
Revise obs space localization for LETKF

rtodling:
## Description

The original lenghscales were generating a completely noisy, unphysical, increment. 

I don't claim scientific correctness ... just reasonableness. Here is the T-500 increment original and after the revised scales:

letkf_loc500

Note these results use 32 members, at c90 resolution. I will suggest we change to more than 2 members for testing purposes in swell - in another PR - I don't think 2 members makes any sense.

## Dependencies

None

## Impact

None

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ setuptools.setup(
                'configuration/*/*/*',
                'configuration/*/*/*/*',
                'configuration/*/*/*/*/*',
+               'configuration/*/*/*/*/*/*',
                'utilities/pinned_versions/*.yaml'
              ],
     },
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/localization/aircraft.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/localization/aircraft.yaml
@@ -1,4 +1,4 @@
 obs localizations:
   localization method: Horizontal Gaspari-Cohn
-  lengthscale: 1000.e3
+  lengthscale: 500.e3
   max nobs: 100
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/localization/airs_aqua.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/localization/airs_aqua.yaml
@@ -1,4 +1,4 @@
 obs localizations:
   localization method: Horizontal Gaspari-Cohn
-  lengthscale: 1000.e3
+  lengthscale: 500.e3
   max nobs: 100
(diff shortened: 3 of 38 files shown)


Dooruk (dismissed):
Looks good to me, I will wait for @metdyn's approval

rtodling:
@Dooruk  let me make one more change to this ...

rtodling:
I add all obs systems to:

test/suite_tests/localensembleda-tier1.yaml

but I cannot understand why the job fails when trying to add the obs localization:

/gpfsm/dnb05/projects/p139/rtodling/cylc-run/swell-getkf-all-suite/run1/log/job/20211212T0000Z/RunJediLocalEnsembleDaExecutable-geos_atmosphere/01/job.err

rtodling:
It seems the localization directory is not being installed ...

rtodling:
I think I got the localization directory installed now ... changed the setup.py

Dooruk:
~You need to edit https://github.com/GEOS-ESM/swell/blob/develop/setup.py~
Beat me to it

rtodling:
32 members seem to be too much for the default PE choice ... trying w/ 10 members

rtodling:
@Dooruk @metdyn This works - not ideal cause only a few instruments and very few members - but it runs ... the problem being that it we need to add something in the yaml handling logic for the obs that tells the code to bypass the writing of HofXs - for now: that is, for example, for aircraft, we need to bypass:

      obsdataout:
        engine:
          obsfile: /discover/nobackup/rtodling/SwellExperiments/SLES15/swell-getkf-all/run/20211212T000000Z/geos_atmosphere/swell-getkf-all.aircraft.20211211T210000Z.nc4

this means that we'll need to have if statements around each observation file projecting this for "if not ensemble" then write files out ... 


also, I tried changing the defaults:

  save posterior ensemble: true
  save posterior ensemble increments: true
  save posterior mean: false
  save posterior mean increment: true


to simply:

  save posterior ensemble: false
  save posterior ensemble increments: false
  save posterior mean: false
  save posterior mean increment: true 

but I had troubles w/ Swell not liking what I did ... perhaps @metdyn you can look at these issues ... and add to what here in this branch?

metdyn:
Thank you and I will look at this.
That the yaml control did not work is really inconvenient and I will check
that.

setup.py
```diff
@@ -45,6 +45,7 @@
                'configuration/*/*/*',
                'configuration/*/*/*/*',
                'configuration/*/*/*/*/*',
+               'configuration/*/*/*/*/*/*',
```
metdyn, line 48: @rtodling @Dooruk ,
If you donot mind, I will delete this line.
Without it, I already get to the step jedi_localensembleda_log.log in cylc. 
``` 

LinearModel<MODEL>::forecastTL: Starting
----------------------------------------------------------------------------------------------------
Increment print | number of fields = 8 | cube sphere face size: C90
```

Dooruk: @metdyn I believe @rtodling included that for `Localizations` folder to work. If you are %100 sure that it's not needed I don't care.

metdyn: Thank you, @rtodling, for the great work!   We should merge this long due PR, @Dooruk! (No, this line does not bother me either)

metdyn (dismissed):
Thanks a lot for the great work!

Dooruk (approved):

PR #478, opened 2024-12-04
replace conversion and use of aircraft bias CSV files with convert/use of nc4 files to use with updated bias correction

gmao-msienkie:
### Description

Several scripts are updated to remove reference to the CSV bias files and enable generation and use of the new netCDF4 format files.

### Other important details

- Updating of SWELL atmosphere default experiments to x0050 from x0048, hence the default cycle times change to 2023/10/10 0z and 6z
- `aircraft` observer is now handled as two separate observers `aircraft_temperature` and `aircraft_wind` 
- `ufo_testing` fails but we will abandon using it for tier 1 & 2 tests https://github.com/GEOS-ESM/CI-workflows/issues/14

FYI @mranst


gmao-msienkie:
My code has 'ascent rate' where the develop had had the ridiculous 'windUpward' before.  I guess 'instantaneousAltitudeRate' is OK too.

gmao-msienkie:
current status:
- The updated src/swell/tasks/gsi_bc_to_ioda.py seems to be working okay 
- For src/swell/tasks/get_observations.py I gave the aircraft coefficient file the type 'satbias' which mean I also had to provide a fake satbias_pc file.  I didn't think I wanted to stick with 'csv'.  Maybe at some point the r2d2 could use the newer name "obsbias_coefficients" in r2d2.  I have some sample files stored in my R2D2DataStore on nobackup
- The RunJediHofxExecutable task fails. "Bad parameter: getAllIndices: Can't find element in the vector"  The bias correction configuration was copied from ufo/test/testinput/instrumentTests/Aircraft_obs/aircraft_geos_qc.yaml

gmao-msienkie:
current status:
- The updated src/swell/tasks/gsi_bc_to_ioda.py seems to be working okay 
- For src/swell/tasks/get_observations.py I gave the aircraft coefficient file the type 'satbias' which mean I also had to provide a fake satbias_pc file.  I didn't think I wanted to stick with 'csv'.  Maybe at some point the r2d2 could use the newer name "obsbias_coefficients".  I have some sample files stored in my R2D2DataStore on nobackup
- The RunJediHofxExecutable task fails. "Bad parameter: getAllIndices: Can't find element in the vector"  The bias correction configuration was copied from ufo/test/testinput/instrumentTests/Aircraft_obs/aircraft_geos_qc.yaml

gmao-msienkie:
I modified the title to try to make clear that the CSV files are not being converted, rather that the original text GSI bias coefficient files will be converted to NC4 format files.
Any existing CSV bias coefficient files can be replaced with NC4 files converted directly from the GSI bias coefficient files.

rtodling:
@gmao-msienkie I just pulled your branch and tried setting a 3dvar case ... the GetObservation task fails due to not finding the new bias correction (nc4) file ... I assume you are pointing to a location of yours that has such files - that is, you probably have an r2d2_config.yaml that allows for the proper testing. Can you point me to that, please?

src/swell/tasks/get_observations.py
```diff
@@ -192,31 +192,46 @@ def execute(self) -> None:
...
+#                   fetch(date=background_time,
+#                         target_file=target_file,
+#                         provider='gsi',
+#                         obs_type=target_file_type,
+#                         type='bc',
+#                         experiment=obs_experiment,
+#                         file_type='csv')
+#
+#                   # Change permission
+#                   os.chmod(target_file, 0o644)
```
Dooruk:
```suggestion
```

src/swell/tasks/get_observations.py
```diff
@@ -192,31 +192,46 @@ def execute(self) -> None:
...
-
-                    self.logger.info(f'Processing aircraft bias file {target_file}')
+                self.logger.info(f'Processing aircraft bias file {target_abccoef}')
+                fetch(date=background_time,
+                      target_file=target_abccoef,
+                      provider='gsi',
+                      obs_type=observation,
+                      type='bc',
+                      experiment=obs_experiment,
+                      file_type='satbias')
```
Dooruk:
```suggestion
                      file_type='acftbias')
```

Dooruk: I will create this extension in R2D2, this is still an nc4 file but R2D2 needs this extension..

src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/aircraft.yaml
```diff
@@ -14,6 +14,29 @@ obs space:
       obsfile: '{{cycle_dir}}/{{experiment_id}}.aircraft.{{window_begin}}.nc4'
   simulated variables: [airTemperature, windEastward, windNorthward]
 
+obs bias:
+  input file: '{{cycle_dir}}/aircraft_abias_air.{{window_begin}}.nc4'
```
Dooruk:
```suggestion
  input file: '{{cycle_dir}}/aircraft_abias_air.{{window_begin}}.acftbias'
```

Dooruk (changes requested):
I tested running ncdiag, because of the way filenames changed we need to do a few changes but other than that it should be good to go.

And this needs to be merged with develop. @gmao-msienkie I can handle those changes.

Dooruk:
Ok, this works for x0050 only.  However only for 2023/10/10 and backgrounds has to come outside of R2D2. @rtodling please give it a shot and see if c180 background works or if we need to adjust that as well.

Dooruk:
thank you  @gmao-msienkie for adding this.

Dooruk (dismissed):

rtodling:
I just pushed the split of aircraft into temperature and wind individual ob-spaces - the bias correction inflation filter for temperature is not working - it needs an update to be consistent with latest JEDI.

EVA is still not working so, the 3dvar suite fails due to that ... looking into it.

rtodling:
There is other stuff missing - like in the UFO testing and the corresponding split in yaml info

Dooruk:
@rtodling, I don't see the `aircraft_wind` or `aircraft_temperature` yamls in the observation folder, just that you erased the `aircraft.yaml`.

Dooruk:
You need to update `src/swell/configuration/jedi/observation_ioda_names.yaml` as well

Dooruk:
When did this JEDI change go in? What version (time-wise) of JEDI do we need Aircraft to work? @gmao-msienkie

rtodling (approved):
As far as I know everything is good to go.

Issue #427, opened 2024-09-17
(SE) Ability to run Tier 2 tests locally


ashiklom:
The ability to trigger tier 1 tests via command line (e.g., `swell t1test ufo_testing`) is really convenient. We should create a similar capability for tier 2 (possibly, with finer-grained control, because the tests are bigger and will take longer to run).

The current Tier 1 _local_ tests are defined here: https://github.com/GEOS-ESM/swell/blob/develop/src/swell/test/suite_tests/suite_tests.py

A similar structure could probably be used for Tier 2 tests.

The shell version of the Tier 2 workflows are defined here:

https://github.com/GEOS-ESM/CI-workflows/blob/main/.github/workflows/swell-tier2_application_discover.yml

Dooruk:
Do these tests automatically use the default `platform`? I used `t1test` on SLES12 yesterday and received a lot of errors for different tasks. I suspect it might have been `platform` related?

mranst:
> Do these tests automatically use the default `platform`? I used `t1test` on SLES12 yesterday and received a lot of errors for different tasks. I suspect it might have been `platform` related?

They do use the default platform, which is set at sles15. I think I might have set that when changing the default platform. I think I changed all references to sles12 to sles15. The option for platform should be probably be propagated down to the suite tests like it is for other swell commands

https://github.com/GEOS-ESM/swell/blob/7812c4114de3c587a026cc8c7453ad3e6b7e4528/src/swell/test/suite_tests/suite_tests.py#L78-L81

ashiklom:
Closed via #481

PR #481, opened 2024-12-11
Ability to run tier 2 tests locally

mranst:
#427 

This PR adds support for running tier2 tests from the command line, using an extension of the existing `run_suite` method, as well as the ability to specify the platform to run tier1 and tier2 tests on using the `-p` flag.

Tier 2 tests are run similarly to tier 1 tests:
`swell t2test <suite> -p <platform>` 
Where platform will default to `nccs_discover_sles15` if unspecified. Overrides are used from `~/.swell/swell-test.yaml`. Tier 2 tests will start by building  JEDI unless an existing JEDI directory is specified in ~/.swell/swell-test.yaml, using the lines: 

`jedi_build_method: use_existing`
`existing_jedi_source_directory: /path/to/source/
`existing_jedi_build_directory: /path/to/build/`

The JEDI build experiment directory is placed alongside the experiment directory for the suite being run and begins with the same time code

I’ve tested building JEDI and running suites on sles12 and sles15 and both work


src/swell/utilities/shell_commands.py
```diff
@@ -57,6 +57,7 @@ def run_track_log_subprocess(
     rc = process.poll()
     if rc != 0:
         command_string = ' '.join(command)
+        print(command_string)
```
Dooruk: Is this print a leftover?

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -11,13 +12,55 @@
 from swell.utilities.dictionary import update_dict
 
 
-def run_suite(suite: str):
+def build_jedi_for_tier2(test_dir: str, experiment_id_root: str, platform: str, test_config: dict):
+    suite_overrides_file = (resources.files("swell") /
+                            "test" /
+                            "suite_tests" /
+                            "build_jedi-tier1.yaml")
```
Dooruk, line 24: Do all bundles get built by default? Is that why you are introducing this file?

Dooruk (dismissed):
I was able to follow instructions and run t2test, thanks @mranst.

This is beyond the scope of this PR but our intention eventually is to t1tests to run `*-tier1.yaml` and t2tests to run `*-tier2.yaml` correct? Main difference being the resolution change, so essentially `tier2.yaml`s would only contain few YAML keys different than `tier1.yaml` (resolution, npx_proc, npy_proc). You can see the changes in my severely outdated branch [here](https://github.com/GEOS-ESM/swell/compare/develop...feature/tier1_c13).

src/swell/utilities/shell_commands.py
mranst: Yep, sorry

src/swell/test/suite_tests/suite_tests.py
mranst: I thought they were when run in the application tests, do we not need to run them locally? It certainly takes a long time

https://github.com/GEOS-ESM/CI-workflows/blob/fdb66bd3203e525044804b9eef4705c02e1963fa/.github/workflows/swell-tier2_application_discover.yml#L42-L81

mranst:
> I was able to follow instructions and run t2test, thanks @mranst.
> 
> This is beyond the scope of this PR but our intention eventually is to t1tests to run `*-tier1.yaml` and t2tests to run `*-tier2.yaml` correct? Main difference being the resolution change, so essentially `tier2.yaml`s would only contain few YAML keys different than `tier1.yaml` (resolution, npx_proc, npy_proc). You can see the changes in my severely outdated branch [here](https://github.com/GEOS-ESM/swell/compare/develop...feature/tier1_c13).

I was wondering how to go about this. In the latest commit I added the capability to user overrides from tier2 files if they are present, that way people can add them and test them in the future. I tested it with the file you linked

src/swell/test/suite_tests/suite_tests.py
Dooruk: Well `fv3-jedi` and `soca` will be required and they have a big list of dependencies to pretty much all of the repos except two or three (`iodaconv` IODA converters and `geos-aero` aerosol. You can see the required repo dependencies here:

https://github.com/GEOS-ESM/jedi_bundle/blob/develop/src/jedi_bundle/config/bundles/fv3-jedi.yaml
https://github.com/GEOS-ESM/jedi_bundle/blob/develop/src/jedi_bundle/config/bundles/soca.yaml

So in this case `ufo` will be built but we could omit `iodaconv` but up to you. Moving forward as we introduce other models certain applications might require less repos and then we could create the bundle list dynamically depending on the type of suite/application.

Dooruk:
> I was wondering how to go about this. In the latest commit I added the capability to user overrides from tier2 files if they are present, that way people can add them and test them in the future. I tested it with the file you linked

Thank you, that accommodates my request. When people run `swell create` it defaults to using Tier1 YAMLs (which overrides defaults in `task_questions.yaml`, clearly not the best software design!), which are set to low resolution currently, at least for ocean. Perhaps we could also switch `swell create` to default to Tier2 YAMLs which would contain more useful resolutions for users to test with and still keep Tier1 YAML as low-res.

Swell PRs would trigger Tier 1 tests, with low resolution, compares results against test outputs.
Nightlies trigger Tier 2 with higher resolution, compares results against previous night's JEDI build.
Tier 3 in the long term with GEOSgcm changes etc.

src/swell/test/suite_tests/suite_tests.py
ashiklom: I defer to @Dooruk 's expertise on which JEDI stuff needs to be built.

But if we do need to specify JEDI stuff that needs to be built, I would recommend against the "YAML file as constant" approach in favor of just embedding this information in a Python dict directly in here:

```python
suite_overrides = {
    "jedi_build_method": "create",
    "bundles": ["fv3-jedi", "soca", "iodaconv", "ufo"]
}
```

_Eventually_, I want to start capturing this kind of information in dataclasses that can be validated (e.g., below), but that requires a lot of other engineering changes.

```python
# In a file like `jedi_bundles.py`...

from dataclasses import dataclass
from enum import Enum

class JediBuildMethod(Enum):
    CREATE = "create"
    ... # Other build methods...

class JediBundle(Enum):
    FV3_JEDI = "fv3-jedi"
    SOCA = "soca"
    IODACONV = "iodaconv"
    UFO = "ufo"
    ... # More...

@dataclass
class JediBuild:
    jedi_build_method: JediBuildMethod
    bundles: list[JediBundle]

# Here...

from jedi_bundles import JediBuild, JediBuildMethod, JediBundle

suite_overrides = JediBuild(
    jedi_build_method = JediBuildMethod.CREATE,
    bundles = [JediBundle.FV3_JEDI, JediBundle.SOCA, JediBundle.IODACONV, JediBundle.UFO]
)
```

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -11,13 +12,55 @@
...
+
+    suite_path = str(experiment_dir / f"{experiment_id}-suite")
+    log_path = str(experiment_dir / "log")
+
+    launch_experiment(suite_path, True, log_path)
+
+    return experiment_dir
+
+
+def run_suite(suite: str, platform: str, test_tier: int = 1):
```
ashiklom: `test_tier` is an excellent place to practice using Python's built-in `enums`. We want to have some flexibility later to revise our test naming (e.g., we might have options like `tier1-ci` or `tier2-weekly`; this locks us into an integer-based test naming scheme.

I suggest:

```python
# Near the top of the file
from enum import Enum, auto

class TestSuite(Enum):
    TIER1 = auto()
    TIER2 = auto()

# ...
# Then, here:
def run_suite(suite: str, platform: str, test_tier: TestSuite):
    ...

# ...

# This will work fine:
run_suite("hofx", "nccs_discover_sles15", TestSuite.TIER1)

# But the type checker will complain about this:
run_suite("something", "some_other", "tier1")  # This will run, but the type checker will complain
```

src/swell/utilities/shell_commands.py
```diff
@@ -57,6 +57,7 @@ def run_track_log_subprocess(
     rc = process.poll()
     if rc != 0:
         command_string = ' '.join(command)
+
```
ashiklom: Revert this change to avoid unnecessarily contaminating the `git blame` on this file.

```suggestion
```

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -11,13 +12,55 @@
...
+    if "override" in test_config:
+        override = update_dict(override, test_config['override'])
+
+    experiment_dir = test_dir / experiment_id
+    experiment_dir.mkdir(parents=True, exist_ok=True)
+    override_yml = experiment_dir / "override.yaml"
+
+    with open(override_yml, "w") as f:
+        yaml.dump(override, f)
+    print(platform)
```
ashiklom: Stray debug by print?
```suggestion
```

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -71,12 +130,35 @@ def run_suite(suite: str):
...
+    # Build JEDI for tier 2 tests if existing build is not specified in user yaml
+    build_jedi_for_test = False
+    if test_tier == 2:
+        if ("jedi_build_method" in test_config
+           and test_config["jedi_build_method"] == "use_existing"
+           and 'existing_jedi_source_directory' in test_config
+           and 'existing_jedi_build_directory' in test_config):
+            build_jedi_for_test = False
+        else:
+            build_jedi_for_test = True
```
ashiklom: `build_jedi_for_test` is set redundantly here. Simplify the logic. Maybe:

```python
if test_tier == 2:
  build_jedi_for_test = True
  if ("jedi_build_method" in test_config
    and "existing_jedi_source_directory" in test_config
    ...):
    build_jedi_for_test = False
    if build_jedi_for test:
      jedi_dir = ...
      # etc...
```

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -71,12 +130,35 @@ def run_suite(suite: str):
...
+        jedi_dir = build_jedi_for_tier2(testdir, experiment_id_root, platform, test_config)
+
+        tier2_override = {"jedi_build_method": "use_existing",
+                          "existing_jedi_source_directory": f"{jedi_dir}/jedi_bundle/source",
+                          "existing_jedi_build_directory": f"{jedi_dir}/jedi_bundle/build"}
+
+        override = update_dict(override, tier2_override)
+
+        if suite == "build_jedi":
+            sys.exit()
```
ashiklom: Strongly advise against using `sys.exit()` here (or anywhere) because that will kill the parent python process. Much better to early `return`. If you need to make sure you don't do any other stuff outside of this function, then `return None` or some explicit object and check for that in the calling code.

src/swell/swell.py
```diff
@@ -244,15 +244,40 @@ def test(test: str) -> None:
 
 
 @swell_driver.command()
+@click.option('-p', '--platform', 'platform', default="nccs_discover_sles15", help=platform_help)
```
ashiklom: We have a fixed number of options for this, right? If so, this should probably be `type=click.Choice(("nccs_discover_sles15", ...))`.

Also, since this is common information, we should track it somewhere; i.e., just as we have a `valid_tests` variable, we should also have a `valid_platforms` variable or similar.

src/swell/swell.py
```diff
@@ -244,15 +244,40 @@ def test(test: str) -> None:
...
     """
-    run_suite(suite)
+    run_suite(suite, platform, 1)
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+@swell_driver.command()
+@click.option('-p', '--platform', 'platform', default="nccs_discover_sles15", help=platform_help)
```
ashiklom: Ditto to above: `type=click.Choice(...)`.

src/swell/swell.py
ashiklom: Again, maybe not strictly for this PR (depending on exactly how much work it would be to implement), but down the line: Python's enums provide an elegant solution here. If we define a fixed number of platforms in one place as a `Platform` enum, then we can pass it into `click.Choice` _and_ we pick up any misuse elsewhere in the code via the type checker.

```python
# Defined elsewhere...
from enum import Enum

class Platform(Enum):
    NCCS_DISCOVER_SLES12 = "nccs_discover_sles12"
    NCCS_DISCOVER_SLES15 = "nccs_discover_sles15"

# ...and now, here:
from _wherever import Platform
# ...
@click.option("-p", "--platform", "platform", type=click.Choice([p.value for p in Platform]))
```

src/swell/test/suite_tests/suite_tests.py
mranst: Oh, I didn't even realize introducing that file would override the `task_questions.yaml`. I see `iodaconv` listed as default under `task_questions.yaml`, but we can omit it from the tier tests? I don't see why not if we don't need it

Dooruk: Yeah, on second though, `iodaconv` is used by `convert_ncdiags` and will be used by `convert_bufr` that @cohen-seth is working on. It is a lightweight repo compared to others so we could keep it.

However, I think we should make some repos suite specific, for instance there is a COMPO package in JEDI that should only be build when GEOS-CF related suites are involved. Not straightforward and beyond the scope of this PR though, just thinking out loud.

mranst:
> > I was wondering how to go about this. In the latest commit I added the capability to user overrides from tier2 files if they are present, that way people can add them and test them in the future. I tested it with the file you linked
> 
> Thank you, that accommodates my request. When people run `swell create` it defaults to using Tier1 YAMLs (which overrides defaults in `task_questions.yaml`, clearly not the best software design!), which are set to low resolution currently, at least for ocean. Perhaps we could also switch `swell create` to default to Tier2 YAMLs which would contain more useful resolutions for users to test with and still keep Tier1 YAML as low-res.
> 
> Swell PRs would trigger Tier 1 tests, with low resolution, compares results against test outputs. Nightlies trigger Tier 2 with higher resolution, compares results against previous night's JEDI build. Tier 3 in the long term with GEOSgcm changes etc.

Since it was a very simple change to make, I made it so that swell create will use the tier2 overrides if available in the latest commit. If you think it's too early to introduce that I can roll it back

https://github.com/mranst/swell/blob/8f0c4148ab86a7a2da67da34402b908d24b8c1e0/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L302-L308

Dooruk:
> 
> Since it was a very simple change to make, I made it so that swell create will use the tier2 overrides if available in the latest commit. If you think it's too early to introduce that I can roll it back
> 
> https://github.com/mranst/swell/blob/8f0c4148ab86a7a2da67da34402b908d24b8c1e0/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L302-L308

Not at all, I don't want to hold this PR back, it's good to go in on my end and @ashiklom can suggest changes after he comes back from his time off.

Dooruk (dismissed):
all good, thanks @mranst  👍

ashiklom (changes requested):
Sorry --- I did this review before the holidays, but realized I forgot to hit "submit"!

I had a few suggestions, some easier to implement than others. I don't necessarily expect you to do 100% of what I requested here, so I defer to your judgment what you can bang out quickly for this PR vs. what we defer to a future issue.

src/swell/test/suite_tests/suite_tests.py
ashiklom: ...and if you really want to terminate execution of the entire program, `raise` an `Exception` of some kind; then, other code at least has the option of handling the exception if it chooses to.

src/swell/test/suite_tests/suite_tests.py
ashiklom: Note that one downside of this approach is that `TestSuite.TIER_1` is much more verbose. But, I think the added clarity and robustness (and integration with good tooling; e.g., auto-completion and static type-checking!) makes it worth it.

mranst: I like this method a lot more!

src/swell/test/suite_tests/suite_tests.py
mranst: The logic for this should be much simpler now

src/swell/test/suite_tests/suite_tests.py
mranst: Set to return None

src/swell/swell.py
mranst: I added `type=click.Choice(get_platforms())` like we do elsewhere in `swell.py`. That accounts for all platforms in the platforms directory, which seems like it might be simpler to develop adding new platforms, but I don't think it would be a challenge to incorporate it in this way

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -47,6 +95,22 @@ def run_suite(suite: str):
     with suite_overrides_file.open("r") as f:
         suite_overrides = yaml.safe_load(f)
 
+    # If it exists, update suite overrides from (suite)-tier2.yaml
+    if test_tier == 2:
```
ashiklom:
```suggestion
    if test_tier == TestSuite.TIER2:
```

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -5,19 +5,67 @@
...
 
 from swell.deployment.create_experiment import create_experiment_directory
 from swell.deployment.launch_experiment import launch_experiment
 from swell.utilities.dictionary import update_dict
 
 
-def run_suite(suite: str):
+class TestSuite(Enum):
+    TIER1 = auto()
+    TIER2 = auto()
```
ashiklom: I know `auto()` was in my original suggestion. But I think giving these literal string values may be a bit more powerful.

```suggestion
class TestSuite(Enum):
    TIER1 = "tier1"
    TIER2 = "tier2"
```

(Note: We can also consider the `StrEnum` for stuff like this, which offers [some ergonomic shorthands](https://tsak.dev/posts/python-enum/), but that locks us into at least Python 3.11 and, more importantly, accommodates more laziness in Enum usage than I think we want to allow). 

First, it lets you iterate over both names and values; for example:

```python
for test in TestSuite:
    print(f"{test.name} = {test.value}")
# TIER1 = tier1
# TIER2 = tier2
```

This can be exploited in list comprehensions for, e.g., `click.option` in code like this (not how we do it right now, but might be relevant at some future point):

```python
@click.option("-s", "--suite", "testsuite", type=click.Choice([t.value for t in TestSuite]))
```

Also (TIL!), the newly created `TestSuite` class constructor can be used for runtime data validation. This is especially useful for when we're parsing data from outside of Python (e.g., command line argument, environment variable, YAML file):

```python
TestSuite(TestSuite.TIER1)  # Can safely use the constructor just for data validation.
# <TestSuite.TIER1: 'tier1'>
TestSuite("tier2")
# <TestSuite.TIER2: 'tier2'>
TestSuite("tier3")
ValueError: 'tier3' is not a valid TestSuite
```

src/swell/test/suite_tests/suite_tests.py
ashiklom: My previous comment here got buried, so re-upping it for consideration: https://github.com/GEOS-ESM/swell/pull/481/files#r1892203782

I'm OK with leaving this as is since we're keeping other Tier test options in the `suite_tests` folder; that does provide some internal consistency. But, _in general_, I would really like us to start doing less storing constants in YAML moving forward --- YAML is a lot less flexible and powerful than Python, and adds an extra place developers have to look for settings outside the code.

src/swell/test/suite_tests/suite_tests.py
```diff
@@ -71,12 +135,29 @@ def run_suite(suite: str):
     experiment_dir = testdir / experiment_id
     experiment_dir.mkdir(parents=True, exist_ok=True)
 
+    # Build JEDI for tier 2 tests if existing build is not specified in user yaml
+    if test_tier == TestSuite.TIER2:
+        if ~("jedi_build_method" in test_config
```
ashiklom:
```suggestion
        if not ("jedi_build_method" in test_config
```

The tilde operator in Python is the invert operation, _not_ (necessarily) a negation. Try this:

```python
if ~True:
  print("yes")
```

Huh? What does `~True` return? And then, [study this](https://stackoverflow.com/questions/8305199/the-tilde-operator-in-python) 😉

mranst: Oops I forgot about this, thanks

ashiklom (approved):

PR #483, opened 2024-12-17
MW (AllSky): Missing Surf Jacobian check

rtodling:
## Description

I believe these the surface Jacobian check is missing from the All-sky MW obs yam files.

I believe this should be on for all MW - including those treated as all-sky ... when I do it for GMI I have no problems and indeed see the increment of Ts (without modified any of the coeffs being contested in another PR) seem to get closer to GSI's  ... here is a fig of increments of Tskin differences w/ the GSI increment when JEDI does no include the surface Jac term (top) and when it does (bottom).

 
gmi-sfcjac-incdiff

unfortunately, JEDI doesn't seem to like when this term is added to the MHS 

## Dependencies

None


rtodling:
I am closing this for now - but I am quite convinced there are some issues in the all-sky analysis.

PR #485, opened 2024-12-18
Keep the order of YAML configurations within run tasks

### LINKED ISSUE #409: Order of filters in yaml's
As most of you know, the order of filters in the observers matters. Certain actions must be done first, others later, etc.  This issue requests that we work to make sure the following two things are done:

1) That the order of the filters as implemented in the swell yamls is indeed the order they should be applied - logically consistent and also consistent w/ what GSI does.

2) That the machinery that processes the observation yamls and creates the final yaml for the variational analysis, or UFO, of HOF, does not get shuffled or alphabetically organized. Within a given filter, the order of statements is not important, but the sequence of each filter is of fundamental importance and it should not be changed.

Dooruk:
In `Run..` tasks the configurations are put together by the `jedi_dictionary_iterator` method and gets reordered alphabetically. This PR fixes that, see #409 for more details.

This closes #409 for now. There could be better ways but @rtodling needs a quick solution.

I mentioned some other options here:
https://github.com/GEOS-ESM/swell/issues/409#issuecomment-2293740551

### DIFF
--- a/src/swell/tasks/run_jedi_variational_executable.py
+++ b/src/swell/tasks/run_jedi_variational_executable.py
@@ -7,7 +7,7 @@
 
 # --------------------------------------------------------------------------------------------------
 
-
+from collections import OrderedDict
 import os
 import yaml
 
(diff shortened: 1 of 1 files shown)


Dooruk:
I tested with `t1test` `hofx` and `3dvar` passes.

ashiklom (approved):
Assuming you've tested this and it works, I'm happy to approve, at least as a short-term solution.

rtodling (approved):
I confirm that this works and gives zero diffs ... oh boy, it is so much more pleasant to read the yaml file now.

@Dooruk Thank you a million times for this.

Dooruk:
> I confirm that this works and gives zero diffs ... oh boy, it is so much more pleasant to read the yaml file now.
> 
> @Dooruk Thank you a million times for this.

My pleasure, it's a Christmas gift for you 🎁

Issue #378, opened 2024-06-28
BUFR2IODA Conversion


Dooruk:
The following lays out steps required for us to incorporate the capabilities of having swell convert bufr files to IODA on the fly. This is an edited version of the original written by @Dooruk. I will try to put some names around the items below to sort of give an idea of who might be best to work on a given item, but please take those as suggested names. If anyone feels that's not quite something within their realm, no problem, just let us know.

- [x] GEOSadas: move obsys.rc to sidb in mksi (Ricardo) - GEOS_mksi tag: g5.41.2
- [x] GEOSadas: pull acquire applications into a repo of their own - see below (Ricardo)
- [x] SWELL needs to learn how to extract the obs classes available and applicable to a given cycle time; that is, given a set of obs-classes, SWELL should read from the present databased (obsys.rc), extract the class of interest, see if data is available for the date/time of request, and create a list of fully resolve file names from the database that can then be passed on the either (or both) of the step below. (Ricardo) - Akira and R had a conversation about this. R will bring the GEOSDAS machinery to bring observations into the run experiment given a series of obsclass, a date and a couple of other parameters. This capability is based on a series of Perl scripts; they will require the installation in SWELL to know than simply handling Python scripts. I will need to work with Michael and Alexey on this. The command line to bring in data into the run env of converter suite will be as simple as the following:

`acquire_obsys -v -d $FVWORK -s $SPOOL -la $nlook -ssh  \
                 -strict $nymd $nhms $nfreq $nstep $OBSCLASS
`

where OBSCLASS defines the classes, as for example: disc_airs_bufr,disc_amsua_bufr,gmao_amsr2_bufr
nymd and nhms are the GEOS convention for 8-digit date and 6-digit time, and I will explain the other variables when I add the PR to that will allow swell to pull the required codes the an adequate ESM-repo (see https://github.com/GEOS-ESM/swell/pull/498)

To be more specific about how this can be exercised in SWELL, I am inserting here an set of c-sh lines (that can be placed in a file and turned into an executable shell program - this illustrates how to actually use the command line above. The only parameters that need to be dealt with in a swell context are the env variables used in the c-sh script below and the date and time parameters: nymd, nhms - everthing else can (for the time being) be wired w/o loss of functionality:

`#!/bin/csh

source /discover/nobackup/projects/gmao/advda/swell/JediBundles/fv3_soca_SLES15/build-intel-release/modules

[...]

PR #486, opened 2024-12-18
Automatically set up GMAO_perllib

mranst:
This PR modifies `setup.py` to clone and install `GMAO_perllib` at the `g1.0.1 tag upon pip install (see #378). Source files are cloned to `src/`, and installed to `~/.local/bin` (so ~/.local must exist).

 I wasn’t entirely sure how to go about installing the library. I don’t think pip has much capability to elegantly support installing non-python libraries, so this basically just takes the form of a script running in setup.py that manually runs the install commands in the shell. I contemplated installing it to the python virtual environment, but that feels like it could potentially get very messy with different users having different python environment setups. I settled on installing to `~/.local/bin`, since I think other `swell` commands are stored in that directory but I’m unsure of the most ideal location.

This uses the `CMakeLists.txt` from `GMAO_perllib`, which uses a command `esma_set_this()`, which I can’t figure out the nature of. From the comments, it seems to be for interactions with other libraries, so I’ve made it so that line is commented out before running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,5 @@ dmypy.json
 .pyre/
 *.egg-info
 
+# GMAO_perllib files
+src/GMAO_perllib
--- a/requirements-github.txt
+++ b/requirements-github.txt
@@ -9,3 +9,4 @@ f90nml==1.4.4
 questionary==1.10.0
 flake8==6.0.0
 netCDF4==1.6.5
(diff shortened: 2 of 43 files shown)


rtodling (commented):
Michael, I have problems checking out your branch - can you please redo this in a branch following out conversations:

feature/mranst/setup_perllib

Thank you.

PR #488, opened 2024-12-19
Platform for cascade lake nodes running sles15

### LINKED ISSUE #487: (SE) Cascade nodes on SLES15
**NCCS update:**

To access the SLES15 Cascade Lake nodes, please add the following Slurm
directives to your job script:

```
#SBATCH --reservation=sles15_cas
#SBATCH --constraint=“cas”
```

This should be rather straightforward to introduce, just as a separate platform so the `slurm.yaml` in `/platforms` would change. The confusing part will be using different node model & resolution combinations for different applications on two different type of nodes but we can worry about it later.

I think we could introduce another platform, say `nccs_discover_cascade`, and at some point rename `nccs_discover_sles15` to `nccs_discover_milan` and erase `nccs_discover`.

mranst:
This PR adds a platform for cascade lake nodes on Discover running sles15 (#487). I had to adjust the code tests since the convention that ‘sles15=milan node’ is no longer applicable. We probably need to figure out a better way to better distinguish milan and cascade nodes in the future. I tested various methods in the python `platform` package and I couldn’t find one that reports the microarchitecture, they all just report “x68_64”. For now I just changed the platform test so that it asserts that `sles4.12=nccs_discover`

### DIFF
--- a/src/swell/deployment/platforms/nccs_discover_cascade/modules
+++ b/src/swell/deployment/platforms/nccs_discover_cascade/modules
@@ -1,0 +1,60 @@
+# Module initialization
+# ---------------------
+source /usr/share/lmod/lmod/init/bash
+
+# Purge modules
+# -------------
+module purge
+
+# Spack stack modules
+# -------------------
+module use /discover/swdev/jcsda/spack-stack/scu16/modulefiles
+module load ecflow/5.11.4
+module use /gpfsm/dswdev/jcsda/spack-stack/scu16/spack-stack-1.7.0/envs/ue-intel-2021.10.0/install/modulefiles/Core
+module load stack-intel/2021.10.0
+module load stack-intel-oneapi-mpi/2021.10.0
+module load stack-python/3.10.13
+module load git-lfs/3.4.0
+module load py-pip/23.1.2
+module load jedi-fv3-env
+module load soca-env
+module load gmao-swell-env/1.0.0
+module unload gsibec crtm fms
+module load fms/2023.04
+
+# JEDI Python Path
+# ----------------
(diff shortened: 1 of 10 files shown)


Dooruk:
Great, were you able to test creating and launching a suite with sles15 cascade nodes?

mranst:
> Great, were you able to test creating and launching a suite?

Yes, should've mentioned that

src/swell/deployment/platforms/nccs_discover_cascade/modules
```diff
@@ -0,0 +1,60 @@
...
+# ---------------------
+source /usr/share/lmod/lmod/init/bash
+
+# Purge modules
+# -------------
+module purge
+
+# Spack stack modules
+# -------------------
+module use /discover/swdev/jcsda/spack-stack/scu16/modulefiles
```
Dooruk: Hmm, I'm a bit surprised SCU16 modules worked for this because they were introduced for SLES12 as SCU16 was strictly using SLES12 OS.

mranst: Interestingly, I tested it with both and it doesn't seem to make a difference. I changed it to scu17 in the latest commit

Dooruk: Sounds good, if it works it works!

Dooruk (approved):
This is good to merge once it is updated with `develop`. The bit about identifying OS will become obsolete once SLES12 OS is abandoned so I wouldn't worry about that.

mranst:
> This is good to merge once it is updated with `develop`. The bit about identifying OS will become obsolete once SLES12 OS is abandoned so I wouldn't worry about that.

Oops, thought I had done that, thanks

PR #491, opened 2025-01-17
A prototype for alternative task and suite questions

mranst:
This is a prototype for an alternative method of handling task and suite questions. Rather than having a set of questions that tell Swell what tasks they belong to, tasks and suites can be defined as a list of questions, whose properties are defined separately. All Swell questions have been refactored as python dataclass objects, and have had their task/suite declarations transformed into a list of questions associated with each task and suite. 

`prepare_config_and_suite.py` has been modified to accept this new approach. Rather than reading the entire list of all potential suite and task questions for everything in swell, only the list of questions for the particular suite being run are read, as well as only questions for the tasks needed by the suite.

From what I’ve tested, it behaves almost identically to the previous method as far as creating `experiment.yaml` is concerned, with the exception that some questions appear in a different order, which should probably be fixed for readability. `code_tests` will need to be modified as well, but I wanted to get opinions on whether this is even a good method.

I believe the major positive to this method are that it is more intuitive to look at things from a suite and task perspective, rather than having questions defining what task and suites they are a part of, like Alexey said in (#314). Having questions defined per suite and task also gives some flexibility in letting us spread out definitions a bit, rather than having to read it all in a centralized place, then throw out what we don’t need. The python dataclasses also will allow for some more powerful options for value assignments.

As for cons, in its current configuration I’m not sure this does much to fix the concern of having all swell-wide questions within two files. It adds separate files for declaring all questions for each task and suites `task_question_list.py`, `suite_question_list.py`, which we could spread out into separate files (maybe even into the existing task python files). It also doesn’t yet address the need to potentially have different defaults depending on the context. I think we could probably implement a simple system where `default_value` is defined as a dictionary with some basic dependencies like task or suite being run, but that might get messy if you have multiple tasks using the same question during setup. 

Let me know if this looks like a good way to proceed, or any suggestions!

### DIFF
--- a/src/swell/suites/suite_question_list.py
+++ b/src/swell/suites/suite_question_list.py
@@ -1,3 +1,4 @@
+# --------------------------------------------------------------------------------------------------
 # (C) Copyright 2021- United States Government as represented by the Administrator of the
 # National Aeronautics and Space Administration. All Rights Reserved.
 #
@@ -7,9 +8,6 @@
 
 # --------------------------------------------------------------------------------------------------
 
-
-import os
-from dataclasses import dataclass
 
 from swell.utilities.swell_questions import SuiteQuestionList
 import swell.suites.suite_questions as sq
--- a/src/swell/suites/suite_question_list.py
+++ b/src/swell/suites/suite_question_list.py
@@ -1,0 +1,212 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+import os
+from dataclasses import dataclass
+
+from swell.utilities.swell_questions import QuestionList
+
+
+# --------------------------------------------------------------------------------------------------
+
+all_models = QuestionList(
+    list_type="model_dependent",
+    name="all_models",
+    questions=[
+        "cycle_times",
+        "ensemble_hofx_packets",
+        "ensemble_hofx_strategy",
+        "skip_ensemble_hofx",
+        "window_type"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+geos_marine = QuestionList(
+    list_type="model_dependent",
+    name="geos_marine",
+    questions=[
+        "marine_models"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+_3dfgat_atmos = QuestionList(
+    list_type="suite",
+    name="3dfgat_atmos",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+_3dfgat_cycle = QuestionList(
+    list_type="suite",
+    name="3dfgat_cycle",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "marine_models",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+_3dvar = QuestionList(
+    list_type="suite",
+    name="3dvar",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "marine_models",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+_3dvar_atmos = QuestionList(
+    list_type="suite",
+    name="3dvar_atmos",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+_3dvar_cycle = QuestionList(
+    list_type="suite",
+    name="3dvar_cycle",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "marine_models",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+all_suites = QuestionList(
+    list_type="suite",
+    name="all_suites",
+    questions=[
+        "experiment_id",
+        "experiment_root"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+convert_ncdiags = QuestionList(
+    list_type="suite",
+    name="convert_ncdiags",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+forecast_geos = QuestionList(
+    list_type="suite",
+    name="forecast_geos",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "start_cycle_point"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+hofx = QuestionList(
+    list_type="suite",
+    name="hofx",
+    questions=[
+        "cycle_times",
+        "final_cycle_point",
+        "marine_models",
+        "model_components",
+        "runahead_limit",
+        "start_cycle_point",
+        "window_type"
+    ]
+)
+
+
+# --------------------------------------------------------------------------------------------------
+
+localensembleda = QuestionList(
+    list_type="suite",
+    name="localensembleda",
+    questions=[
+        "cycle_times",
+        "ensemble_hofx_packets",
+        "ensemble_hofx_strategy",
+        "final_cycle_point",
+        "marine_models",
+        "model_components",
+        "runahead_limit",
+        "skip_ensemble_hofx",
+        "start_cycle_point"
+    ]
+)
+
+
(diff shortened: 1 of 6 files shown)


mathomp4:
@mranst Just a query, as I notice you are using a fork. Do you not have write access here?

mranst:
> @mranst Just a query, as I notice you are using a fork. Do you not have write access here?

I believe I do, I'm just used to working on my own fork. I should probably branch from here directly in the future.

src/swell/suites/suite_question_list.py
```diff
@@ -0,0 +1,212 @@
...
+    list_type="model_dependent",
+    name="all_models",
+    questions=[
+        "cycle_times",
+        "ensemble_hofx_packets",
+        "ensemble_hofx_strategy",
+        "skip_ensemble_hofx",
+        "window_type"
+    ]
+)
```
ashiklom: Thinking out loud: Any reason not to store the question objects themselves directly in these objects? I.e., Something like:

```python
import swell.suites.suite_questions as sq
# [...]
all_models = QuestionList(
    list_type="model_dependent",
    name="all_models",
    questions=[
        sq.cycle_times,
        sq.ensemble_hofx_packets,
        sq.ensemble_hofx_strategy,
        sq.skip_ensemble_hofx,
        sq.window_type
    ]
)
```

That eliminates an additional layer of indirection and lets us rely that much more on Python's static checkers to catch issues.

Moreover, if each question is a class or a function rather than a static object (i.e., something that takes arguments), we can put initialization parameters directly into this question list (if that's a capability we need). One use case of this might be specifying different defaults for different models; for example, multiple suites might prompt the user about skipping ensemble HOFX, but some suites might default to skipping HOFX while others might default to _not_ skipping HOFX.

```python
# [...]
question_list = [
  cycle_times(retry=False),
  skip_ensemble_hofx(default=False),
  ...
]
```

src/swell/utilities/swell_questions.py
```diff
@@ -0,0 +1,62 @@
...
+    question_type: str
+    default_value: str
+    prompt: str
+    depends: dict = None
+    models: list = None
+    ask_question: bool = False
+    options: str = None
+
+    def get(self, attr, default=None):
+        return getattr(self, attr, default)
```
ashiklom: More brainstorming: We should look for opportunities to be a bit stricter with types here. In particular, some of these things can be specific enums with specific values (for example, `suite_type` and `list_type` both have a small number of fixed values that would be well supported by enums). Other things can be more precise types or even custom swell-specific classes; for example, `depends` is not just _any_ dict but really a list of Swell questions with specific configurations (right?).

mathomp4:
> > @mranst Just a query, as I notice you are using a fork. Do you not have write access here?
> 
> I believe I do, I'm just used to working on my own fork. I should probably branch from here directly in the future.

No problem. Just making sure there wasn't a GitHub issue! Forks are nice in that you have control!

ashiklom:
This is a solid first pass, nice work! It's great to see a real version of this, even in draft form.

I left a few isolated thoughts as comments, and I may add more as we go. Big picture: I would "checkpoint" the current commit and now start doing a bit more exploration of the practical value of this refactor to fully justify it. Some ideas:

1. Be stricter/more sophisticated with our types, to start taking advantage of Python's static analysis tools.
2. Simplify our implementation of the relationship between questions and tasks. I.e., Right now, there's a lot of reasonably complex code for actually asking the questions, processing the answers, and then passing them into the tasks. I would love to simplify the design of that machinery through more elegant object-oriented composition/inheritance in Python.
3. Shrink and/or reorganize the current huge task questions list? E.g., If there's a lot of redundancy in questions across tasks, maybe we can capture that redundancy in lists that multiple tasks can share. Or, rather than defining questions in a single big file, maybe we define questions related to specific Swell functionality alongside the implementation of that functionality (e.g., put JEDI-related questions next to Swell's JEDI-related functions).

Note: Probably not all of these ideas will work out in practice. Also, some of these ideas will point to tweaks to the implementation of the Questions/Tasks classes and lists.

src/swell/suites/suite_question_list.py
mranst: Great point, much more powerful that way

mranst:
> This is a solid first pass, nice work! It's great to see a real version of this, even in draft form.
> 
> I left a few isolated thoughts as comments, and I may add more as we go. Big picture: I would "checkpoint" the current commit and now start doing a bit more exploration of the practical value of this refactor to fully justify it. Some ideas:
> 
>     1. Be stricter/more sophisticated with our types, to start taking advantage of Python's static analysis tools.
> 
>     2. Simplify our implementation of the relationship between questions and tasks. I.e., Right now, there's a lot of reasonably complex code for actually asking the questions, processing the answers, and then passing them into the tasks. I would love to simplify the design of that machinery through more elegant object-oriented composition/inheritance in Python.
> 
>     3. Shrink and/or reorganize the current huge task questions list? E.g., If there's a lot of redundancy in questions across tasks, maybe we can capture that redundancy in lists that multiple tasks can share. Or, rather than defining questions in a single big file, maybe we define questions related to specific Swell functionality alongside the implementation of that functionality (e.g., put JEDI-related questions next to Swell's JEDI-related functions).
> 
> 
> Note: Probably not all of these ideas will work out in practice. Also, some of these ideas will point to tweaks to the implementation of the Questions/Tasks classes and lists.

Thanks, these are good points! I'll see what I can do

mranst:
@rtodling Per request, here's a simple example of the old method vs. the new method of defining questions:

# The old yaml syntax:

## Suite questions
https://github.com/GEOS-ESM/swell/blob/53e69e9c89c096dd54a4d6d49109f4eaeb2178d7/src/swell/suites/suite_questions.yaml#L1-L29

## Task questions
https://github.com/GEOS-ESM/swell/blob/53e69e9c89c096dd54a4d6d49109f4eaeb2178d7/src/swell/tasks/task_questions.yaml#L15-L41

Each question defines what suites and tasks they belong to. Every single question is read in at the start, then whittled down depending on what is needed. Let's say that we're running `3dvar`, which only has `PrepareAnalysis` as a task, as defined in it's `flow.cylc` (just a crude example). 

The suite questions are  read at the beginning, then `ensemble_hofx_strategy` is eliminated since `3dvar` is not listed in its applicable `suites`.

Like with suites,  listed task questions are input into a dictionary at the start, then  `background_error_model` is deleted, while `analysis_variables` remains. This method necessitates all questions for everything in Swell to be read at the start, then removed as appropriate. Being in yaml format, it also complicates dynamically setting default values based on suite or task.

# The new syntax:

Each suite and task is defined as a corresponding list of questions
## Suite questions
https://github.com/GEOS-ESM/swell/blob/53e69e9c89c096dd54a4d6d49109f4eaeb2178d7/src/swell/suites/suite_question_list.py#L47-L57

## Task questions
https://github.com/GEOS-ESM/swell/blob/53e69e9c89c096dd54a4d6d49109f4eaeb2178d7/src/swell/tasks/task_question_list.py#L46-L54

When building the question dictionary, swell now only seeks out the questions listed under the suite and tasks that are applicable, though some whittling down is still necessary depending on model, etc. The properties of the questions themselves are defined separately:

https://github.com/GEOS-ESM/swell/blob/53e69e9c89c096dd54a4d6d49109f4eaeb2178d7/src/swell/tasks/task_questions.py#L33-L43

As Alexey noted, in its current form this PR is more of a syntax change than a fix to the underlying issues with the current system of questions, though there is potential to leverage python to address these issues. Like he noted above, we could assign different default values per-task or suite in the question lists. If you wanted different options for `path_to_ensemble` depending on suite, for example:

In `suite_question_list.py`:
```python
var = SuiteQuestionList(
    questions = [
        sq.path_to_ensemble(default_value = 'lat-lon')
    ]
)
letkf = SuiteQuestionList(
    questions = [
        sq.path_to_ensemble(default_value = 'cubed')
    ]
)
```

We could also set lists of commonly associated questions (like workflows that use JEDI, for example) to reduce the huge number of questions in the lists, and better use type-checking to verify validity.


Edited to fix permalinks

rtodling:
@mranst Thank you so much the example above - and the comments that come w/ it. I understand things considerably better now. I like the idea, or the potential, of eventually separating things a little more and if I understand it well, this is what the changes open the door for happening. 

I wouldn't have a problem having these initial changes go in, provided you can get zero diff in the tests - just stating the obvious.

Thank you for writing the example down.

Dooruk:
Thanks for providing the example, it made it more clear for me. Like you mentioned, for now looks mostly like a refactoring of the current structure but tidier. 

One question I have, as I understand, with this new structure we are abandoning parsing `flow.cylc`. That provided some advantage for me in terms of activating/deactivating certain tasks and hence relevant task questions:

https://github.com/GEOS-ESM/swell/blob/966ec27883d0e1e5e43a6aa936cc7076485726b3/src/swell/suites/3dfgat_cycle/flow.cylc#L97-L103

Could we have similar sort of control, conditionals depending on yaml values, with the new structure?

mranst:
> Thanks for providing the example, it made it more clear for me. Like you mentioned, for now looks mostly like a refactoring of the current structure but tidier.
> 
> One question I have, as I understand, with this new structure we are abandoning parsing `flow.cylc`. That provided some advantage for me in terms of activating/deactivating certain tasks and hence relevant task questions:
> 
> https://github.com/GEOS-ESM/swell/blob/966ec27883d0e1e5e43a6aa936cc7076485726b3/src/swell/suites/3dfgat_cycle/flow.cylc#L97-L103
> 
> Could we have similar sort of control, conditionals depending on yaml values, with the new structure?

To my knowledge (admittedly I get confused sometimes with all the different interactions going here), at the step of the process when the suite file is parsed, everything should be the same as before, i.e. `self.experiment_dict` has equivalent members as with the original method, so it would have the same behavior as before. Though, I may have to do some more extensive testing once I finish what I want to accomplish with the revision.

https://github.com/GEOS-ESM/swell/blob/966ec27883d0e1e5e43a6aa936cc7076485726b3/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L388-L391

PR #492, opened 2025-01-22
Adding EMC L3 SST, getting rid off obsolete observers

Dooruk:
We were using old EMC observers but now we rely mostly on GMAO/ODAS observations. Obs sources are similar between EMC and ours (i.e., altimeter), however for SST retrievals (and sea_ice data in the future), we will use EMC/GDAS sources, so we can use two `obs_provider` to check both sources (`odas`, `gdas_marine`). 

EMC conducts online conversion for their observers, meaning  we have to devise a way to obtain those files in the long term, in the short term we are bound to use observations during their experiment period (golden period) only.

I started optional partitions in observation yamls, some GeoVals, such as `sea_ice_area` depend on `cice` state being active or not. This is controlled by `marine_models` key in SWELL so Jinja2 will check that automatically.

Further, this standardizes filters in observation YAMLs, and uses `absolute_threshold` instead of `threshold`.

@mranst just tagging you for FYI, but this branch passes the tier 1 tests.

- [x] Fix MERRA2 sea-ice geovals


Dooruk:
Yikes, forgot about `hofx` and cycling suites, converting this to draft 🙅

Dooruk:
This is ready for a review and merge now.

DavidSRussell (approved):
Didn't test it, but looks good to me. Progress!

Issue #435, opened 2024-09-27
Adding Bufr2IODA to SWELL


rtodling:
This lay out the steps require for us  to incorporate the capabilities of having swell convert bufr files to IODA on the fly.

[] GEOSadas: move obsys.rc to sidb in mksi.

There are two options to implement the converters:
1. Offline converter:
  [] create location within SWELL to hold yaml files involved in convert of all relevant files
  [] work on radiance first
3. Online converter:

Dooruk:
There is the following issue already, would you like to add some these steps there?

https://github.com/GEOS-ESM/swell/issues/378

PR #496, opened 2025-01-29
Adding feature: iodaconverters for bufr files.

cohen-seth:
## Description
This PR adds the base suite and tasks for integrating the IODA converter 'bufr2ioda.x' within the swell workflow. This is only for BUFR files specifically. 

This PR addresses the issues link below:
- https://github.com/GEOS-ESM/swell/issues/435
- https://github.com/GEOS-ESM/swell/issues/378 - the section starting at '1. Offline converter: once SWELL the steps above are complete:'

I used existing files for ncdiag as templates for the flow.cylc and get_bufr.py files. 

Purpose of scripts:

(1) src/swell/tasks/**bufr_to_ioda.py**: Executes the conversion using both get_bufr.py and get_bufr_iodaconv_yaml.py

(2) src/swell/tasks/**get_bufr.py** : Locates the bufr files and creates links in the swell experiments run folder(s). 

(3) src/swell/tasks/**get_bufr_iodaconv_yaml.py** : The ioda converter (bufr2ioda.x) requires a yaml file within which is the data input path, the data output path and information about variables which are specific to observation type (i.e. AMSUA vs GPSRO vs ...) - bufr2ioda.x converts ONE file at a time so **get_bufr_iodaconv_yaml.py** enables the process of locating the input bufr file, identifying the observation type, specifying the output path, and generating the yaml file based on the observation type. The current method uses template yaml files from the iodaconv/test/testinput/ folder; these files are generated by swell when it builds JEDI so the relative paths for them should stay consistent.

(4) src/swell/suites/convert_bufr/**flow.cylc** : the suite for all the above

## Dependencies

1.  IODA Converters (JEDI) must be included in the jedi build
2. The yaml files from the iodaconv/test/testinput/ folder must be included in the jedi build

### DIFF
--- a/src/swell/suites/convert_bufr/flow.cylc
+++ b/src/swell/suites/convert_bufr/flow.cylc
@@ -1,0 +1,91 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+# --------------------------------------------------------------------------------------------------
+
+# Cylc suite for executing geos_atmosphere ObsFilters tests
+
+# --------------------------------------------------------------------------------------------------
+
+[scheduler]
+    UTC mode = True
+    allow implicit tasks = False
+
+# --------------------------------------------------------------------------------------------------
+
+[scheduling]
+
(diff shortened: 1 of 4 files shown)


rtodling (commented):
Seth, I will say that as I have laid out in the original issue I filed on the bufr to Ioda , the bring of bufr files needs to use the existing GMAO database. Michael has introduced the machinery from GEOS that controls the obs-db ... we need to connect what you are doing here with that ... we'll talk more about it on Monday.

PR #500, opened 2025-02-10
Fix broken links in docs

ashiklom:
Fixes a few broken links in our documentation. These are files that don't / no longer exist or were moved without updates to the corresponding links.

I found these by clicking through the docs by hand. Everything else seems to be fine, but there may be additional broken links in there that I missed.

(A fun little side project would be to write a Python script and GitHub action that scans the docs for broken links...)


Dooruk:
@ashiklom is this ready for merge or are you still working on it?

ashiklom:
Ready for review and, assuming it looks good, merge.

Dooruk (approved):
Looks good. I am in the process of reviving a cycling tier1 test which includes running the coupled model so will review these again.

Dooruk:
Question for next time: Is it possible not to trigger Tier 1 actions when only document changes are involved?

PR #511, opened 2025-02-18
Fixed bug where models in override.yaml would overwrite tier1 defaults

### LINKED ISSUE #510: For `3dvar_cycle`, filling in of `experiment.yaml` depends on whether `geos_ocean` model is included in override yaml
When creating a `3dvar_cycle` experiment, a different set of default values is used to fill out the `experiment.yaml` file if I include the `geos_ocean` under `models` in the override yaml than if I have no `models` listed. For example, without `geos_ocean` in the override file, the variables give a low-resolution `geos_ocean` with a 6-hour window, where with `geos_ocean`, they give high resolution with a 12-hour window. It does not appear to depend on which elements are actually included under `geos_ocean` in the override.

mranst:
This PR fixes a bug (#510 ) where any model being added in `override.yaml` would cause the entire override subdictionary for that model to be overwritten. Since the defaults that swell uses are actually sourced from the tier1 overrides immediately before that step, it means that the usual defaults are deleted and the values from elsewhere (`task_questions.yaml`) are used. This fix just changes the updating of the override dict to use the `update_dict` function from utilities, which can handle replacing values in a dict without overwriting the entire dict.

### DIFF
--- a/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
+++ b/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
@@ -311,10 +311,10 @@ class PrepareExperimentConfigAndSuite:
         if self.override is not None:
 
             if isinstance(self.override, dict):
-                override_dict.update(self.override)
+                override_dict = update_dict(override_dict, self.override)
             elif isinstance(self.override, str):
                 with open(self.override, 'r') as ymlfile:
-                    override_dict.update(yaml.safe_load(ymlfile))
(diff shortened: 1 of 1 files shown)


ashiklom (approved):
Good find!

For future reference: As of Python 3.9, you can use the `|` and `|=` operators to update dictionaries. There may be places we can use that to simplify our code for stuff like this.

https://stackoverflow.com/questions/64210846/when-to-use-dictionary-merge-vs-update-operator

Dooruk (approved):
good find!

PR #512, opened 2025-02-19
Reviving 3dvar_cycle and 3dfgat_cycle suite defaults

### LINKED ISSUE #499: creating tile.bin in cycle_dir

Dooruk:
- Includes updating 5-deg experiment setup and restart file folders.

- 3fgat_cycle uses cice6 + mom6 marine_models meanwhile 3dvar_cycle uses mom6 only

- Soca2cice configuration had some changes recently, this PR also addresses that.

- I resorted to using `marine_models` key in geos_marine configurations to separate cases where mom6 is active and cice6 is not. While this is a working approach not the only one. https://github.com/GEOS-ESM/swell/pull/491 could address this in a different way.

- I'm taking out `ufo_testing` from tier1 and tier2 tests and adding `3dvar_cycle` so after months we will have tier2 test running (fingers crossed)

- As I mentioned, this is another step closer to abandoning `geos_ocean`, almost there.

@DavidSRussell this is ready for testing, please take a look (you will need spack-stack 1.9 modules)

closes #499

### DIFF
--- a/src/swell/__init__.py
+++ b/src/swell/__init__.py
@@ -9,4 +9,4 @@ import os
 repo_directory = os.path.dirname(__file__)
 
 # Set the version for swell
-__version__ = '1.9.6'
+__version__ = '1.9.7'
--- a/src/swell/configuration/jedi/interfaces/geos_marine/model/background.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_marine/model/background.yaml
@@ -6,9 +6,11 @@ ocn_filename: 'MOM6.res.{{local_background_time}}.nc'
 ice_filename: 'cice.res.{{local_background_time}}.nc'
 {% endif %}
 state variables:
+{% if 'cice6' in marine_models %}
   - sea_ice_area_fraction
   - sea_ice_thickness
   - sea_ice_snow_thickness
+{% endif %}
   - sea_water_salinity
   - sea_water_potential_temperature
   - sea_surface_height_above_geoid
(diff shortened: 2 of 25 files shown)


Dooruk:
Still having a problem with 3dfgat_cycle (cice6+mom6) but 3dvar_cycle should work @DavidSRussell and @gmao-cda

DavidSRussell:
OK, so `3dvar_cycle` seems to work for me, as I get no task failures and `cycle_done` in the run directory for both times. Two minor things:
- I still had to change Python 3.10 to 3.11 in the `.lua` file before running
- There's a run directory for `20210702T180000Z` even though the final time is 12Z. Maybe that's expected?

Dooruk:
> OK, so `3dvar_cycle` seems to work for me, as I get no task failures and `cycle_done` in the run directory for both times.
Great, can you give an approval please? Though if something breaks it's your fault 😄 (kidding ofcourse)

 Two minor things:
> 
>     * I still had to change Python 3.10 to 3.11 in the `.lua` file before running
Yeah that is related to spack-stack modules update

>     * There's a run directory for `20210702T180000Z` even though the final time is 12Z. Maybe that's expected?
I don't have a save final step option yet so for now it is expected. Those restart files can be used for another test though.

DavidSRussell (approved):
Both `3dvar_cycle` and `3dfgat_cycle` worked for me with the default setup after I changed to Python 3.11 in the module file.

PR #514, opened 2025-02-21
A prototype for revised suite and task questions, redux

mranst:
See conversation in (#491), I re-submitted this PR as a branch to swell.

The latest commit extends this method with several features:

- Question defaults are now stored in dataclasses, so that they can be easily called with different default values or options
- Question lists can be created directly with question objects, or with references to other lists. The method `expand_question_list()` iteratively parses the questions and expands sublists to create a list of questions per suite and task. Model-dependent questions can also have seperate defaults defined per model.

https://github.com/GEOS-ESM/swell/blob/0c22359d261ec9b43662130fbac1bc60691c96d9/src/swell/suites/suite_questions.py#L35-L45

https://github.com/GEOS-ESM/swell/blob/0c22359d261ec9b43662130fbac1bc60691c96d9/src/swell/suites/suite_questions.py#L59-L64

Questions have a hierarchy depending on order. If there are duplicate questions (denoted by name), the ones further down the list take priority. For example:

```python
list1 = QuestionList(
    questions=[
        sq.start_cycle_time(default_value='2025-02-20'),
        sq.final_cycle_time(default_value='2025-02-21')
    ]
)
list2 = QuestionList(
    questions=[
        list1,
        sq.start_cycle_time(default_value='2025-02-19')
    ]
)
```
Using `list2.expand_question_list()` returns a list containing the objects created by
```
sq.start_cycle_time(default_value='2025-02-19')
 sq.final_cycle_time(default_value='2025-02-21')
```

Using this hierarchy-based approach allows for efficient declaration of question lists, while also allowing for complete control for suites to define their own defaults. The question dictionary for each experiment is constructed in task-to-suite ascending priority.

```mermaid
flowchart TD
A[Get the tasks used by the suite according to flow.cylc] --> B{Load the questions associated with each task} --> C[Check the suite questions]--> D{If there are duplicates, use the ones from the suites}
```

This control means that defaults can easily be set directly, for instance tier1 defaults can be set directly in `suite_questions`, rather than needing a separate file to override defaults that are set by the questions:
https://github.com/GEOS-ESM/swell/blob/0c22359d261ec9b43662130fbac1bc60691c96d9/src/swell/suites/suite_questions.py#L77-L114

We could also easily create different configurations using this extension methodology, for example setting tier2 defaults could be as simple as 
```python
_3dvar_tier2 = QuestionList(
    questions=[_3dvar_tier1]
    geos_ocean=[
        tq.horizontal_resolution(default_value=1440x1080),
        tq.vertical_resolution(default_value=1440x1080)
    ]
)
```
This PR also adds support for setting model defaults dynamically. I haven't made use of it yet but going forward, defaults can be set like:
```python
@dataclass 
class question:
    default_value = {'depends_on_model': {'geos_ocean': 'x', 'geos_marine': 'y', 'geos_atmosphere': 'z'}
```
Conceivably, this could be used to assimilate the seperate yamls for each model directly into the question defaults, though it may be preferable to keep them separate


src/swell/utilities/swell_questions.py
```diff
@@ -0,0 +1,113 @@
...
+from typing import List, Union, Optional
+
+
+# --------------------------------------------------------------------------------------------------
+
+@dataclass
+class SwellQuestion:
+    """Basic dataclass for defining Swell questions for suites and tasks"""
+    question_name: str
+    dtype: str
```
ashiklom: Can we be more precise here with either a `Literal` or, better yet, an `Enum`? Something like:

```suggestion
    dtype: SQdtype
```

...where:

```python
class SQdtype(Enum):
    STRING = "string"
    STRING_CHECK_LIST = "string-check-list"
    BOOLEAN = "boolean"
    ISO_DURATION = "iso-duration"
    INTEGER = "integer"
    # etc...
```

At a minimum, Enums give us stricter type checking and make it easier to quickly figure out the full range of acceptable values. But, since Enums are just a class, they can also be extended with some useful methods; for example, here is some inspiration from ChatGPT:

```python
from enum import Enum
from datetime import datetime
import re

class DType(Enum):
    STRING = "string"
    CUSTOM_STRING = "custom-string"
    ISO_DATETIME = "iso-datetime"

    @property
    def is_string_type(self) -> bool:
        """Check if the type is string-based"""
        return self in (DType.STRING, DType.CUSTOM_STRING)

    @property
    def is_datetime_type(self) -> bool:
        """Check if the type is datetime-based"""
        return self == DType.ISO_DATETIME

    def validate_value(self, value: str) -> bool:
        """Validate if a value matches the dtype requirements"""
        if self == DType.STRING:
            return isinstance(value, str)
        
        elif self == DType.CUSTOM_STRING:
            # Example: Requires alphanumeric with underscores, 3-20 chars
            pattern = r'^[a-zA-Z0-9_]{3,20}$'
            return bool(re.match(pattern, value))
        
        elif self == DType.ISO_DATETIME:
            try:
                datetime.fromisoformat(value)
                return True
            except ValueError:
                return False

    def format_value(self, value: str) -> str:
        """Format a value according to dtype specifications"""
        if not self.validate_value(value):
            raise ValueError(f"Invalid value for {self.name}: {value}")

        if self == DType.STRING:
            return value
        
        elif self == DType.CUSTOM_STRING:
            return value.lower()
        
        elif self == DType.ISO_DATETIME:
            return datetime.fromisoformat(value).isoformat()

# Usage example:
def process_data(dtype: DType, value: str):
    if dtype.is_string_type:
        print("Processing string type data")
    
    if dtype.validate_value(value):
        formatted = dtype.format_value(value)
        print(f"Formatted value: {formatted}")
    else:
        print(f"Invalid value for {dtype.name}")

# Example usage:
process_data(DType.CUSTOM_STRING, "Hello_123")  # Valid
process_data(DType.ISO_DATETIME, "2023-12-25T12:00:00")  # Valid
process_data(DType.CUSTOM_STRING, "Hello@World")  # Invalid
```

mranst:
My latest commit refactors a lot of what I had before. All tier1 defaults have been assimilated into separate suite files. I've added a new code test for this method, and implemented new type checking for the cli and defaults. I've added a user's guide for adding suites to swell, it is fairly detailed but I'm sure there are some oddities I'm missing. For reference - the last 2 subsections detail how the new question method works.

Dooruk:
I see it says draft but is this ready to be tested in CLI?

mranst:
> I see it says draft but is this ready to be tested in CLI?

I kept it as draft since it's a somewhat sweeping change, and there's a few settings I need to assimilate from other pull requests before merging. It should be ready to do some cli testing, though

docs/adding_a_suite.md
```diff
@@ -0,0 +1,392 @@
...
+For context, tasks in `R1` are run as setup for later tasks. Here, `CloneJedi` is triggered first. When `CloneJedi` completes, `BuildJediByLinking` is triggered, but the question mark denotes that it is optional to complete. Should it fail, the task `BuildJedi` is triggered.
+
+The actions associated with these tasks is defined in the `runtime` section of `flow.cylc`. Here is the runtime section for the tasks above:
+
+```
+[[runtime]]
+
+...
+
+[runtime]
```
ashiklom, line 89: Two `[runtime]` sections. Is this a typo?

ashiklom: Or rather, I suspect the `[[runtime]]` above is a typo.

docs/adding_a_suite.md
```diff
@@ -0,0 +1,392 @@
...
+        default_value: str = "defer_to_platform"
+        question_name: str = "existing_jedi_build_directory"
+        ask_question: bool = True
+        depends: Dict = field(default_factory=lambda: {
+            "jedi_build_method": "use_existing"
+        })
+        prompt: str = "What is the path to the existing JEDI build directory?"
+        widget_type: WidgetType = WidgetType.STRING
+```
+Calling `existing_jedi_build_directory()` creates an object matching the default values for the class. Fields set as `defer_to_platform`, `defer_to_model`, and `defer_to_code` are later override with corresponding values sourced from:
```
ashiklom:
```suggestion
Calling `existing_jedi_build_directory()` creates an object matching the default values for the class. The specific values `defer_to_platform`, `defer_to_model`, and `defer_to_code` are special values that will be substituted as follows, depending on the `platform` and `model`: 
```

ashiklom: Not for this PR, but at some point, it would be nice to think about whether this could be done more elegantly, e.g., via a `field` constructor along the lines of:

```python
  default_value: str = field(default_factory = lambda: f"src/swell/deployment/platforms/{platform}")
```

That presupposes a way to access a variable called `platform` at runtime. One way to do this is to add `platform` as a required field and then use a [`__post_init__` hook](https://docs.python.org/3/library/dataclasses.html#post-init-processing) to fix these values. But that's really annoying if we have to pass platform explicitly into every one of these. Another possibility is to get clever with the `field(default_factory =. ...)` such that it is smart about detecting the platform...but that effectively introduces a (small) side effect, which may not be ideal either. Something to think about!

mranst: Agreed. Fortunately `platform` is specified at swell main and should be accessible at the point of experiment construction. I'm not super familiar with what `field` constructors are capable of, but it'd be great to be able to grab values easily like that

docs/adding_a_suite.md
```diff
@@ -0,0 +1,388 @@
...
+            qd.existing_jedi_build_directory(),
+            qd.existing_jedi_build_directory_pinned(),
+            qd.jedi_build_method()
+        ]
+    )
+```
+
+During experiment creation, Swell scans the suite's `flow.cylc` file to find all of the tasks used in the workflow. It then finds the corresponding task lists in `src/swell/tasks/task_questions.py`, and fits together a list of uniquely named questions from all of the lists. Questions have a priority depending on order. In the case of duplicate questions, those further DOWN the list take priority. For this reason, it is NOT RECOMMENDED to set different default values for tasks in `task_questions.py`, since questions may be overridden by a questions in a different task. 
+
+In this question infrastructure, suites priority over tasks. Any question specified in a suite configuration will override the default value for a question in one of its member tasks. This allows for easily setting different configurations for suites without having to specify redundant questions. For ease of use, model-dependent questions can be assigned directly in their respective lists.
```
ashiklom:
```suggestion
In this question infrastructure, **suites take priority over tasks**. Any question specified in a suite configuration will override the default value for a question in one of its member tasks. This allows for easily setting different configurations for suites without having to specify redundant questions. For ease of use, model-dependent questions can be assigned directly in their respective lists.
```

docs/adding_a_suite.md
```diff
@@ -0,0 +1,388 @@
...
+
+Consider the following example of suite questions for `3dvar` (in python, variable names cannot begin with digits):
+
+```python
+class SuiteQuestions(QuestionContainer, Enum):
```
ashiklom, line 302: For clarity:

```suggestion
from swell.utilities.question_defaults import QuestionDefaults as qd
from swell.suites.suite_questions import SuiteQuestions as sq

class SuiteQuestions(QuestionContainer, Enum):
```

ashiklom: Also, why do we inherit from `Enum` here? (Not saying we shouldn't --- just curious to know the reason).

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -93,77 +102,98 @@ def __init__(
...
+            # -----------------------------------------------------------------------
+            if key not in question_dictionary:
+                question_dictionary[key] = value
+            else:
+
+                # Otherwise, we need to check the suite question to
+                # see if there are any model-dependent fields which are not specified
+                # ----------------------------------------------------------------------------------------------------------------------
+                for sub_key, sub_val in question_dictionary[key].items():
+                    if isinstance(sub_val, dict) and 'depends_on_model' in sub_val.keys():
```
ashiklom: Here, and elsewhere you use `isinstance`: 

Strictly speaking, `isinstance` with a concrete class (rather than an abstract base class) is not recommended. "Program to an interface, not to an implementation".

Two possible alternatives:

1. (Recommended in this specific case) Use the `Mapping` abstract base class, which implicitly lets you eventually use this code with a dict _or anything dict-like_ (e.g., if we decide to use a special class for this in the future; the _literal_ thing this is doing checking that the object in question has dict methods, or "adheres to the Mapping interface"):

```python
from collections.abc import Mapping
# ...
if isinstance(sub_val, Mapping) and 'depends_on_model' in sub_val.keys():
  # do stuff...
```

2. "Easier to ask for foregiveness than permission" (EAFP) style.

```python
try:
  # do stuff with sub_val['depends_on_model']
except (TypeError, KeyError):
  pass
```

In this specific example, (1) is easier to implement (because having a lot of stuff that happens in a `try-except` block is also bad practice). But including (2) for future consideration.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -318,24 +346,24 @@ def override_with_external(self) -> None:
...
+                    val['default_value'] = override_dict[key]
+
+            # Iterate over the model_dep dictionary and override
+            # --------------------------------------------------
+            if self.suite_needs_model_components and 'models' in override_dict.keys():
+                for model, model_dict in self.question_dictionary_model_dep.items():
+                    for key, val in model_dict.items():
+                        if model in override_dict['models']:
+                            if key in override_dict['models'][model]:
+                                val['default_value'] = override_dict['models'][model][key]
```
ashiklom, line 373: Extra indentation changes the interpretation of this block. All of this will only run if `self.override is not None`.

Is that correct?

src/swell/suites/3dfgat_atmos/suite_config.py
```diff
@@ -0,0 +1,103 @@
...
+from swell.utilities.swell_questions import QuestionContainer, QuestionList
+from swell.utilities.question_defaults import QuestionDefaults as qd
+from swell.suites.suite_questions import SuiteQuestions as sq
+
+from enum import Enum
+
+
+# --------------------------------------------------------------------------------------------------
+
+class SuiteConfig(QuestionContainer, Enum):
```
ashiklom, line 20: Why do all of these also need to inherit from `Enum`?

src/swell/suites/3dfgat_atmos/suite_config.py
```diff
@@ -0,0 +1,103 @@
...
+
+    _3dfgat_atmos_base = QuestionList(
+        list_name="3dfgat_atmos",
+        questions=[
+            sq.common
+        ]
+    )
```
ashiklom: This is only used once, so I would delete this variable and embed everything in the `_3dfgat_atmos_tier1` (here and elsewhere).

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -93,77 +102,97 @@ def __init__(
...
-                        val.pop('suites')
-        for key in keys_to_remove:
-            del question_dictionary[key]
+                self.questions_per_task[task] = []
+
+        # Get a list of all questions associated with the suite, except for those specified
+        # seperately for models
+
+        suite_config_obj = getattr(importlib.import_module(
+            f'swell.suites.{self.suite}.suite_config'), 'SuiteConfig')[self.suite_config].value
```
ashiklom: This approach feels more complicated and brittle than it needs to be. My suggestion would be to do the following:

1. Define each suite directly as a python object, probably of a common class like `Suite`.
2. Define an `Enum` called `all_suite_configs` or something that contains all the suites from (1).
3. Here, just query that `all_suite_configs` object by name.

(If this requires major re-engineering then let's punt on it till a future PR).

src/swell/suites/3dfgat_atmos/suite_config.py
ashiklom: Also, per my other comment, I suggest exposing the underlying objects in here directly; e.g.,

```python
from swell.utilities.swell_questions import QuestionContainer, QuestionList
from swell.utilities.question_defaults import QuestionDefaults as qd
from swell.suites.suite_questions import SuiteQuestions as sq

_3dfgat_atmos = QuestionList(
  list_name = "3dfgat_atmos",
  questions = [
    sq.common,
    qd.start_cycle_point(...),
    ...
  ]
)
```

Then, elsewhere (in a different file), we have a master list of suites along the lines of:

```python
from swell.suites.3dfgat_atmos.suite_config import _3dfgat_atmos, _3dfgat_atmos_tier1
from swell.suites.hofx.suite_config import hofx, hofx_tier1
# ...etc...

from enum import Enum

all_suites = Enum("all_suites", [
  ("3dfgat_atmos", _3dfgat_atmos),
  ("3dfgat_atmos_tier1", _3dfgat_atmos_tier1),
  ("hofx", hofx),
  # etc...
])
```

And then, instead of the `importlib` stuff earlier, we could just do:

```python
from swell.suites.all_suites import all_suites_dict
# ...
  suite_configs = all_suites[suite]
```

This adds a (small) amount of manual work and duplication in registering swell suites, but significantly simplifies the implementation, and gives us other flexibility down the line.

src/swell/utilities/suite_utils.py
```diff
@@ -10,27 +10,47 @@
...
+        suite_configs = getattr(suite_module, 'SuiteConfig')
+
+        [suite_sub_list.append(suite_config[1:] if suite_config[0] == '_' else suite_config)
+         for suite_config in suite_configs.get_all()]
+
+        suite_config_list.extend(sorted(suite_sub_list))
 
     # List all directories in platform_directory
-    return suites
+    return suite_config_list
```
ashiklom, line 53: If we take the `all_suites` approach I described earlier, we can basically eliminate all of this logic.

src/swell/utilities/swell_questions.py
```diff
@@ -0,0 +1,195 @@
...
+class WidgetType(Enum):
+    STRING = "string"
+    STRING_CHECK_LIST = "string-check-list"
+    STRING_DROP_LIST = "string-drop-list"
+    BOOLEAN = "boolean"
+    ISO_DURATION = "iso-duration"
+    ISO_DATETIME = "iso-datetime"
+    INTEGER = "integer"
+    INTEGER_LIST = "integer-list"
+    FILE_CHECK_LIST = "file-check-list"
```
ashiklom, line 31: Missing `float` here?

src/swell/utilities/swell_questions.py
```diff
@@ -0,0 +1,195 @@
...
+        return self._member_names_
+
+# --------------------------------------------------------------------------------------------------
+
+
+@dataclass
+class QuestionList:
+    """Basic dataclass containing a list of questions for each model, suite, task"""
+    list_name: str
+    questions: list
```
ashiklom: Would love to be more precise here with, e.g.,

```python
questions: list[SwellQuestion]
```

...but this is throwing type errors for me and I can't (quickly) figure out why. If you know a quick solution, let's put the more precise type hint here. If not, we can punt till later.

ashiklom (changes requested):
Great work! This will be a massive design improvement once it's merged. I just have a few relatively minor things. The biggest ones are:

1. I found one possible bug that you should check related to indentation.

2. Consider my comments about using `importlib` to dynamically search for suite configs vs. just registering suites directly in Python objects. Unless there's a good reason, I'm inclined to do the latter, especially if it's relatively easy to implement.

Once those things are considered (not necessarily implemented), I'm happy to approve this so we can start reaping the benefits.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
ashiklom: I'm being a bit pedantic (read: annoying) here...but it's good practice to think through these things carefully, even when it doesn't matter much (which is the case here), so that we can be more robust when it does!

docs/adding_a_suite.md
mranst: > Also, why do we inherit from `Enum` here? (Not saying we shouldn't --- just curious to know the reason).

I think it started as me slightly misunderstanding what we should be using enums for, but there are some minor syntactical conveniences offered by enum. __getitem__  lets you get attributes similar to a dictionary, rather than using getattr (though you need to use .value to get the actual object), the _member_names_ attributes is also quite useful in this case

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
mranst: This is intentional - previously, the override dictionary would first be constructed from the tier1 test files, and then appended with the user override. Since I have removed the process using the tier1 files, the only time that block should run is if there is a user override

ashiklom: Perfect. Just confirming this wasn't a mistake.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
mranst: Interesting -  I'll keep that in mind. I've changed the usage for this method

src/swell/suites/3dfgat_atmos/suite_config.py
ashiklom: I took a stab at this in #520. Conceptually, I think it's a better approach. But, the need to rename all `3dvar`, etc. modules to `_3dvar` to work with Python's import mechanism seems very dangerous and has the potential to break a lot of things. So it needs more extensive testing.

In the meantime, I really want to see this get merged so we can adapt this approach moving forward (including for some of the changes in the currently open PRs).

So, migrating this to #521 and resolving this discussion.

src/swell/utilities/suite_utils.py
ashiklom: See #521.

src/swell/utilities/swell_questions.py
ashiklom: Migrating to #522 and closing here as out of scope for this PR.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
ashiklom: See #521. Out of scope for this PR.

src/swell/utilities/swell_questions.py
mranst: I believe this is fixed now, the issue was probably that questions takes both question and list objects

ashiklom: Nice! I could have sworn I tried the same thing, but I may have made a mistake somewhere. There are still some outstanding minor typing issues, but I want to punt on those in the interest of getting this merged.

mranst:
Thanks for the feedback - I'm working on merging the conflicts and cleaning a few things up

docs/adding_a_suite.md
```diff
@@ -0,0 +1,394 @@
+# A Guide to Suite Configurations in Swell
+
+# Introduction
+
+This document provides a high-level overview of the structure of a Swell suite, with the goal of providing context and basic instructions to a user looking to add a suite to Swell. It can also serve as a primer for users less familiar to Swell. 
+
+## Definitions
```
Dooruk, line 7: While I really appreciate the long descriptions here, I suggest we have these nuggets of information in a more conspicuous location. There are some reduntant parts in `getting started`part of the documentation and these bits and pieces here could replace those sections.

I'm not suggesting these changes for this PR, it could be part of the demonstration preparation while we update the documentation.

Dooruk: I'm not saying just the `Definitions` there is real useful information all over the `adding_a_suite.md` bit

docs/adding_a_suite.md
```diff
@@ -0,0 +1,394 @@
...
+
+An example question class:
+```python
+    @dataclass
+    class existing_jedi_build_directory(TaskQuestion):
+        default_value: str = "defer_to_platform"
+        question_name: str = "existing_jedi_build_directory"
+        ask_question: bool = True
+        depends: Dict = field(default_factory=lambda: {
+            "jedi_build_method": "use_existing"
```
Dooruk: Just a minor comment, since other developers might would have to use this, could we clarify the purpose of using lambda here somewhere?

docs/adding_a_suite.md
```diff
@@ -0,0 +1,394 @@
...
+In this question infrastructure, **suites take priority over tasks**. Any question specified in a suite configuration will override the default value for a question in one of its member tasks. This allows for easily setting different configurations for suites without having to specify redundant questions. For ease of use, model-dependent questions can be assigned directly in their respective lists.
+
+Consider the following example of suite questions for `3dvar` (in python, variable names cannot begin with digits):
+
+```python
+from swell.utilities.question_defaults import QuestionDefaults as qd
+from swell.suites.suite_questions import SuiteQuestions as sq
+
+from swell.utilities.question_defaults import QuestionDefaults as qd
+from swell.suites.suite_questions import SuiteQuestions as sq
```
Dooruk:
```suggestion
```
repetition

docs/adding_a_suite.md
ashiklom:
```suggestion
        # Need to do a lambda here because dataclass fields cannot be
        # initialized to mutable types (like dict or list).
        # https://docs.python.org/3/library/dataclasses.html#mutable-default-values
        depends: Dict = field(default_factory=lambda: {
            "jedi_build_method": "use_existing"
        })
```

docs/adding_a_suite.md
mranst: Yeah, I'll rethink the structure of this a bit when I take a look at the docs

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
mranst: > This approach feels more complicated and brittle than it needs to be. My suggestion would be to do the following:
> 
>     1. Define each suite directly as a python object, probably of a common class like `Suite`.
> 
>     2. Define an `Enum` called `all_suite_configs` or something that contains all the suites from (1).
> 
>     3. Here, just query that `all_suite_configs` object by name.

I came up with a different way of doing what I had before, using a dynamically created enum using a wrapper function. Might be bit hacky, and it still requires the use of importlib, so we may be better off going with manually entered enums. This was actually how an early prototype of the question lists in this PR worked until I decided it was too insane, but this is a better use of it, I think.

https://github.com/GEOS-ESM/swell/blob/2d5cf19e7e1abf60f10a18686ee8d6d5390429c9/src/swell/suites/all_suites.py

mranst: > > This approach feels more complicated and brittle than it needs to be. My suggestion would be to do the following:
> > ```
> > 1. Define each suite directly as a python object, probably of a common class like `Suite`.
> > 
> > 2. Define an `Enum` called `all_suite_configs` or something that contains all the suites from (1).
> > 
> > 3. Here, just query that `all_suite_configs` object by name.
> > ```
> 
> I came up with a different way of doing what I had before, using a dynamically created enum using a wrapper function. Might be bit hacky, and it still requires the use of importlib, so we may be better off going with manually entered enums. This was actually how an early prototype of the question lists in this PR worked until I decided it was too insane, but this is a better use of it, I think.
> 
> https://github.com/GEOS-ESM/swell/blob/2d5cf19e7e1abf60f10a18686ee8d6d5390429c9/src/swell/suites/all_suites.py

We can revert to a previous commit if it doesn't work

Dooruk:
@mranst, this is a really comprehensive update, great work! I do like the inheritance feature using the dictionaries and while YAMLs are more user friendly visually, dictionaries seem to be more versatile. After this PR we should have a release as SWELL v2. A few random notes:

- Thanks for adding an explanation for the lambda and as we talked yesterday we could expand the documentation using the sections you have put here.

- Since most of these changes are low level, should we ask for feedback from SI-team or other Task 8 folks? I didn't look at that aspect.

- I compared `experiment.yaml` between two SWELL builds with and without this PR and they look identical (except the the order of keys were changed but oh well) ✅ 

- ⚠️ Just a heads up, I tried launching an experiment and all tasks failed with the same error:
Screenshot 2025-03-21 at 10 43 50 AM

- Another minor issue with the questions, **not caused by this PR**. Same questions gets asked three times. This happens without this PR too but figured this might indicate an underlying problem:

Screenshot 2025-03-21 at 10 12 43 AM

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -448,25 +478,23 @@ def ask_questions_and_configure_suite(self) -> Tuple[dict, dict]:
...
-                if any(elem in question_tasks for elem in all_tasks):
+                # Check whether the question is associated with any model dependent tasks
+                if any(question_key in self.questions_per_task[task]
+                       for task in self.all_model_dep_tasks):
 
                     # Ask the question
                     self.ask_a_question(self.question_dictionary_model_ind, question_key)
 
+        # 8. Build a list of tasks for each model component
+        # -------------------------------------------------
```
metdyn: the note `#8.` is found within `#9.1`

mranst: Good catch, thanks!

mranst:
Thanks! 

>     * Since most of these changes are low level, should we ask for feedback from SI-team or other Task 8 folks? I didn't look at that aspect.

Can't see how it would be a bad idea. I'm actually not sure who would be most familiar with this aspect of swell, if you or @ashiklom know who to ping on this

>     * I compared `experiment.yaml` between two SWELL builds with and without this PR and they look identical (except the the order of keys were changed but oh well) ✅
> 
>     * ⚠️ Just a heads up, I tried launching an experiment and all tasks failed with the same error:
> 
> 
> Screenshot 2025-03-21 at 10 43 50 AM

Oops... mistakenly removed a module

>     * Another minor issue with the questions, **not caused by this PR**. Same questions gets asked three times. This happens without this PR too but figured this might indicate an underlying problem:
> 
> 
> Screenshot 2025-03-21 at 10 12 43 AM

Fixed in latest commits, caused by improperly checking if the question had been asked already. I also fixed a bug with building 3dvar from the cli, which was present in the previous version as well, there may be other cases of it

ashiklom:
> Since most of these changes are low level, should we ask for feedback from SI-team or other Task 8 folks? I didn't look at that aspect.

I don't think this is really relevant to the SI team.

From Task 8, now that Callum is gone, the main person that may have input on this is @jardizzo. As new Swell developers, @sabin413 and @ftgoktas may also want to weigh in (especially, on the new documentation), though recognizing that this may not mean as much to them if they are not as experienced with Swell functionality or its codebase.

src/swell/__init__.py
```diff
@@ -9,4 +9,4 @@
 repo_directory = os.path.dirname(__file__)
 
 # Set the version for swell
-__version__ = '1.9.6'
+__version__ = '1.9.7'
```
ashiklom: This is a pretty major change (albeit, not a user-facing one), so this should probably be at least a minor version bump.

```suggestion
__version__ = '1.10.0'
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -575,11 +571,26 @@ def ask_a_question(
                 self.questions_dict[question_key] = qd['prompt']
             else:
                 self.experiment_dict['models'][model][question_key] = \
-                    self.config_client.get_answer(self.logger, question_key, qd)
+                    self.config_client.get_answer(self.logger, question_key, qd, model)
                 self.questions_dict[f'models.{model}.{question_key}'] = qd['prompt']
 
     # ----------------------------------------------------------------------------------------------
 
+    def question_not_been_asked(self, question_key, model):
```
ashiklom: Right?

```suggestion
    def question_not_been_asked(self, question_key: str, model: str) -> bool:
```

src/swell/utilities/question_defaults.py
```diff
@@ -1020,6 +1006,23 @@ class path_to_gsi_nc_diags(TaskQuestion):
...
+        default_value: str = None
+        question_name: str = "perhost"
+        ask_question: bool = True
+        options: List[bool] = field(default_factory=lambda: [
+            True,
+            False
+        ])
+        models: List[str] = field(default_factory=lambda: [
+            "geos_atmosphere"
+        ])
```
ashiklom: The type checking is a bit tricky here because Python dataclasses are mutable (unless created with `@dataclass(frozen=True)`), so type checkers like `pyright` will complain about subtyping (because the type of the parent class (e.g., `Optional[list]` for `models`) -- doesn't exactly match the type here (`List[str]`; no option for `None`). As is usually the case with type issues, this isn't just about types but captures a deeper design issue.

Note that subtyping is OK if our dataclasses are actually immutable (or quasi-immutable, via `frozen=True). I.e., A type checker will error on this:

```python
@dataclass
class A():
    x: int | str

@dataclass
class B(A):
    x: int
```

...but not on this:

```python
@dataclass(frozen=True)
class A():
    x: int | str

@dataclass(frozen=True)
class B(A):
    x: int
```

However, here, we would have to apply `frozen=True` on both the parent and all children. Conceptually, that may actually be what we want. But, that will also require more restructuring because we won't be able to modify any instances of any of these objects.

Another approach: Rather than marking `options` and `models` required, remove them from the `SwellQuestion` definition. Then, when we actually process these questions, use something like `models = getattr(some_question, "models", None)` or `options = getattr(some_question, "options", None)` to avoid issues with all the questions that are not model-specific and/or don't have options.

https://microsoft.github.io/pyright/#/type-concepts-advanced?id=inheritance-of-class-and-instance-variables

https://github.com/microsoft/pyright/issues/6564#issuecomment-1830284962

mranst: I have implemented the second approach. I understand why it's like that, but I sort of wish it was more convenient  to set different types for child classes

ashiklom (dismissed):
Other than changing `self` to `cls` in the class methods (as we discussed --- tiny thing), and assuming all tests pass after merge, I think this looks good. Great work!!

mranst:
> Other than changing `self` to `cls` in the class methods (as we discussed --- tiny thing), and assuming all tests pass after merge, I think this looks good. Great work!!

Thanks, I believe this is ready to go! FYI @Dooruk

Dooruk:
> > Other than changing `self` to `cls` in the class methods (as we discussed --- tiny thing), and assuming all tests pass after merge, I think this looks good. Great work!!
> 
> Thanks, I believe this is ready to go! FYI @Dooruk

Sure, did you have a chance to run Tier 1 tests? I started an action but I see some tests are failing:

https://github.com/GEOS-ESM/swell/actions/runs/14359508109

mranst:
> Sure, did you have a chance to run Tier 1 tests? I started an action but I see some tests are failing:
> 
> https://github.com/GEOS-ESM/swell/actions/runs/14359508109

I ran 3dvar successfully, apparently the only one that passes :/

mranst:
> > Sure, did you have a chance to run Tier 1 tests? I started an action but I see some tests are failing:
> > https://github.com/GEOS-ESM/swell/actions/runs/14359508109
> 
> I ran 3dvar successfully, apparently the only one that passes :/

They should pass now. This happened because I got rid of the marine models task question. I've made it so that all suite questions are visible in the task config now (which I actually thought was the case before). This will allow suite questions to be used for both flow.cylc templating and in the tasks themselves, eliminating the need to have duplicate task and suite questions.

Dooruk:
Still failing I'm afraid, would you like to run t1test options locally?

https://github.com/GEOS-ESM/swell/actions/runs/14388210115

mranst:
> Still failing I'm afraid, would you like to run t1test options locally?
> 
> https://github.com/GEOS-ESM/swell/actions/runs/14388210115

I've run the failing tests with the fix, fingers crossed it _should_ be good now.

Dooruk (approved):
All good! A new SWELL era 😎 

Waiting for unresolved comment to be resolved before merging.

src/swell/suites/3dfgat_atmos/suite_config.py
Dooruk: @mranst, this is shown as an unresolved comment and prevents me from merging. I say let's keep this in mind, by taking a note, and address it later but up to you if it's a simple change.

You can see unresolved conversations from the `Conversations` drop-down menu in the review section.

PR #515, opened 2025-03-14
LGETKF with x0050 bkg

metdyn:
## Description

The x0050 ebkg from @rtodling is used for the GETKF with linear observer. The `get_ensemble_geos_experiment.py` is provided by @Dooruk.

### the newly added functions include:
- `obs_thinning_rej_fraction` is introduced

####  command line:
  swell create localensembleda -o override.yaml -s slurmfile.yaml -p nccs_discover_sles15
  cylc 

perhost is used to alleviate memory usage, otherwise the JEDI code hangs. 
```
override.yaml
models:
  geos_atmosphere:
    perhost: 32
```

### How many obs can be assimilated on C90?
- [x] 27
- [] 32
   We excluded IASI.

### How many ens members can be run on C90 ?
Technically, I am still testing how many ensemble members I can run with C90 on discover
- [x] mem=16,  perhost=32, np=96
- [] mem=32,  not tested

in collaboration with @rtodling and @Dooruk

## Dependencies

## Impact

- [ ] repo
- [ ] ...


metdyn:
bring this PR off for improvements

metdyn:
Hi Ricardo, 

I am reopening this PR, which contains `perhost` option.  I am making sure the branch updates to my current work

src/swell/deployment/create_experiment.py
```diff
@@ -488,7 +488,15 @@ def prepare_cylc_suite_jinja2(
...
-        render_dictionary['scheduling'][slurm_task]['execution_time_limit'] = 'PT1H'
+        x = render_dictionary['scheduling'][slurm_task]['execution_time_limit']
+        x = x if x is not None else 'PT1H'
+        render_dictionary['scheduling'][slurm_task]['execution_time_limit'] = x
+
+        # perhost be eliminatd before render flow.cylc
+        for mod in ['all'] + model_components:
+            if mod in render_dictionary['scheduling'][slurm_task]['directives']:
+                x = render_dictionary['scheduling'][slurm_task]['directives'][mod]
+                x.pop('perhost', None)
```
ashiklom: See my comment on #517 . That is the better long-term solution.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -206,10 +205,14 @@ def execute(self) -> None:
         # Compute number of processors
         # ----------------------------
         np = eval(str(model_component_meta['total_processors']))
+        slurm_dict = self.slurm_dict()
+        classname = self.__class__.__name__
+        m = self.get_model()
+        sub_dict = slurm_dict.get(classname, {}).get(m, {}) or slurm_dict.get(classname, {}).get('all', {})
+        perhost = sub_dict.get('perhost', None)
```
ashiklom: If we implement #517 then all of this extra `slurm_dict` stuff becomes unnecessary and we can just grab the `perhost` value directly from `self.config.perhost()`.

metdyn: Hi @ashiklom,  I am trying to figure out if I can shorten this 5 lines into 1 or 2. 
Meanwhile, I do oppose mingling slurfile.yaml  with model variables in config, where the discussion you have in https://github.com/GEOS-ESM/swell/pull/517 

The details from my view is:

Adding `self.config.perhost()` appears to be the most simple way, very quick and it will work for one task, but the consequences are:

case-1.  if we add to  perhost=32 to `override.yaml`, it will generate `experiment.yaml`,  then all the tasks:  3dvar, hofx, LocalEnsembleDa are forced to use the same perhost, which is not right.

case-2. if we customize to each task, one would have to add `perhost` literally to each file in `swell/test/suite_tests`, if he will work on both 3dvar and ensemble and hofx.

`3dfgat_atmos-tier1.yaml  3dvar_atmos-tier1.yaml  3dvar-tier1.yaml       convert_ncdiags-tier1.yaml  geosadas-tier1.yaml  localensembleda-tier1.yaml
3dfgat_cycle-tier1.yaml  3dvar_cycle-tier1.yaml  build_jedi-tier1.yaml  forecast_geos-tier1.yaml    hofx-tier1.yaml      `

For debug, we have to frequently change perhost with different ensemble members or the model resolution C90/C180 etc.  It quickly become inconvenient to go into the source code

case-3:  we put in the slurmfile.yaml, where the user has a full control under their runtime directory. I am trying to see how to get rid of the above 5 lines by using self.config.perhost(). [not sure yet].

metdyn: Another benefit from here is we donot need to register `perhost` to the task_question.yaml, due to separating model config from slurmfile.yaml

src/swell/deployment/create_experiment.py
metdyn: Like @rtodling, I merely want the both `perhost` and `execution_time_limit` to simply work with minimal user intervention. My modification will do this, although software design and consistency is at another level.

metdyn:
work offline to add perhost to config

metdyn:
This PR is ready for review.  I moved `perhost` to config.  
With 16 mem and 27 observations, after 1 hr, the run is at crtm:

ufo_radiancecrtm_simobs: // ObsDiagnostic is unsupported, emissivityJacobian_1
ufo_radiancecrtm_simobs: // ObsDiagnostic is unsupported, emissivityJacobian_2

I have small left over problems with pre-steps of computing mean and variance, which will be added in another PR.

src/swell/suites/localensembleda/flow.cylc
```diff
@@ -213,7 +211,8 @@
 #        {%- endfor %}
 
     [[SaveObsDiags-{{model_component}}]]
-        script = "swell task SaveObsDiags $config -d $datetime -m {{model_component}}"
+        script = true
+#        script = "swell task SaveObsDiags $config -d $datetime -m {{model_component}}"
```
ashiklom: Am I understanding correctly that this effectively disables the `SaveObsDiags` module, because the module "runs" but doesn't actually do anything?

If so, this feels brittle and likely to cause confusion. If we're going to skip this task, we should properly remove it.

At an absolute minimum, instead of using the uninformative value `true`, we should do something like this (replacing `###` with a Swell GitHub issue to re-enable this). Since the `echo` command should silently succeed, this accomplishes the same result but in a more informative way.

```
script = "echo 'SaveObsDiags is disabled for this task. See GH issue ###'"
```

src/swell/tasks/get_ensemble_geos_experiment.py
```diff
@@ -0,0 +1,131 @@
...
+        6-hour window.
+
+        """
+
+        # Parse config
+        # ------------
+        background_experiment = self.config.background_experiment()
+        geos_x_ensemble_directory = self.config.geos_x_ensemble_directory()
+        # background_time_offset = self.config.background_time_offset()
+        background_time_offset = 'PT3H'
```
ashiklom: This needs to be a properly configured default value.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -227,12 +226,16 @@ def execute(self) -> None:
...
-                self.logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
                 run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
-                               jedi_config_file, output_log_file)
+                               jedi_config_file, output_log_file, perhost=perhost)
         else:
+            mpi_command = "mpirun"
+            if not (perhost is None or perhost == "None"):
+                mpi_command += f" -perhost {perhost}"
+            mpi_command += f" -np {np} {jedi_executable_path} {jedi_config_file} {output_log_file}"
+            print(f'intended mpi_command = {mpi_command}')
```
ashiklom, line 238: Remove debug by print. If needed, this should be a `self.logger.debug(...)`.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -227,12 +226,16 @@ def execute(self) -> None:
...
-                               jedi_config_file, output_log_file)
+                               jedi_config_file, output_log_file, perhost=perhost)
             else:
-                self.logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
                 run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
-                               jedi_config_file, output_log_file)
+                               jedi_config_file, output_log_file, perhost=perhost)
         else:
+            mpi_command = "mpirun"
+            if not (perhost is None or perhost == "None"):
```
ashiklom, line 235: `perhost` is configured as an integer or `None`. It should never be a string, so this check is unnecessary.

```suggestion
            if perhost is not None:
```

src/swell/tasks/run_jedi_obsfilters_executable.py
```diff
@@ -130,7 +126,7 @@ def execute(self, ensemble_members: Optional[list] = None) -> None:
             name, extension = os.path.splitext(filename)
             new_filename = f"{name}_orig{extension}"
             new_obsfile_in = '/'.join(elements[:-1])+'/'+new_filename
-            copy(obsfile, new_obsfile_in)
+            shutil.move(obsfile, new_obsfile_in)
```
ashiklom, line 124: This is an important side-effect that deletes the original contents of `obsfile`. From a quick search, it looks like the `obsfile`s are also accessed in `save_obs_diags.py` and `eva_observations.py`.

I'm not super versed in how those things work or interact with this code. _If you're sure_ that this won't break those other modules, we can just resolve this conversation (though a comment indicating why this is OK could be useful).

But if preserving those other locations is important, we should change this back to `copy`.

src/swell/tasks/run_jedi_obsfilters_executable.py
```diff
@@ -155,16 +151,13 @@ def execute(self, ensemble_members: Optional[list] = None) -> None:
         jedi_executable_path = os.path.join(self.experiment_path(), 'jedi_bundle',
                                             'build', 'bin', jedi_executable)
 
+        np = 1
```
ashiklom, line 155: Suggest instead defining it next to `run_executable` and with a clearer variable name; e.g.,

```python
            NPROCS = 1
            self.logger.info('Running '+jedi_executable_path+' with '+str(NPROCS)+' processors.')
            run_executable(self.logger, self.cycle_dir(), NPROCS, jedi_executable_path,
                           jedi_config_file, output_log_file)
```

src/swell/utilities/run_jedi_executables.py
```diff
@@ -152,10 +151,10 @@ def run_executable(
             jedi_executable_path,
             jedi_config_file
         ]
+    print('mpi_command='+" ".join(command)+' '+output_log)
```
ashiklom: Stray debug by print. Either delete or use `self.logger.debug(...)`.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: It is perplexing to me as well: 
`if perhost is not None:` does not work at here, only
`if not (perhost is None or perhost == "None"):`  works. 
So there is a bug somewhere. 
I found this line from @rtodling's PR for perhost, and it worked here, the other way fails.

Dooruk: Yeah, this was my "quick fix" because @rtodling's PR failed.

ashiklom: OK -- let's keep it as is for now, and I'll open an issue.

src/swell/tasks/get_ensemble_geos_experiment.py
metdyn: Thanks!  should have recovered it (I did it now), but it is also a reminder we need to merge the PR:
- https://github.com/GEOS-ESM/swell/pull/519

metdyn: Wait, I think it might be from @Dooruk, and it works well, but if it should be the default, I am not sure.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: good point,  but my intent is not to read it but to copy and paste this command for me to debug the code.  Hence, self.logger.info(...) will look nice below, but not what I want.
```
RunJediLocalEnsembleDaExecutable: intended mpi_command = mpirun -perhost 32 -np 96 /discover/nobackup/projects/gmao/aist- ...
 RunJediLocalEnsembleDaExecutable:  nr/yyu11/br_jedi_scu17/run_swell/gkf5/jedi_bundle/build/bin/fv3jedi_letkf.x ...
 RunJediLocalEnsembleDaExecutable:  /discover/nobackup/projects/gmao/aist-nr/yyu11/br_jedi_scu17/run_swell/gkf5/run/20231010T000000Z/geo ...
 RunJediLocalEnsembleDaExecutable:  s_atmosphere/jedi_localensembleda_config.yaml /discover/nobackup/projects/gmao/aist-nr/yyu11/br_jedi ...
 RunJediLocalEnsembleDaExecutable:  _scu17/run_swell/gkf5/run/20231010T000000Z/geos_atmosphere/jedi_localensembleda_log.log
```

src/swell/tasks/run_jedi_obsfilters_executable.py
metdyn: Thanks for detailed work. You are right about the later save and eva. Nonetheless, this obsfile has to be renamed and moved away, because the jedi code wants to write into this filename, and writing HDF5 fails if the filename exists. It is a bug in IODA. I wrote the previous code and it leads to problems when writing tasks individually.

src/swell/tasks/run_jedi_obsfilters_executable.py
metdyn: Let's keep it light and replace  `np = 1` with `np = 1   # np: NPROCS`.  The other way feels over emphasizing this 1-processor mpi run.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: I think this is a bug assigning None to an integer,  GPT wrote
```
x = 10
x = None  # Allowed, but x is no longer an integer
```

With the code from previous PR:

https://github.com/GEOS-ESM/swell/blob/39f35f8179a51d3386c29f59619c09dfae56f6bc/src/swell/tasks/task_questions.yaml#L833-L841

The `perhost` default is a "None" string, which costed me an hour to wonder why the `if perhost is not None` does not work here.   We should give default 0 for perhost.

ashiklom: The issue here is that `yaml` doesn't know to treat the string `None` as a special value; it just reads it as a string, `"None"`. The best solution is to replace `default_value: None` with `default_value: null` (or, less obviously, a blank entry -- `default_value: `; or even less obviously, a `~`, which yaml interprets as empty -- `default_value: ~`), which Python will read in as a `None` object.

```python
import yaml

doc1 = """
a: 1
b:
    b1: null
    b2: yes
"""

d1 = yaml.safe_load(doc1)
d1["b"]["b1"] is None
# True

doc2 = """
a: 1
b:
    b1:
    b2: yes
"""

d2 = yaml.safe_load(doc2)
d2["b"]["b1"] is None
# True


doc3 = """
a: 1
b:
    b1: None
    b2: yes
"""
d3 = yaml.safe_load(doc3)
d3["b"]["b1"] is None
# False -- returns the string "None"
```

The _bigger_ issue is that we don't have a good way of catching these kinds of parsing/type errors at runtime (and it's even tricky to do with static type checking). Which is all the more reason I'm looking forward to #514 --- once these things are Python objects, we pick up a lot more power from static and runtime checking.

ashiklom: In any case, this isn't your bug to fix; we can resolve it once this is merged.

src/swell/tasks/get_ensemble_geos_experiment.py
Dooruk:
```suggestion
        background_time_offset = self.config.background_time_offset()
```

should work, as long as `task_questions.yaml` is set for `GetEnsembleGeosExperiment` to access `background_time_offset`.

src/swell/suites/localensembleda/flow.cylc
Dooruk: I agree, if you want to skip this task for now you can comment out the script  in [runtime] and [scheduling].

This task for other suites otherwise.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.yaml
```diff
@@ -5,7 +5,7 @@ members from template:
     filetype: cube sphere history
     provider: geos
     datapath: '.'
-    filenames: ['geos.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4',
+    filenames: ['ebkg/mem%mem%/geos.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4',
```
Dooruk, line 8: This is just a comment but do we want to keep `mem%` in the folder name as well as in the filename, asking @rtodling?

src/swell/utilities/shell_commands.py
```diff
@@ -89,8 +88,8 @@ def run_subprocess(
         subprocess.run(command, check=True, stdout=stdout, stderr=stderr, **kwargs)
     except subprocess.CalledProcessError as e:
         print(e)
```
Dooruk:
```suggestion
```
this is obsolete

src/swell/tasks/run_jedi_obsfilters_executable.py
```diff
@@ -114,7 +110,7 @@ def execute(self, ensemble_members: Optional[list] = None) -> None:
 
         # Filter Thinning
         # ----------------------
-        filter_thinning = [{'filter': 'Thinning', 'amount': 0.75,
+        filter_thinning = [{'filter': 'Thinning', 'amount': obs_thinning_rej_fraction,
                             'random seed': 0, 'member': 1,
```
Dooruk: Just curious, will this random seed number or member number change?

Dooruk (changes requested):
Some minor changes but almost ready.

src/swell/tasks/get_ensemble_geos_experiment.py
metdyn: @Dooruk: It works when overriding background_time_offset in localensembleda-tier1.yaml. The default value is PT9H and the setup varies in 3dvar-tier1.yam vs 3dfgat_cycle-tier1.yaml vs 3dvar_cycle-tier1.yaml.

src/swell/tasks/run_jedi_obsfilters_executable.py
metdyn: We can see its usage from the source code copied below from ufo/src/ufo/filters/Thinning.cc, which involves both random seed and member!   This means I can assign 'random seed' with a random number for each ob for a better randomness! 
Thank you for a very good question! 

```
  // create random numbers for each observation based on some seed
  unsigned int random_seed = parameters_.randomSeed.value().value_or(std::time(0));
  random_seed += parameters_.member;

  util::UniformDistribution<float> rand(max_gindex, 0.0, 1.0, random_seed);
```

metdyn: solved, setting random seed ranging from 1 to 100.

Dooruk: Awesome 👍

src/swell/suites/localensembleda/flow.cylc
Dooruk: I think this is resolved now?

Dooruk:
Looks like a few changes left and there are some unresolved comments from @ashiklom, otherwise, @metdyn this is good to go in on my end.

src/swell/suites/localensembleda/flow.cylc
metdyn: Right, that is now resolved.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: This issue is resolved and perhost is set in the config.

metdyn:
BTW, @Dooruk, these modifications are meant for the ETKF (localEnsemble workflow) and it should not affect others' work.  @rtodling will check that and we will have more iterations I believe later on. If those further issues do occur, we can quickly fix it.  All the current issues are resolved.

Dooruk:
> BTW, @Dooruk, these modifications are meant for the ETKF (localEnsemble workflow) and it should not affect others' work. @rtodling will check that and we will have more iterations I believe later on. If those further issues do occur, we can quickly fix it. All the current issues are resolved.

Sounds good, if you merge this with develop I can test tier1 tests and merge.

Dooruk (approved):

Dooruk:
Hmm it says there is a comment left but I can't see it and doesn't let me merge without resolving that comment.

metdyn:
> Some minor changes but almost ready.

I am not sure if this is the comment that I need to reply to let the PR merge

PR #517, opened 2025-03-18
Add perhost opt for mpi execs

rtodling:
## Description

Some mpirun execs require a perhost option in the mpirun command line. 

This works now.

## Dependencies

None

## Impact

Allows for memory handling when job needs nodes need to be split and info passed to mpirun command line

### DIFF
--- a/src/swell/tasks/run_jedi_variational_executable.py
+++ b/src/swell/tasks/run_jedi_variational_executable.py
@@ -163,13 +163,8 @@ class RunJediVariationalExecutable(taskBase):
         # -----------------------
         if not generate_yaml_and_exit:
             self.logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
-            if perhost is None:
-               run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
-                              jedi_config_file, output_log_file)
-            else:
-               perhost = eval(perhost)
-               run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
-                              jedi_config_file, output_log_file, perhost)
+            run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
+                           jedi_config_file, output_log_file, perhost)
         else:
             self.logger.info('YAML generated, now exiting.')
 
--- a/src/swell/tasks/run_jedi_variational_executable.py
+++ b/src/swell/tasks/run_jedi_variational_executable.py
@@ -38,6 +38,7 @@ class RunJediVariationalExecutable(taskBase):
         observations = self.config.observations()
         jedi_forecast_model = self.config.jedi_forecast_model(None)
         generate_yaml_and_exit = self.config.generate_yaml_and_exit(False)
+        perhost = self.config.perhost(None)
 
         # Set the observing system records path
         self.jedi_rendering.set_obs_records_path(self.config.observing_system_records_path(None))
@@ -81,7 +82,6 @@ class RunJediVariationalExecutable(taskBase):
         self.jedi_rendering.add_key('npx_proc', npx_proc)
         self.jedi_rendering.add_key('npy_proc', npy_proc)
         self.jedi_rendering.add_key('total_processors', self.config.total_processors(None))
-        self.jedi_rendering.add_key('perhost', self.config.perhost(None))
 
         # Observations
         # ------------
@@ -152,7 +152,6 @@ class RunJediVariationalExecutable(taskBase):
         # Compute number of processors
         # ----------------------------
         np = eval(str(model_component_meta['total_processors']))
-        perhost = str(model_component_meta['perhost'])
 
         # Jedi executable name
         # --------------------
--- a/src/swell/tasks/run_jedi_variational_executable.py
+++ b/src/swell/tasks/run_jedi_variational_executable.py
@@ -81,6 +81,7 @@ class RunJediVariationalExecutable(taskBase):
         self.jedi_rendering.add_key('npx_proc', npx_proc)
         self.jedi_rendering.add_key('npy_proc', npy_proc)
         self.jedi_rendering.add_key('total_processors', self.config.total_processors(None))
+        self.jedi_rendering.add_key('perhost', self.config.perhost(None))
 
         # Observations
         # ------------
@@ -151,6 +152,7 @@ class RunJediVariationalExecutable(taskBase):
         # Compute number of processors
         # ----------------------------
         np = eval(str(model_component_meta['total_processors']))
+        perhost = str(model_component_meta['perhost'])
 
         # Jedi executable name
         # --------------------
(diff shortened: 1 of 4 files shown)


metdyn:
Hi Ricardo, 

We double worked on this perhost.
I have a better solution to add it to slurmfile.yaml.

```
slurm_directives_tasks:
    execution_time_limit: 'PT2H30M'
    geos_atmosphere:
      nodes: 3
      ntasks-per-node: 126
      perhost: 32

  RunJediObsfiltersExecutable:
    execution_time_limit: 'PT5M'
    all:
      nodes: 1
      ntasks-per-node: 1

```

Reading from config has the drawback of separating `nodes` and `execution_time_limit` from perhost.

rtodling:
> Hi Ricardo,
> 
> We double worked on this perhost. I have a better solution to add it to slurmfile.yaml.
> 
> ```
> slurm_directives_tasks:
>   RunJediLocalEnsembleDaExecutable:
>     execution_time_limit: 'PT2H30M'
>     geos_atmosphere:
>       nodes: 3
>       ntasks-per-node: 126
>       perhost: 32
> 
>   RunJediObsfiltersExecutable:
>     execution_time_limit: 'PT5M'
>     all:
>       nodes: 1
>       ntasks-per-node: 1
> ```
> 
> Reading from config has the drawback of separating `nodes` and `execution_time_limit` from perhost.

I don't think per host is a slurm option ... it's something that must be placed in the mpirun command line ... how's what you suggest  going to fall in the command line ?

metdyn:
I modified the `slurm option` to add this `perhost` keywords, so that it will not fail. 
I modified the same way as you did for `command=[ mpirun, etc.]`
My modification reads in the `slurm_directives_tasks`  from utilities/slurm.py and systematically call it from suite. I also fixed the 'execution_time_limit' problem. 
My suggestions is please you review my PR for all this.  I will make your code to work as soon as possible.

rtodling:
> I modified the `slurm option` to add this `perhost` keywords, so that it will not fail. I modified the same way as you did for `command=[ mpirun, etc.]` My modification reads in the `slurm_directives_tasks` from utilities/slurm.py and systematically call it from suite. I also fixed the 'execution_time_limit' problem. My suggestions is please you review my PR for all this. I will make your code to work as soon as possible.

Let me call you in a few minutes ...

Dooruk:
I'm not familiar with the `perhost` option I'm afraid. I understand it will be platform specific, i.e., Cascade vs. Milan compute nodes. If it's also going to be model resolution specific we could craft this more carefully.

ashiklom:
I think @rtodling 's solution here is the more correct long-term solution than @metdyn 's solution of sticking this into the `slurmfile`. Similar to my comments on #518 , `slurmfile` should be reserved strictly for slurm directives (and even the existence of `slurmfile` itself is a hack; we should really be configuring everything through the `overrides.yaml` or questionary interface).

That said, the fact that it's easier to put things into the `slurmfile` rather than add them as new Swell configuration options is a symptom of a major fundamental design flaw in Swell. @rtodling is completely correct that this needs to be _much_ simpler to do. I'm hoping that #514 will be a step in the right direction towards making all of this simpler and cleaner.

src/swell/tasks/run_jedi_variational_executable.py
```diff
@@ -162,8 +163,13 @@ def dict_to_ordereddict(d):
...
             self.logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
-            run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
-                           jedi_config_file, output_log_file)
+            if perhost is None:
+               run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
+                              jedi_config_file, output_log_file)
+            else:
+               perhost = eval(perhost)
+               run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
+                              jedi_config_file, output_log_file, perhost)
```
ashiklom: Your `run_executable` already handles the case of `perhost = None`, so you can simplify this:

```suggestion
            run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
                           jedi_config_file, output_log_file, perhost)
```

(This also fixes a minor Python style error that `pycodestyle` was complaining about.

src/swell/utilities/run_jedi_executables.py
```diff
@@ -112,14 +112,19 @@ def run_executable(
     np: int,
     jedi_executable_path: str,
     jedi_config_file: str,
-    output_log: str
+    output_log: str,
+    perhost: int=None
```
ashiklom: To satisfy the type checker (otherwise, it complains that `None` cannot be coerced to type `int`).

```suggestion
    perhost: Optional[int] = None
```

(This also fixes a pycodestyle complaint).

rtodling:
I agree that the slurm.yaml should control only slurm directives ... so I do prefer my way of doing this; however, my way is not task specific - which what @metdyn did is - so there is good and bad in both ... this discussion can take time ...

What I suggest is that we pick one or the other for now. And then we let the software engineers thing on a way to accommodate a generalization. 

Since we know we don't what the slurm.yaml to incorporate this change - w/ all due respect to @metdyn - how about we approve my change and @ashiklom and @ashiklom work on further, knowing that we would prefer the change to apply to individual tasks.

src/swell/utilities/run_jedi_executables.py
```diff
@@ -112,14 +112,19 @@ def run_executable(
...
     # -----------------------
-    logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
 
-    command = ['mpirun', '-np', str(np), jedi_executable_path, jedi_config_file]
+    if perhost is None:
+       logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors.')
+       command = ['mpirun', '-np', str(np), jedi_executable_path, jedi_config_file]
+    else:
+       logger.info('Running '+jedi_executable_path+' with '+str(np)+' processors & perhost '+str(perhost))
+       command = ['mpirun', '-np', str(np), '-perhost', str(perhost), jedi_executable_path, jedi_config_file]
```
ashiklom: This fixes `pycodestyle` complaints about indentation and line length.

(Also, I'm personally partial to the f-string syntax -- `f"the value of x is {x}"` --- over string concatenation --- `"the value of x is "+x` --- but to each their own...).

```suggestion
    if perhost is None:
        logger.info(f"Running {jedi_executable_path} with {str(np)} processors.")
        command = [
            'mpirun',
            '-np', str(np),
            jedi_executable_path,
            jedi_config_file
        ]
    else:
        logger.info(
            f"Running {jedi_executable_path} with {str(np)} processors & perhost {str(perhost)}"
        )
        command = [
            'mpirun',
            '-np', str(np),
            '-perhost', str(perhost),
            jedi_executable_path,
            jedi_config_file
        ]
```

ashiklom:
@rtodling I am happy to approve this once you:

1. Click "Ready for review" (removing this from "draft" status).
2. Apply my suggested fixes (these address `pycodestyle` complaints that are currently causing the tests to fail; they also slightly simplify the implementation). You should just be able to do it from GitHub, via the "commit suggestion" button.

src/swell/utilities/run_jedi_executables.py
rtodling: @ashiklom can you explain to me what was the problem with/ the original set of lines here? What you have is identical to what I have except that the command is split into multiple lines!

ashiklom: Our "python coding norms" automated test was failing because the lines were too long (>100 characters). So I split the lines so our checker approves this.

https://github.com/GEOS-ESM/swell/actions/runs/13970973305/job/39112388175#step:6:15

rtodling:
@ashiklom Thanks for the explanations; I was definitely not using the norms and test checking tools - sorry - @mranst straightened me up.

Dooruk (approved):

Dooruk:
Well, this broke `3dfgat_atmos` and `3dvar_atmos` suites. @rtodling did you test your changes with `_atmos` suites?

https://github.com/GEOS-ESM/swell/actions/runs/13978722143

Screenshot 2025-03-20 at 4 59 58 PM

my bad, I wanted to merge it quickly 🥲

rtodling:
Hum - I did test it ... let me check on it again!

Issue #518, opened 2025-03-20
Read in execution_time_limit from slurmfile


metdyn:
The current code did not read in ` execution_time_limit` for each slurm_task, such as LocalEnsembleDa, 3DVar, etc.  Instead, this time limit is hard coded to PT1H from the source.  Problems arises when a job script such as`RunJediLocalEnsembleDaExecutable` which is managed by cylc cannot finish in one hour, then the slurm will stop the job script. 

https://github.com/GEOS-ESM/swell/blob/848e7acd608f9f8f4b915da3fe6ad3ec35882a98/src/swell/deployment/create_experiment.py#L489-L491

Therefore we need to allow for customizable `execution_time_limit ` for each slurm_task, which can be read in from slurmfile. 
A small remedy will be provided for this exercise.

PR #519, opened 2025-03-20
Read in execution_time_limit from slurmfile.yaml

metdyn:
## Description

Now the code can read the `execution_time_limit` setup from slurmfile, solving issue https://github.com/GEOS-ESM/swell/issues/518

Input:  slurmfile.yaml
slurm_directives_tasks:
  RunJediVariationalExecutable:
    execution_time_limit: '**PT29M**'
    all:
      nodes: 1
      ntasks-per-node: 120

swell create 3dvar_atmos -o override.yaml -s slurmfile.yaml -p nccs_discover_sles15

Output:
3dvar-suite:
flow.cylc
    [[RunJediVariationalExecutable-geos_atmosphere]]
        script = "swell task RunJediVariationalExecutable $config -d $datetime -m geos_atmosphere"
        platform = nccs_discover_sles15
        execution time limit = **PT29M**


ashiklom:
We _really_ shouldn't be using `slurmfile` this way because this has nothing to do with Slurm. What we really need is a way to parse this from the experiment configuration (i.e., the overrides file, or questionary interface).

That said, if this is a blocking issue and there is no other easy solution, I may be OK with this a short term fix, **as long as we recognize that the interface for this will change in a future PR** (i.e., this configuration will be moved from `slurmfile` into the experiment configuration in the overrides).

src/swell/utilities/slurm.py
```diff
@@ -184,6 +184,11 @@ def prepare_scheduling_dict(
             validate_directives(model_directives)
             scheduling_dict[slurm_task]["directives"][model_component] = model_directives
 
+        x = None
+        if slurm_task in experiment_task_directives.keys():
+            x = experiment_task_directives[slurm_task].get('execution_time_limit', None)
+        scheduling_dict[slurm_task]['execution_time_limit'] = x
+
```
ashiklom: This logic seems overly convoluted. I suggest just doing this and eliminating the change in `create_experiment.py`.

```suggestion
        if slurm_task in experiment_task_directives.keys():
            scheduling_dict[slurm_task]["execution_time_limit"] = (
                experiment_task_directives[slurm_task].get("execution_time_limit", "PT1H")
            )
```

The `dict.get(key, default_value)` method will grab `dict[key]` if `key` is in `dict` or, if `key` is not in `dict`, will return `default_value` instead.

src/swell/deployment/create_experiment.py
```diff
@@ -488,7 +488,9 @@ def prepare_cylc_suite_jinja2(
 
     # Default execution time limit for everthing is PT1H
     for slurm_task in render_dictionary['scheduling'].keys():
-        render_dictionary['scheduling'][slurm_task]['execution_time_limit'] = 'PT1H'
+        x = render_dictionary['scheduling'][slurm_task].get('execution_time_limit', None)
+        x = x if x is not None else 'PT1H'
+        render_dictionary['scheduling'][slurm_task]['execution_time_limit'] = x
```
ashiklom: Per my other comment, suggest instead removing this entire block and keeping this logic in `slurm.py`.

src/swell/utilities/slurm.py
metdyn: Good suggestion, and the modification in `create_experiment.py` is consolidated to here!

Dooruk:
@metdyn Could you update this branch with develop?

> That said, if this is a blocking issue and there is no other easy solution, I may be OK with this a short term fix, **as long as we recognize that the interface for this will change in a future PR** (i.e., this configuration will be moved from `slurmfile` into the experiment configuration in the overrides).

@ashiklom, could we address this later properly but merge for now? I will open an issue. There are some major changes @metdyn will introduce and this is holding him back.

src/swell/utilities/slurm.py
```diff
@@ -184,6 +184,12 @@ def prepare_scheduling_dict(
             validate_directives(model_directives)
             scheduling_dict[slurm_task]["directives"][model_component] = model_directives
 
+        # Default execution time limit for everthing is PT1H
+        x = 'PT1H'
+        if slurm_task in experiment_task_directives.keys():
+            x = experiment_task_directives[slurm_task].get('execution_time_limit', x)
+        scheduling_dict[slurm_task]['execution_time_limit'] = x
```
ashiklom, line 183: Simplify the implementation.

```suggestion
        # Default execution time limit for everthing is PT1H
        if slurm_task in experiment_task_directives.keys():
            scheduling_dict[slurm_task]["execution_time_limit"] = (
                experiment_task_directives[slurm_task].get("execution_time_limit", "PT1H")
            )
```

ashiklom:
@Dooruk as long as you, @metdyn, and anyone else using this are prepared to accept changes to the interface in the (near) future, sure, I'm happy to approve this (once my simplification above is incorporated).

Dooruk:
@metdyn please address @ashiklom's changes and then we can merge this.

src/swell/utilities/slurm.py
metdyn: Hi @ashiklom: 

I kind of missed your message. 
Your modification will miss:
`global set for execution_time_limit` or an `else` statement
because 
L187 -- L191 is inside a do loop and scheduling_dict[slurm_task]["execution_time_limit"] was not previously defined:
for slurm_task in slurm_tasks:
    ...
    ...
    if slurm_task in experiment_task_directives.keys():
            scheduling_dict[slurm_task]["execution_time_limit"] = (
                experiment_task_directives[slurm_task].get("execution_time_limit", "PT1H")
   else:
           scheduling_dict[slurm_task]["execution_time_limit"] = "PT1H"
            without else:  the above execution_time_limit is undefined.
    endif 
end do

These are subtle details, and I donot intend to focus on this and I was trying to preserve the logic in the old code.  Code refactoring to create a clean logic and easy to read manuscript will be great and I am on your side!

ashiklom: Ah, I missed that, sorry! We can leave as is.

Dooruk (approved):
Ok, I'm running the tier1 tests, if they pass I will merge this.

PR #520, opened 2025-03-20
Create all_suites object with all suites

ashiklom:
This avoids the need for complicated `importlib`-with-interpolated-string machinations to get swell suites. Instead, we just keep all the suites in a single Python enum for easy access.

NOTE: This also required putting underscores in front of modules with names. **That may break things.**

Given that the underscore stuff has the potential to break a lot of things, it may be a good idea to punt on this, merge #514 more-or-less as is (once minor issues are addressed), and then revisit this later.


ashiklom:
On second thought, this is too much change, all at once, with insufficient testing. So, closing this, but now we have a PR to which we can refer back in the future.

Issue #513, opened 2025-02-19
`-P1` does not work in the UTC mode


Dooruk:
With Cylc, it is possible to create dependencies to a previous cycle task before triggering a task which is a key DA workflow functionality.

With UTC mode, it is not possible to use -P1, so the time needs to be explicitly defined as such: 

https://github.com/GEOS-ESM/swell/blob/f66707607be64580504e00f8968eaf0c3d5a230a/src/swell/suites/3dvar_cycle/flow.cylc#L70-L71

We could convert `window_length` and use in `flow.cylc` for this entry.

This is required before we could cycle with longer window lengths, or else needs to be changed manually.

Dooruk:
Relevant issue:

https://github.com/cylc/cylc-flow/issues/6739

Dooruk:
Could I ask this as a priority? Or is it addressed with the new approach?

mranst:
> Could I ask this as a priority? Or is it addressed with the new approach?

Sure, I can take a look. Can you confirm I'm understanding this correctly? - The problem is that -P1 refers to the previous cycle (regardless of the window length), but this does not work in UTC mode because it can refer only to set time windows, so we need some way of replicating this by dynamically referring to UTC times using increments of window length?

PR #523, opened 2025-03-26
Allow assimilation of Jason-3N

DavidSRussell:
## Description

This allows assimilation of Jason-3N altimetry. By the currently-used testing period of late 2023, Jason-3 had entered a new orbit and is being treated as a different observation type from Jason-3 at GMAO, so Swell might as well do the same.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_marine/observations/adt_jason3n.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_marine/observations/adt_jason3n.yaml
@@ -1,0 +1,34 @@
+obs space:
+  name: adt_jason3n
+  obsdatain:
+    engine:
+      type: H5File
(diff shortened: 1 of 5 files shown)


Dooruk (approved):
Thanks @DavidSRussell, I hijacked this to add a fix for `RECORD_FREQUENCY` issue I mentioned, so you can cycle using P1D. Only change you need to address is:

https://github.com/GEOS-ESM/swell/issues/513

I'm working on another change regarding `MOM_oda_incupd`, where the IAU window for MOM6 is automatically set to half of DA window but I might merge that with another PR.

Dooruk:
closes https://github.com/GEOS-ESM/GEOSgcm_App/issues/697

Eh, I guess i can't link another repo's issue..

Dooruk:
This is good to go in, thanks again @DavidSRussell. Please try naming the branch `feature/` something, you can add your name too if you would like.

Maybe this helps somewhat why is it relevant?

Git_flow

PR #527, opened 2025-03-27
Use `null` instead of `None` in YAML

### LINKED ISSUE #524: Runtime checking when types for "none"
Example for `perhost`: This should be either an integer or `None`. It should _never_ be the string `"None"`. Swell needs to catch this error when parsing the config files.

----

              It is perplexing to me as well: 
`if perhost is not None:` does not work at here, only
`if not (perhost is None or perhost == "None"):`  works. 
So there is a bug somewhere. 
I found this line from @rtodling's PR for perhost, and it worked here, the other way fails.

_Originally posted by @metdyn in https://github.com/GEOS-ESM/swell/pull/515#discussion_r2016806160_

ashiklom:
PyYAML parses `value: None` as `{"value": "None"}` (i.e., the string `"None"`). In YAML, the correct way to represent a `None` object is one of the following:

```
value:

value: null

value: ~
```

I'm using `null` here because, in my opinion, it's the clearest one.

Closes #524.

### DIFF
--- a/src/swell/tasks/task_questions.yaml
+++ b/src/swell/tasks/task_questions.yaml
@@ -832,7 +832,7 @@ path_to_gsi_nc_diags:
 
 perhost:
   ask_question: true
-  default_value: None
+  default_value: null
   models:
   - geos_atmosphere
   prompt: What is the number of processors per host?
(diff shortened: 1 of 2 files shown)


metdyn (approved):
Nice and this helps!

Issue #490, opened 2024-12-20
(SE) Utilizing more Cylc features


Dooruk:
One of the main reasons we use Cylc is its design which includes many useful features for DA cycling. We currently exercise some of them but we should take advantage of more. I will create this as an epic and as users think of more features they can add this can expand.

## Necessary ones:

- [ ] **Execute certain tasks in TSE staging area:** This would mainly be relevant for GEOS simulations but if it provides significant efficiency for others we could benefit from it.
- [ ] **Use compiler flags for different environments:** For compute performance we should use certain flags (compiler and node type dependent), an example for Intel below:

``` bash
setenv I_MPI_FABRICS shm:ofi
setenv I_MPI_OFI_PROVIDER psm3
```
From Matt: The latter is required for Intel MPI on the Milan nodes. the former is good to use well.

- [ ] **Hold before certain tasks:** @rtodling mentioned this one. The ability for a suite to "hold" before a certain task, say `RunJediVarExecutable` so that different configurations could be tested swiftly without the need to `swell create` a whole new suite. JCSDA's https://github.com/JCSDA-internal/skylab uses this feature in a different workflow engine called EWOK.
- [x] **Restart a failed suite from a certain cycle & task:** This one is straightforward with using TUI. Involves using reinstall-reload function. I will add this to the documentation.

## Optional but useful ones:

- [ ] Cylc 8.5.0 has an option to use external editors to view log files, that is a great feature:
https://cylc.discourse.group/t/tip-cylc-tui-external-log-viewer/1236

- [ ] Workflow [sending email](https://cylc.github.io/cylc-doc/stable/html/tutorial/furthertopics/message-triggers.html) if suite stops/fails. I was able to do this with a sandbox Cylc setup.
- [ ] Failed `gmao_ci` workflows posting `job.err` and `job.out` somewhere on git for identifying the problem.
- [ ] **Using more SLURM and Milan node features**:
   - [ ]  **Assign different walltime for tasks:** Default is 1hr but most tasks take less than this, which could improve queue times. 
   - [ ] **Shared nodes:** Milan nodes has a feature to share nodes between users, for low demand jobs this could save queue time.
   - [x] **Per-host:** Not sure about this one, @rtodling uses it, relevant for compute performance.

Dooruk:
Case in point for retrying, the task had SLURM issues (pink square) but now running after two failed attempts and with zero changes. I just happened to be monitoring:

[...]

PR #529, opened 2025-04-07
Retry with Cylc and IAU time

### LINKED ISSUE #530: Retry certain tasks with Cylc

Dooruk:
### The good:
This PR introduces a new Cylc feature to SWELL: `retry`. With the simple [runtime] change (see #490 for more details) it is possible to retry failed task after a buffer time. Currently, it is only implemented for `RunGeosExecutable` since that one tends to fail randomly without throwing any errors and this switch prevented multiple workflow failures already.

### The bad:

@mranst, I'm having trouble with the second part. I'm trying to introduce a key called `mom6_iau_nhours` that depends on `mom6_iau` being active. In ODAS, we need a `mom6_iau` (IAU stands for Incremental Analysis Update) switch. With ocean DA, we are testing different DA windows so there will be different IAU hour lengths (e.g, 3H for 6-hr DA, 18H for 5-day DA). This switch is irrelevant to GEOS ADAS but will be relevant to AODAS.

Anyways, with it current state I receive the following error:

Screenshot 2025-04-07 at 11 51 55 AM

So I think SWELL gets to `mom6_iau_nhours` before having a `mom6_iau` key?

For more context, `PrepGeosRunDir` task doesn't have access to model argument `-m` because there is a `forecast_geos` suite that uses this task which doesn't conduct any DA. Only time entry `forecast_geos` would need is `forecast_duration`.

To provide a larger context, since @mathomp4 is working on updating `gcm_setup`, this question could go into that step instead.  And @mranst's new dictionary PR might be the ultimate savior from this mess.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.yaml
@@ -5,7 +5,7 @@ members from template:
     filetype: cube sphere history
     provider: geos
     datapath: '.'
-    filenames: ['geos.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4',
+    filenames: ['ebkg/mem%mem%/geos.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4',
                 'fv3-jedi/bkg/geos.crtmsrf.{{horizontal_resolution}}.nc4']
     state variables: [ua,va,t,delp,ps,q,qi,ql,qr,qs,o3ppmv,phis,
                   qls,qcn,cfcn,frocean,frland,varflt,ustar,bstar,
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/ensemble_block.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/ensemble_block.yaml
@@ -4,7 +4,7 @@ members from template:
     filetype: cube sphere history
     provider: geos
     datapath: '{{cycle_dir}}'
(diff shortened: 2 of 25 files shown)


mranst:
> So I think SWELL gets to `mom6_iau_nhours` before having a `mom6_iau` key?

This is a problem with model-independent questions having model-dependent dependencies. `mom6_iau` is a model-dependent field, and `mom6_iau_nhours` is model independent. Unfortunately I don't think (#514) would do anything to easily address this issue

Swell was sort of designed so that model-independent questions always come first, and then model-dependent questions. We'd need to come up with a new dependency logic so that independent questions can look at each model and match certain values

Dooruk:
> > So I think SWELL gets to `mom6_iau_nhours` before having a `mom6_iau` key?
> 
> This is a problem with model-independent questions having model-dependent dependencies. `mom6_iau` is a model-dependent field, and `mom6_iau_nhours` is model independent. Unfortunately I don't think (#514) would do anything to easily address this issue
> 
> Swell was sort of designed so that model-independent questions always come first, and then model-dependent questions. We'd need to come up with a new dependency logic so that independent questions can look at each model and match certain values

Ok thanks, your answer helped me to fix this problem! Both are model dependent.

However, now `PrepGeosRunDir` doesn't work because it doesn't have the model dependent config entries (`mom6_iau` or `mom6_iau_nhours` due to lack of `-m` model argument. I don't think there is a solution for this. I understand this is a flawed logic; non-model specific task requires model specific keys but with DA this will likely happen more and more. Initially I got around this issue by checking if certain files exist in the forecast INPUT directory but now I'm having to manipulate a MOM6 input file.

PR #536, opened 2025-04-24
Add EvaIncrement to local_ensemble_da workflow

metdyn:
## Description
Minimal changes to add Eva scripts for plotting increment for local ensemble DA

## Dependencies
None

## Impact
None


src/swell/tasks/eva_increment.py
```diff
@@ -66,6 +66,8 @@ def execute(self) -> None:
         # For now we are only plotting the first one
         iter_no = 1
         incr_file = f'{self.experiment_id()}.increment-iter{iter_no}.{cycle_time_reformat}.nc4'
+        if self.suite_name() == 'localensembleda':
+            incr_file = f'geos.mean-inc.{cycle_time_reformat}.nc4'
```
Dooruk: @rtodling is this naming convention something you are fine with, is it ADAS compatible?

Dooruk:
@metdyn, is the suite ready to be added to CI? I ran before your last commit and it failed:

https://github.com/GEOS-ESM/swell/actions/runs/14650347175

metdyn:
Thanks for letting me know, @Dooruk.  My case run with one obs atms_n20 worked last night. 
The full case with 32 mem needs:
* perhost=32  (96 processors already up: nx=ny=4)


slurm_directives_tasks:
    execution_time_limit: 'PT20H59M'   [ > 2 Hr]
    geos_atmosphere:
      nodes: 3
      ntasks-per-node: 126

  RunJediObsfiltersExecutable:
    execution_time_limit: 'PT5M'
    all:
      nodes: 1
      ntasks-per-node: 1

Dooruk:
@metdyn have you tested this suite with more nodes for `RunJediLocalEnsembleDaExecutable` rather than playing with perhost?

metdyn:
> @metdyn have you tested this suite with more nodes for `RunJediLocalEnsembleDaExecutable` rather than playing with perhost?

This is a C90 case with nearly full obs types, and I used 3 nodes with perhost=32 for 16 members which takes about 90 min on SLES15, longer time is needed for 32 members.  It is very heavy calculation.  More nodes will work, as long as  perhost * nnode = 6 * nx * ny.   It  appears that even with mem=2, we still need perhost=32 and 3 nodes, but the job will fit in PT1H.

Dooruk:
Would it be possible for you to edit:

https://github.com/GEOS-ESM/swell/blob/b3159f98f4a25a8f0243f212273755ace1b0dc39/src/swell/utilities/slurm.py#L45-L49

to assign `localensembleda` an hardcoded exception? Let's stick to assigning 3-4 nodes for CI purposes. For CI we need a lean application. So we could either use less members (less than 16, say 4-6) or use less observers. I would lean towards using less members.

metdyn:
Hi @Dooruk,
Thanks and I will try c12 with 4 members to see this can fit to 3 nodes with 10 min window on discover.

Dooruk:
> Hi @Dooruk, Thanks and I will try c12 with 4 members to see this can fit to 3 nodes with 10 min window on discover.

Thanks,  10 minutes is not strict but 20-30 minutes tops would be ideal.

You can try having two setups, one with c12 and one with c90 for tier 1 and tier 2 tests. Thanks!

metdyn:
offline to fix errors.

metdyn:
Ready for continued reviews

Dooruk:
Done with #598

Issue #497, opened 2025-02-06
SWELL or CYLC too slow in starting tasks


rtodling:
I don't know if anyone has noticed this, but I believe either swell or cylc are slow in starting up tasks. For example, if I were to write a c-shell program that looked at the directory where the obs files are and copy them over to a give location, I guarantee you it would be considerably faster ... there seems to be a delay of when tasks start- and that's applicable to every task -  don't know if this is a cylc thing evaluating what the flow entails and then figuring out what can do in parallel, etc or something.

All I know is that copying the obs file and bkg files into a work directory and constructing the yaml for a Var runs should be a much faster process than it is now - perhaps there is a way to time how the delay between things the launching of the suite and the time when things actually start happening.

ashiklom:
Thanks Ricardo. Agreed that this kind of thing should be basically instantaneous. We had a discussion of this at the Swell development meeting.

First thing we should do is confirm that cylc itself is the culprit. Best way to do this is check the task timings (which we already capture) to make sure that the tasks themselves execute quickly.

https://github.com/GEOS-ESM/swell/blob/develop/src/swell/tasks/base/task_base.py#L323-L329

Second, we should confirm that cylc doesn't introduce significant overhead into super simple tasks. I suggest setting up some trivial cylc workflows with a bunch of `echo "hello world"`-style tasks, just to make sure that these trivial workflows run as instantaneously as they should. (As a bonus, run these same experiments with the `module load` commands Swell's cylc binary expects; that will give us some clues about the overhead of that module loading every time).

Assuming neither of those is the issue, that leaves us with some more complicated issues. One likely culprit we discussed is that the cylc binary loads the full set of required modules every time, which introduces nontrivial overhead. If that's the bottleneck, we need some way to load the modules once, capture the resulting environment, and then propagate that environment (which is basically free) via something like `subprocess.run(..., env=<captured_env>)` in the places where we actually execute cylc tasks.

metdyn:
I agree with Ricardo. I see slowness in running command line on Milan node
`swell create hofx  -o override.yaml -s slurmfile.yaml -p nccs_discover_sles15` 
4-5 second to start printing first line output now vs. 1-2 sec 3 month ago. This is after module load the same spack-stack 1.7.

metdyn:
I have to add that:  
First I execute `swell create hofx  -o override.yaml -s slurmfile.yaml -p nccs_discover_sles15`, it is slow.  (I launch the job afterwards)

[...]

PR #537, opened 2025-04-28
Add entry point to Cylc in Swell

### LINKED ISSUE #538: Something wrong with code test installation
Installation of swell during code tests is failing with an uninformative error. 

Image

At first I thought it was something wrong with my PR #537, since it did change `setup.py`, but it's happening with develop as well despite working fine several days ago

https://github.com/GEOS-ESM/swell/actions/runs/14596706151/job/41303979867

Having difficulty making sense of it

mranst:
This PR addresses #497, eliminating the need for ~/bin/cylc by adding custom entry point to setuptools for swell. Now, when swell is installed, a cylc executable pointing to the correct installation of Swell on Discover will also be installed. I also added tools for detecting the platform on discover. In the case that the platform can't be identified, the executable will just try to run any available cylc in the path.

This PR eliminates the need for configuring cylc, and also lets us control cylc versioning totally within swell. Cylc should run significantly faster now, as well.


mranst:
Believe #538 has been fixed. It seems like the issue was installing python/pip and then seperatealy using conda on top of it, now the code tests should use the pip that was installed by python

src/swell/deployment/platforms/platforms.py
```diff
@@ -83,3 +85,38 @@ def login_or_compute(platform) -> str:
 
 
 # --------------------------------------------------------------------------------------------------
+
+
+class SwellPlatform(Enum):
```
Dooruk, line 90: We should extent this to AWS or Mac laptop later on.

Dooruk (dismissed):
Works good for me. 

Let me check if it works on CI without any changes other than the `~/bin/cylc` file renaming.

Dooruk:
Well, things are running just fine without the `~/bin/cylc`:

https://github.com/GEOS-ESM/swell/actions/runs/14756694620

ashiklom (dismissed):
Clever, creative, and elegant! Nice work. Approved, assuming tests are passing.

ashiklom:
NOTE: Currently blocked by @Dooruk 's comment about extending this to other platforms down the line. Let's convert that to an issue and resolve to get this in.

That said, if you have time, I suggest also updating our documentation to reflect this change. For now, we can leave the old text but put a warning/section at the top of the current cylc configuration page that says something like, "After PR 537 / commit XXXX, swell will configure the cylc binary automatically, so no change is needed. Instructions below are for earlier versions of Swell and will be removed in later versions."

Dooruk:
Ok, resolved my comment. Ready for merge unless @mranst wants to make the documentation edit.

mranst:
> Ok, resolved my comment. Ready for merge unless @mranst wants to make the documentation edit.

I'm editing the docs right now

Dooruk (approved):
All good!

Issue #442, opened 2024-10-17
Swell workflows for common comparisons


ashiklom:
Two common use cases (but there may be others):

- Compare two different sets of pinned versions of JEDI
- Compare two different versions of GEOS

Each of these comparisons would probably be separate workflows (e.g., `swell create compare_jedi`; `swell create compare_geos`).

Key to these workflows is not only making sure that both workflows execute without error, but also to calculate relevant comparison diagnostics (e.g., absolute diffs; maybe some maps / time series).

Dooruk:
Some quick thoughts:

1) (optional) A simple prelimianry test could be running the `ctest`s inside the JEDI build directories (`build-intel-release/soca`, build-intel-release/fv3-jedi`) and see if the same number of tests are passing between two versions (it's never %100 pass on JEDI side!). Of course, there will be different cmake tests added occasionally.

From my notes, 

### Running ctests:

Request interactive compute nodes and source `modules` in `build-intel-release`. Navigate to `build-intel-release/soca` and then run the tests by issuing `ctest`. This will confirm that that everything was installed successfully. Some useful commands:

- ctest -N (list all the tests)
- ctest -V -I (run all the tests displaying full (verbose) output)
- ctest -R `<name>` (run a particular test using the name of the test)
- ctest -I 4,6 (run test numbers 4,5,6)
- ctest -V -I 5,5 (run test 5 with all the output)
  
2) For variational tests, at the end of the `jedi_variational_log` file, OOPS puts out the following cost functions stats for each observer (this one is for SOCA, and hence marine observations) and forecast-background for state variables.

```bash
End Jo Observations Errors
CostJo   : Nonlinear Jo(adt_3a_egm2008) = 266.798601, nobs = 5802, Jo/n = 0.045984, err = 0.100000
CostJo   : Nonlinear Jo(adt_3b_egm2008) = 256.433901, nobs = 5900, Jo/n = 0.043463, err = 0.100000
CostJo   : Nonlinear Jo(adt_c2_egm2008) = 165.992805, nobs = 3529, Jo/n = 0.047037, err = 0.100000
CostJo   : Nonlinear Jo(adt_coperl4) = 0.000000 --- No Observations
CostJo   : Nonlinear Jo(adt_j3_egm2008) = 328.166562, nobs = 7318, Jo/n = 0.044844, err = 0.100000
[...]

PR #540, opened 2025-05-05
Overhaul to flow.cylc templating

mranst:
## Description

This PR introduces a completely new method for templating flow.cylc during experiment creation, using an object-oriented approach in python. This is a functional prototype, but still incomplete. I plan to use this method to introduce new features to Swell, like #442 and #490. I wanted to open it up for comment, since it does change a lot.

The process for creating an experiment now looks like this:
1. Suite questions are asked and configured
2. The suite's workflow is then initialized, which defines the tasks used. Generating the graph can make full use of any suite questions.
3. From the graph, the model independent and dependent tasks are determined, which then are used to configure and ask task questions.
4. The answered task and suite questions are then used to complete the workflow file, by generating the workflow string.

This is a slight change to the way the process worked before, but means that there is a great deal of control over defining the workflow. Like the recent questionary overhaul, this comes at the cost of readability and adjustment to the new method.

Workflows are now defined by two major parts:
The `workflow.py` file, which sets the sections of flow.cylc
Task definitions in `task_runtimes.py`, which stores information about the runtime requirements for tasks.

Since there is a lot of common language in `flow.cylc` across swell suites, the object-oriented approach means that setting up a suite primarily requires just defining the graph. This is handled by string processing for now, but can eventually be advanced into a more abstract method.

This is what it looks like for `3dvar`: 
https://github.com/GEOS-ESM/swell/blob/71370dd6533720dd99dff6e66401be0dea79a696/src/swell/suites/3dvar/workflow.py#L25-L104

The other part is defining the tasks. Since there are really only a couple of types of tasks, setting them is pretty simple. I've set automation to reduce on redundant work, that can be overridden as necessary
https://github.com/GEOS-ESM/swell/blob/71370dd6533720dd99dff6e66401be0dea79a696/src/swell/tasks/task_runtimes.py#L23-L49

Tasks can now have different defaults set directly, and for different platforms. For example, slurm defaults can now be set in the section, rather than in `slurm.py`. Different platform defaults will give us more power to set time limits, ect.


src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,184 @@
...
+        if self.environment is not None:
+            environment_section = self.create_new_section('environment', self.environment)
+            runtime_section.add_subsection(environment_section)
+
+        if self.slurm is not None:
+            slurm_dict = {}
+            for key, value in self.slurm.items():
+                slurm_dict[key] = self.match_platform(value, platform)
+            else:
+                slurm_dict = {}
```
ashiklom: Indent error? Not following the logic here.

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,184 @@
...
+        return runtime_section
+
+# --------------------------------------------------------------------------------------------------
+
+
+@dataclass
+class Model(Task):
+    def __post_init__(self):
+        self.is_model = True
+        super().__post_init__()
```
ashiklom: Does this need to be set via `__post_init__`? Unless I'm missing something, seems like you should be able to just do:

```python
class Model(Task):
  is_model: bool = True
```

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,184 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+
+@dataclass
+class Cycling(Task):
+    def __post_init__(self):
+        self.is_cycling = True
+        super().__post_init__()
```
ashiklom: Ditto here.

src/swell/utilities/cylc_runtime.py
mranst: Yes, me neither apparently

src/swell/utilities/cylc_runtime.py
mranst: It's necessary to make use of the multiple inheritance structure I've got going with the tasks. If you had

```python
class Model(Task):
    is_model: bool = True

class Cycling(Task):
   is_cycling: bool = True

class Example(Model, Cycling):
   pass
```
you would end up with `Example` having the the attributes `is_model = True` and `is_cycling = False`, because only the `__init__` of the first parent is called. This may be a slight hack but I like the simplicity of the syntax when it comes to setting up tasks

ashiklom: Got it, thanks!

This feels a bit overengineered. Multiple inheritance in Python is tricky to reason about -- you pretty quickly have to start understanding advanced concepts like the Method Resolution Order and exactly how `super()` works.

My suggestion might be to have a conceptually simpler (albeit more verbose) approach that just uses the base `Task` class and passes these things on directly.

```python
    @dataclass
    class BuildJedi(Task):
        has_slurm: bool = True
        time_limit: bool = True

    @dataclass
    class StageJedi(Task):
        is_model: bool = True

    @dataclass
    class StageJediCycle(Task):
        is_cycling: bool = True
        is_model: bool = True
        base_name: str = "StageJedi"
        scheduling_name: str = "StageJediCycle-{model}"
```

This makes it easier to quickly customize each Task. Otherwise, you may have to jump back to the parent class to apply customizations.

The main reason _not_ to do this would be if `Model`, `Cycling`, etc. had some custom methods. (This is approximately the ["mixin" pattern](https://realpython.com/inheritance-composition-python/#mixing-features-with-mixin-classes)). But in our (current) design, all of the logic for handling `is_model`, `is_cycling`, etc. lives in the `Task` class anyway. And note that in the "Mixin" approach, the point is actually to _extend_ the _child_ class with new methods.

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,194 @@
...
+        elif self.time_limit:
+            time_limit = self.match_platform(self.time_limit, platform)
+            runtime_dict['execution time limit'] = time_limit
+
+        # Set the retry if this task needs it
+        if self.retry is True:
+            runtime_dict['execution retry delays'] = '2*PT1M'
+        if self.retry:
+            retry = self.match_platform(self.retry, platform)
+            runtime_dict['execution retry delays'] = retry
```
ashiklom: Should be `elif` for the second block? Otherwise, the second block will effectively always overwrite the first block, which I think is not what we want...?

mranst: This is intentional - I wanted there to be the capability to set a default time by simply setting the value to `True`. If the value is some other definition (string, platform-dependent dictionary), then it will accept that value - This may also be over engineered

src/swell/utilities/cylc_runtime.py
mranst: Yeah, makes sense. I was trying to reduce verbosity, but wasn't considering this

src/swell/utilities/cylc_runtime.py
ashiklom: I'm still not following the logic. If `self.retry = True` then _both of these blocks run_, which means effectively:

```python
runtime_dict['execution_retry_delays'] = "2*PT1M"
runtime_dict['execution_retry_delays'] = self.match_platform(self.retry, platform)
```

If that's the correct logic, then we should make the intent a bit clearer by inlining the logic or tweaking `match_platform` to be able to return a default value if `retry=True` or something.

src/swell/deployment/create_experiment.py
```diff
@@ -16,14 +16,15 @@
 import yaml
 from typing import Union, Optional
 
-from swell.suites.all_suites import AllSuites
+from swell.suites.all_suites import SuiteConfigs
```
ashiklom: Duplicate import below.

src/swell/utilities/cylc_runtime.py
mranst: Oh! Sorry, I missed what you were saying, this is what the logic should be:

https://github.com/GEOS-ESM/swell/blob/0ec63162898cb8da893d5db5c6433c4659436257/src/swell/utilities/cylc_runtime.py#L114-L121

src/swell/deployment/create_experiment.py
```diff
@@ -92,49 +93,38 @@ def prepare_config(
...
+    prepare_config_and_suite.set_model_independent_tasks(model_ind_tasks)
+    prepare_config_and_suite.set_model_dependent_tasks(model_dep_tasks)
+
+    # Ask the task questions
+    # ----------------------
+    experiment_dict, comment_dict = prepare_config_and_suite.configure_and_ask_task_questions()
+
+    # Update the workflow with the answered task questions
+    # ----------------------------------------------------
+    workflow.set_experiment_dict(experiment_dict)
```
ashiklom: Any particular reason to use a method here rather than just assigning to the attribute directly?

```python
workflow.experiment_dict = experiment_dict
```

Unless there's a [good reason](https://realpython.com/python-getter-setter/#deciding-whether-to-use-getters-and-setters-or-properties-in-python), I would probably suggest just setting the attribute directly and killing the setter method (and if needed, we can set properties to do some preprocessing later).

src/swell/deployment/create_experiment.py
```diff
@@ -92,49 +93,38 @@ def prepare_config(
...
-            logger.abort(f'SLURM file contains invalid keys: {slurm_invalid_keys}')
-        experiment_dict = {**experiment_dict, **slurm_dict}
+    # Get the list of tasks from the workflow's graph
+    # -----------------------------------------------
+    model_ind_tasks, model_dep_tasks = workflow.get_independent_and_model_tasks()
+
+    # Set the tasks to be used in preparing the suite
+    # -----------------------------------------------
+    prepare_config_and_suite.set_model_independent_tasks(model_ind_tasks)
+    prepare_config_and_suite.set_model_dependent_tasks(model_dep_tasks)
```
ashiklom: Ditto here to my comment below. Simpler to just do:

```python
self.model_independent_tasks = model_ind_tasks
self.model_dependent_tasks = model_dep_tasks
```

...and get rid of the extra methods.

src/swell/suites/all_suites.py
```diff
@@ -85,9 +89,60 @@ def wrapper(suite_config_enum):
...
+class SuiteConfigs(Enum):
     pass
 
 
 # --------------------------------------------------------------------------------------------------
+
+
+@construct_workflow_enum()
+class Workflows(Enum):
+    pass
```
ashiklom: This decorator approach is clever, but I wonder if it's a bit too clever, forcing future developers to learn a moderately advanced Python feature to wrap their head around what this does.

I think we can accomplish the same thing more simply by just creating `SuiteConfigs` and `Workflows` in this file as if it were a script (and, optionally, `del`-ing all intermediate variables afterwards).

src/swell/utilities/cylc_formatting.py
```diff
@@ -0,0 +1,121 @@
...
+    for line in string.split('\n'):
+        if reset:
+            line = line.strip()
+
+        if len(line) > 0:
+            line = f'{indent*level}{line}'
+
+        out_string += f'{line}\n'
+
+    return out_string
```
ashiklom: Can this be replaced by (or simplified with) builtin `textwrap.indent`?

src/swell/utilities/cylc_workflow.py
```diff
@@ -0,0 +1,284 @@
...
+        for line in string.split('\n'):
+            line = line.strip()
+
+            if len(line) > 0:
+                start = True
+
+            if start:
+                out_string += line
+
+        return out_string
```
ashiklom, line 66: As above, consider [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) from the standard library.

src/swell/utilities/slurm.py
```diff
@@ -36,163 +37,61 @@ def prepare_scheduling_dict(
...
-        }
+    # Expand experiment dict with SLURM overrides.
+    # NOTE: This is a bit of a hack. We should really either commit to using a
+    # separate file and pass it around everywhere, or commit fully to keeping
+    # everything in `experiment.yaml` and support it through the Questionary
+    # infrastructure.
+    # ----------------------------------
+    if slurm_file is not None:
+        logger.info(f"Reading SLURM directives from {slurm_file}.")
+        assert os.path.exists(slurm_file)
```
ashiklom: I know I was previously a fan of `assert` but I learned recently that this is not recommended outside of tests (because `python -O` and certain other optimizations will ignore these).

Better to handle exceptions explicitly:

```python
try:
  with open(slurm_file, "r") as slurmfile:
    slurm_overrides = yaml.safe_load(slurmfile)
except FileNotFoundError as err:
  raise FileNotFoundError(f"Slurm config {slurm_file} not found.")
```

src/swell/suites/all_suites.py
mranst: Yeah, I wondered about that too, I've implemented it more simply as an object

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,168 @@
...
+
+from typing import Union, Optional, Self
+from collections.abc import Mapping
+from dataclasses import dataclass
+
+from swell.utilities.cylc_formatting import CylcSection, indent_lines
+
+# --------------------------------------------------------------------------------------------------
+
+indent = '    '
```
ashiklom, line 18: Unused constant?

src/swell/utilities/cylc_formatting.py
```diff
@@ -0,0 +1,94 @@
...
+# --------------------------------------------------------------------------------------------------
+
+
+def indent_lines(string: str, level: int = 0, reset: bool = False):
+    # Reset line indentation for string, and indent lines by level
+
+    if reset:
+        string = textwrap.dedent(string)
+
+    string = textwrap.indent(string, '    '*level) + '\n'
```
ashiklom, line 36: I might consider defining a module-level constant for this, to avoid accidentally introducing inconsistent indentation. Maybe something like this to make it completely unambiguous how many spaces there are.

```python
INDENT=" "*4
```

src/swell/utilities/cylc_workflow.py
```diff
@@ -0,0 +1,279 @@
...
+        in_cycle = False
+
+        for line in self.scheduling.split('\n'):
+            comment = False
+            sub_strings = line.split(' ')
+
+            for sub_string in sub_strings:
+                sub_string = sub_string.strip()
+
+                if '#' in sub_string:
```
ashiklom, line 171: Can we assume all comments start at the beginning of the line? As written, this will also classify as comments likes like:

```
x => y   # Some comment
```

Since you've already stripped leading whitespace above, you can probably do:

```python
if sub_string.startswith("#"):
```

src/swell/utilities/cylc_workflow.py
```diff
@@ -0,0 +1,279 @@
...
+                    model_tasks[model].append(task_name)
+            else:
+                ind_tasks.append(task)
+
+        return ind_tasks, model_tasks
+
+    # --------------------------------------------------------------------------------------------------
+
+    def define_runtime_task_overrides(self) -> dict:
+        return {}
```
ashiklom, line 222: Unnecessary function?

src/swell/utilities/cylc_workflow.py
```diff
@@ -0,0 +1,279 @@
...
+from typing import Union, Optional, Self, Tuple
+from collections.abc import Mapping
+
+from swell.utilities.cylc_formatting import CylcSection, indent_lines
+from swell.tasks.task_runtimes import TaskRuntimes
+
+# --------------------------------------------------------------------------------------------------
+
+
+class CylcWorkflow():
```
ashiklom, line 19: This is a great example of a class that strongly benefits from some unit tests.

A basic set of tests just creates a few representative workflows, then _manually_ writes out the expected `flow.cylc` files, and then confirms that your code produces identical workflows to the hard-coded ones.

A more sophisticated set of tests might also generate a variety of different cylc workflows and push them through `cylc validate` to make sure they're valid.

ashiklom (commented):
Overall, I'm happy with this -- nice work! A few minor tweaks. The most important suggestion is to add some unit testing of the dynamically generated cylc workflows. If we agree to this approach (and I think we should!), it's going to be a pretty essential piece of internal plumbing that we'll want to make sure is pretty bulletproof.

mranst:
Closing this PR, obselete because of #576

PR #543, opened 2025-05-06
Add fix to handle aircraft bias convertion

rtodling:
## Description

This provides a correction for converting aircraft biases. The convert stopped working after we split aircraft into temperature and wind.

## Dependencies

none

## Impact

allows for convert to work.


Dooruk:
FYI @imoradi

Dooruk:
@rtodling this requires a simple pycodestyle fix

imoradi:
Thank you for trying to fix this, but there seems to be still an issue with aircraft files.

rtodling:
> Thank you for trying to fix this, but there seems to be still an issue with aircraft files.

@imoradi what issue? I have have no issues - I've ran this for multiple cases and different periods ... can you be specific?

imoradi:
That is what we discussed last time we chatted on Teams. The problem is resolved if we remove the missing obs diags  from Yaml file. So I was able to run it. Thanks

Dooruk:
@rtodling this needs pycodestyle fixes

mranst (approved):
👍 FYI @Dooruk

PR #546, opened 2025-05-08
Always save cice restart

### LINKED ISSUE #547: Running `geos_marine` without cice6 analysis skip saving cice background

Dooruk:
CICE restart wasn't being saved if it isn't in marine models.


Dooruk:
nope this is not it

PR #549, opened 2025-05-09
Using EVA timeseries

### LINKED ISSUE #347: EVA time series task
Implementing the fresh EVA feature into SWELL

- This task should be able to run once every multiple cycles (say once a week). So in `flow.cylc` it should be treated differently.
- It should be able to check if the file exists already and able to append. The ideal location for the outputs would be under `{exp_dir}/run/{model}`
- We could replicate the parallelization strategy employed in `eva_observations`.
- Reinstall the new EVA version (Doruk)

Dooruk:
Introducing the new eva_timeseries task with a new suite eva_capabilities.  This suite currently relies upon previous SWELL runs and stored IODA outputs in R2D2. 

get_ncdiags -> looks for stored ncdiag, or SWELL IODA outputs, locally or shared from a previous experiment

The idea is that new EVA features and plotting approaches can be tested under a non-model specific suite, there can be different EVA suites, didn't think this as the only one.

With the help of new suite questions (thanks @mranst!) , it is possible to create model specific suites via `eva_capabilities` (marine only) or `eva_capabilities_atmosphere`.

Some atmosphere observations are missing, timeseries fails if no IODA outputs were found.

Adds timestamps to `cycle_done` per @gmao-yzhu's request

@ftgoktas  I'm tagging you so you can see what timeseries task is doing and you should be able to run it.

closes #347


Dooruk:
@mranst with the new question dictionary approach, do you think there is a way to set different cycle times for a suite depending on the model type? For instance I only want to run `geos_marine` during 2021 and `geos_atmopshere` during 2023. Seems possible but couldn't wrap my head around it. I'm not suggesting two model cycle times should run together.

Screenshot 2025-05-20 at 12 09 38 PM

mranst:
> @mranst with the new question dictionary approach, do you think there is a way to set different cycle times for a suite depending on the model type? For instance I only want to run `geos_marine` during 2021 and `geos_atmopshere` during 2023. Seems possible but couldn't wrap my head around it. I'm not suggesting two model cycle times should run together.
> Screenshot 2025-05-20 at 12 09 38 PM

If I'm understanding the problem correctly (you're only going to be running one model in an experiment), the easy method is to create separate sub-configurations based on separate models, for example you could have `eva_capabilities_atmosphere` and `eva_capabilities_marine`:

```python
eva_capabilities_atmosphere = QuestionList(
    list_name="eva_capabilities_atmosphere",
    questions=[
        eva_capabilities,
        model_components=['geos_atmosphere']
        qd.start_cycle_point('2023')
        qd.end_cycle_point('2023')
    ]
)
```


```python
eva_capabilities_marine = QuestionList(
    list_name="eva_capabilities_marine",
    questions=[
        eva_capabilities,
        model_components=['geos_marine']
        qd.start_cycle_point('2021')
        qd.end_cycle_point('2021')
    ]
)
```
You could also probably do it using the `depends` variable, but my intuition is that could get messy

Dooruk:
> If I'm understanding the problem correctly (you're only going to be running one model in an experiment), the easy method is to create separate sub-configurations based on separate models, for example you could have `eva_capabilities_atmosphere` and `eva_capabilities_marine`:

Yeah this is exactly what I needed, thanks! So, when I run `swell create --help` it shows me `eva_capabilities|ev                   a_capabilities_marine|eva_capabilities_atmosphere` which is fine but with the new method the suite folder name does not carry any weight as I understand?

Dooruk:
I would like to get a review on this. Maybe I can talk about it on Monday as I see it is hard to review what is happening just by looking at the code.

@ftgoktas, if you would like to give this a try, the following should work now:
`swell create eva_capabilities`

DavidSRussell:
Nice! Just tried this branch for the first time and the default example works for me.

src/swell/tasks/base/task_base.py
```diff
@@ -235,16 +236,28 @@ def cycle_time(self) -> str:
...
     # ----------------------------------------------------------------------------------------------
 
-    def first_cycle_time(self) -> str:
+    def start_cycle_point(self) -> str:
 
         return self.__start_cycle_point__.string_iso()
 
     # ----------------------------------------------------------------------------------------------
 
-    def first_cycle_time_dto(self) -> dt:
```
mranst, line 244: GetObservations task has a call to `first_cycle_time_dto`, could you change it there?
https://github.com/GEOS-ESM/swell/blob/469ba3fdde87f684c6099fc130d50810d090b01f/src/swell/tasks/get_observations.py#L205

mranst (changes requested):
Ran the default cases for marine and atmosphere and had a look at the plots, looks great! I can approve after the one change I requested

Dooruk:
While testing this branch, I had to install a newer version of EVA. New EVA installations fails during the EvaObservations task (other Eva tasks work btw), I tried installing with `requirements.txt` and  `requirements_sles15.txt` but no avail.

This is the latest error I'm receiving:
Screenshot 2025-08-11 at 11 57 52 AM

The default installed version, `eva/sles15_skylab9` was installed a few months ago and contains different package versions compared to my new installation with pip3. I tried to capture the versions previously installed with `pip freeze` and other packages like `pipdeptree` and `pipreqs` but that didn't work. So I declare defeat, I even tried limiting package versions using @mranst previous installation.

This issue impacts EVA more as it imports lots of packages and hence too many dependencies.

FYI @ashiklom

src/swell/tasks/task_questions.py
```diff
@@ -651,3 +656,29 @@ class TaskQuestions(QuestionContainer, Enum):
     )
 
     # --------------------------------------------------------------------------------------------------
+
+    StageJedi = QuestionList(
+        list_name="StageJedi",
+        questions=[
+            swell_static_file_questions,
+            qd.gsibec_configuration(),
+            qd.horizontal_resolution(),
```
mranst, line 669: You should add gsibec_nlats and gsibec_nlons back to this task, that's contributing to part of the problem in #612 (the file names having NonexNone in them). The code tests didn't pick this up because these fields are optional for some suites but not others

Dooruk: Ok, I think I did in the latest version

Dooruk:
@mranst this is ready to go in but I need an approval please

mranst (approved):
Looks great!

PR #552, opened 2025-05-14
Point directly to cylc installation on Discover

### LINKED ISSUE #551: swell launch exp/exp-suite  fails to complete
After swell create exp ..,  I find `swell launch` is halted on the terminal at discover.  I am using the develop version. 

gpt wrote:  This kind of error usually comes from a bug in the workflow launcher or a wrapper script that keeps launching child shells.

I wonder if others are finding the same issue today.

swell launch /discover/nobackup/projects/gmao/aist-nr/$USER/br_jedi_scu17/run_swell/t1/t1-suite

 DeployWorkflow: Launching workflow defined by files in '/gpfsm/dnb07/projects/p234/.../br_jedi_scu17/run_swell/t1/t1-suite'.
 DeployWorkflow: Experiment name: t1-suite
bash: warning: shell level (1000) too high, resetting to 1
bash: warning: shell level (1000) too high, resetting to 1

mranst:
## Description

The `cylc_swell` entry point now directly points to the cylc executable on discover. I suspect it will fix #551, but this is a more robust method regardless.

### DIFF
--- a/src/swell/cylc_swell.py
+++ b/src/swell/cylc_swell.py
@@ -51,8 +51,6 @@ def configure_cylc_environment(append_dict: dict = {}) -> dict:
 def execute_cylc(argv=sys.argv) -> None:
     ''' Custom entry point for Cylc on Discover. '''
 
-    cylc_command = ['cylc'] + argv[1:]
-
     logger = Logger('SwellCylcEntryPoint')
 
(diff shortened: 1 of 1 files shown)


Dooruk (approved):
It works, looking back we should have warned people about `~/bin/cylc` 😄

PR #553, opened 2025-05-15
Assimilate ADT from Sentinel-6A satellite

DavidSRussell:
## Description

The Sentinel-6A altimetry satellite entered service in 2021 and GMAO has IODA observation files for it, but Swell does not yet have it in its list of observation types. This update adds it to the list. I've tested it with a single 6-hr cycle of 3D-FGAT and the EVA diagnostics look as expected.


Dooruk (approved):
Looks good 👍

PR #566, opened 2025-05-30
Remove redundancy in data/time that swell cannot handle

rtodling:
## Description

Very unclear how the code before the fix here would have ever made it in develop! 

In any case, the change - update of mksi - fixes the issue which turns out to point to the disadvantage of revamping code: whereas the original FORTRAN database code can handle date/time redundancy; the SWELL revamp cannot and this is what cause the code to break in SWELL but not in GEOS.

## Dependencies

None

## Impact

None


Dooruk:
ok, triggered the Tier1.

Dooruk (approved):

rtodling:
This now  works.

Issue #563, opened 2025-05-23
EVA: Workflow for plotting from GEOS-JEDI obs-space output


rtodling:
Presently, GEOS-JEDI produces obs-space output in the following form:

expid/jedi/obs/Yyyyy/Mmm/expid.jedi_hofx.yyyymmdd_hhz.tar

When the tar-ball hold all obs-space (OMB and OMA)  output from a cycle.

The request here that we develop a Workflow that will:

1) pull the tar-ball from given GEOS-JEDI exp
2) unfold the tar-ball
3) determine what it finds
4) produce similar figures to those produced, say here: 
      https://gmao.gsfc.nasa.gov/intranet/science/gms/x0051/TERRA_html/obs2html/Y2023/M11/radiance/tovs_high_spectral/

 For (4), I suggest we name figure w/ exactly the same naming convention as produced in the obs2html framework.  This way we can simply use  the same html part and simply fill in figures in proper arrangement.

 From the perspective of a user looking for results from an experiment, this will function exactly the same way, so the user would not even notice she/he is looking at a GEOS-JEDI generated data.

 obs2html has different capabilities: 

  4A) maps of instantaneous snapshots, e.g., 
       
![](https://gmao.gsfc.nasa.gov/intranet/science/gms/x0051/TERRA_html/obs2html/Y2023/M11/radiance/tovs_high_spectral/G49/Y2023/M11/x0051.obsplot.2023110100.png)

 4B) Time series figure such as:

 
![](https://gmao.gsfc.nasa.gov/intranet/science/gms/x0051/TERRA_html/obs2html/Y2023/M11/radiance/tovs_high_spectral/G49/x0051.pseries.L4.2023110100-2023113018.png)

 4C) and summary plots such as the following (produced from averaging over time)

 
![](https://gmao.gsfc.nasa.gov/intranet/science/gms/x0051/TERRA_html/obs2html/Y2023/M11/radiance/tovs_high_spectral/G49/x0051.ptstats.Global.2023110100-2023113018.png)

I suggest that after steps (1)-(3) are done, we start w/ step (4A) since it does too require programs for average the data  specially or anything of this sort.

Dooruk:
@rtodling, is it possible for you to point some experiment outputs on Discover?

Dooruk:
Another plot type I captured during the last Friday's monitoring meeting (happy to explain this):

[...]

PR #576, opened 2025-06-12
(WIP) Workflows for common comparisons and misc additions

mranst:
## Description

This PR lays the groundwork for introducing workflows for common comparisons, and also introduces features such as email messaging and the option to set certain tasks for pause #490 .

Two rudimentary suites based on discussions in #442 have been introduced. `compare_variational` runs a simple fgrep command to find the residual norms in jedi log output. `compare_jedi` runs ctests on jedi builds and runs a crude comparison of which ones passed. To make this work, I introduced two additional options to the `swell task` command, an `iteration` flag which can be used to differentiate the same tasks being run on different experiments, and an `output` option which allows tasks that use it to send their test output to different locations (such as the comparison experiment, rather than the individual swell experiments that are being compared). These comparison suites make use of the updated cylc templating using python.

A command has been added `swell compare <type> (experiment locations)` to easily create comparison suites

Users can now select certain tasks to pause the workflow after by specifying a list in the overrides. This should allow for adjustments when developing/debugging. Using cylc's native capability for jinja2 templating, this can be toggled on when running `swell launch` using the flag `-d`.

Options have been added for sending email messages using cylc. This should theoretically work, according to cylc's syntax and example's I've seen from Doruk and Seth, but for some reason I don't actually receive email messages. Maybe someone else can try it. 

Adding a file `~/.swell/swell-settings.yaml` with the entry `email_address = <email address>`, and setting the configuration option `cylc_message_parameters` will add the language to flow.cylc. Email message won't be set unless they are toggled on using the flag `-m`, for convenience. All tasks will have message on failure/submit-failure preference by default, or can be set with custom events using a dictionary.


mranst:
Every suite has been incorporated into this new method, and the code tests have been updated to work. @Dooruk I've added a task that runs eva comparison plots for soca increments, if we the way this works I'll work on adding more. If you have any suggestions for the formatting let me know, I can't figure out how to get the title centered, for example

inc_Temp

Dooruk:
Can I test this directly from `feature/mranst/workflows_for_comparison`?

mranst:
> Can I test this directly from `feature/mranst/workflows_for_comparison`?

Yes, you can do `swell create compare_3dvar`, and manually enter locations for `experiment.yaml` for two experiments

Dooruk:
Thanks @mranst, substantial work! I'm looking into this now, I couldn't get it to run but a few questions:

- Is the procedure for comparisons as follows: 
1)`swell compare jedi | variational`, create experiment folder
2) modify `experiment.yaml`, add two experiment paths
3) swell launch
?

- This also includes flow.cylc changes right? Should I try to understand those changes from the other PR first?
- Two issues i see might be; 1) how to handle if two experiments were run on different times, years say. We could either explicitly define the time period for comparisons, or expect Swell to figure that out automatically by parsing `run` directories. 2) This is definitely less obvious but two experiments might have different data assimilation windows (this is only the case for marine workflows, they could be 6-hr,24-hr or 5 days). I think for simplicity we should warn the user and fail the suite if there is mismatch.

mranst:
> Thanks @mranst, substantial work! I'm looking into this now, I couldn't get it to run but a few questions:
> 
>     * Is the procedure for comparisons as follows:
>       1)`swell compare jedi | variational`, create experiment folder
> 
> 
>     2. modify `experiment.yaml`, add two experiment paths
> 
>     3. swell launch
>        ?
> 
> 
>     * This also includes flow.cylc changes right? Should I try to understand those changes from the other PR first?
> 
>     * Two issues i see might be; 1) how to handle if two experiments were run on different times, years say. We could either explicitly define the time period for comparisons, or expect Swell to figure that out automatically by parsing `run` directories. 2) This is definitely less obvious but two experiments might have different data assimilation windows (this is only the case for marine workflows, they could be 6-hr,24-hr or 5 days). I think for simplicity we should warn the user and fail the suite if there is mismatch.

I would specify the two file paths in an override file, run `swell create compare_3dvar -o override.yaml`. You can also add the filepaths after the fact and do `swell task EvaComparisonIncrement -m geos_marine experiment.yaml`.

Yeah handling the cycle times between two experiments is one of the more challenging aspects to this. I kept it very simple in this task, just parsing the common dates in the run directory, but the logic around matching windows should definitely be more fleshed out

The flow.cylc templating and all this other stuff kind of got wrapped up into one big PR, but this eva task doesn't necessarily need to be included in that since that logic is all handled within the task

Dooruk:
> 
> The flow.cylc templating and all this other stuff kind of got wrapped up into one big PR, but this eva task doesn't necessarily need to be included in that since that logic is all handled within the task

Ok, in that sense should I review https://github.com/GEOS-ESM/swell/pull/540 first or this one?

mranst:
> > The flow.cylc templating and all this other stuff kind of got wrapped up into one big PR, but this eva task doesn't necessarily need to be included in that since that logic is all handled within the task
> 
> Ok, in that sense should I review #540 first or this one?

That one is far outdated, I closed it

Dooruk:
Ok, I tried providing two experiment paths, suite immediately stops after `swell launch`. i can take a look tomorrow:

```yaml
# Datetime this file was created (auto added)
datetime_created: 20250717_165317Z

# Computing platform to run the experiment
platform: nccs_discover_sles15

# Record of the suite being executed
suite_to_run: compare_variational

# What is the experiment id?
experiment_id: swell-compare_variational

# What is the experiment root (the directory where the experiment will be stored)?
experiment_root: /discover/nobackup/dardag/SwellExperiments

# Provide paths to two experiments to run comparison tests on.
comparison_experiment_paths: ['/discover/nobackup/projects/gmao/advda/dardag/sles15/fgat_odasemc_01','/discover/nobackup/projects/gmao/advda/dardag/sles15/fgat_jra55_01']

# Enter the model components for this model.
model_components:
- geos_marine

# Since this suite is non-cycling choose how many hours the workflow can run ahead?
runahead_limit: P4

# Provide a dictionary mapping tasks to cylc event statuses, or 'auto' to automatically configure these based on the graph.
task_email_parameters: auto

# Specify any tasks that the workflow should pause on (for development purposes).
pause_on_tasks: []

# Configurations for the model components.
models:

  # Configuration for the geos_marine model component.
  geos_marine:

    # What is the duration between the middle of the window and the beginning?
    window_offset: PT6H

    # Do you want to use a 3D or 4D (including FGAT) window?
    window_type: 3D
```

mranst:
There was a bug, it should be good now

Dooruk:
So I'm trying to test this, I think I made some progress. It's a bit hard to follow with the flow.cylc changes.

Is compare 3dvar meant to obtain some flow.cylc configuration from the `3dvar` suite?

`compare_3dvar` only picked the last 6 cycles from overlapping cycle times for the following two experiments, and didn't grep the outputs..

Screenshot 2025-07-18 at 3 32 31 PM

mranst:
> So I'm trying to test this, I think I made some progress. It's a bit hard to follow with the flow.cylc changes.
> 
> Is compare 3dvar meant to obtain some flow.cylc configuration from the `3dvar` suite?
> 
> `compare_3dvar` only picked the last 6 cycles from overlapping cycle times for the following two experiments, and didn't grep the outputs..
> Screenshot 2025-07-18 at 3 32 31 PM

compare_3dvar takes information about window types and such from the suite config for 3dvar, that way it can be used in the task (like in the regular EvaIncrement task which runs differently for different suite parameters). But right now, the flow.cylc templating for the comparison suite is its own thing. 

It sounds like there's an error when it tries to grep the values, maybe it's trying to run on one cycle but not the other. I've added a check so it should now tell you what the problem is

Right now it'll only run the comparison plots for overlapping cycles, are there ones that it's missing from those?

Dooruk:
Ok, I'm having a better grasp of this so I will suggest a few key changes. This is surely a complicated development, will take a few iterations to make it more flexible/expendable, and is not trivial so let me know if you need hands on help or more clarification.

- I think finding the common timesteps is the essential part of all of these workflow comparisons business. So I suggest taking common timesteps function out of EvaCompare task, make it a utility to be used in the suites level, maybe under `all_suites.py`. Because there will be different types of suites going forward. Then, it can be a two step feature.

  Step 1: User defines start/finish cycle times
  Step 2: Swell figures out overlapping cycles,  between user time bounds and if none exists warns the user.

  While I was testing your PR, Swell struggled finding the common timesteps. I see you look at the timesteps under the `run` directory which is the right approach but while doing comparisons you might want to use datetime objects so you can order them properly.
  
- Maybe not in this PR, but going forward we may want to copy/link saved experiment outputs (whether from R2D2 or wherever @rtodling stores his experiments) to comparisons cycle directories. With my cycling tests, I save restarts or backgrounds and erase certain files, such as the increments. With R2D2 it would be easy, since it is standardized but without it that's more tricky. He did provide some bits of information here https://github.com/GEOS-ESM/swell/issues/563

  Below is what I do with `eva_timeseries` suite:
  https://github.com/GEOS-ESM/swell/pull/549

- `FGrepResidualNorm` is one type of an OOPS log parser but there could be more going forward since there is a lot more information inside the OOPS log. So I suggest giving this task a generic name like "oops_log_parser" (or someting more creative) and make this task consisting of different methods. I believe the proper SE term for this is "factory method" where you have an entry point function and have multiple subfunctions depending on what type of a function you would like utilize.

  This is what I'm doing in this following task and I can choose between `explicit_diffusion` or `bump` or add more options going forward.
  
  https://github.com/GEOS-ESM/swell/blob/839ff036a3270fe020cf046724ca0d016121c7e1/src/swell/tasks/generate_b_climatology.py#L50-L61
  
  The inputs could be implemented by a key in `experiment.yaml`, say `parser_options:`
  
-  There are some build in methods for what you are doing in a few places, e.g. `self.get_model()`, or `self.cycle_time_dto()`

mranst:
@Dooruk I've made some updates to the cycle matching. You can choose a start and stop cycle point and cycling interval manually, or set them to None and it will automatically match overlapping intervals. It will then cycle like a normal swell suite. It'll check that the window length and offset are the same between the experiments. I also have a version of it that directly matches the directories like I had it before and runs the cycles via integer cycling, but I like this way better

Dooruk:
A quick comment, the commented out part didn't work for me:

```py

    compare_3dvar = QuestionList(
        list_name="compare_3dvar",
        questions=[
            _3dvar,
            compare_variational,
            qd.model_components(['geos_marine']),
            qd.start_cycle_point(),
            # qd.start_cycle_point(default_value=None, widget_type=WidgetType.STRING),
            qd.final_cycle_point(),
            # qd.final_cycle_point(default_value=None, widget_type=WidgetType.STRING),
        ]
    )
```

I receive the following error:

```
    start_cycle_dto = max(start_cycle_dtos)
                      ^^^^^^^^^^^^^^^^^^^^^
ValueError: max() arg is an empty sequence

```

The parser task fails for me and I'm taking a closer look at that but the Cylc and the workflow aspect seems to be working.

mranst:
> A quick comment, the commented out part didn't work for me:
> 
> ```python
>     compare_3dvar = QuestionList(
>         list_name="compare_3dvar",
>         questions=[
>             _3dvar,
>             compare_variational,
>             qd.model_components(['geos_marine']),
>             qd.start_cycle_point(),
>             # qd.start_cycle_point(default_value=None, widget_type=WidgetType.STRING),
>             qd.final_cycle_point(),
>             # qd.final_cycle_point(default_value=None, widget_type=WidgetType.STRING),
>         ]
>     )
> ```
> 
> I receive the following error:
> 
> ```
>     start_cycle_dto = max(start_cycle_dtos)
>                       ^^^^^^^^^^^^^^^^^^^^^
> ValueError: max() arg is an empty sequence
> ```
> 
> The parser task fails for me and I'm taking a closer look at that but the Cylc and the workflow aspect seems to be working.

Do you have any experiments specified in override.yaml? This should only occur when it has nothing to search for, but I've pushed a fix for it

Dooruk (changes requested):
Ok I figured out the problem. In `JediOopsLogParser` `cycle_dir` is used, as such:

```
class JediOopsLogParser(taskBase):

    def fgrep_residual_norm(self):

        cycle_dir = self.cycle_dir()

```

this will not use the `run` folders from the original experiment folders but rather where you created your experiment directory.`cycle_dir` is defined in the base class.

Logs are located under at `comparison_exp1` and `comparison_exp2` `run`. For instance for my case i need my logs from the run/20210701T120000Z etc.

Screenshot 2025-07-31 at 2 50 07 PM

There are many ways to create this but one way might be to utilize `comparison_experiment_paths` by dropping the last two levels and combining `run` and isofolder information:

`/discover/nobackup/projects/gmao/advda/dardag/sles15/fgat_jra55_01/fgat_jra55_01-suite/experiment.yaml`

One last suggestion for subprocess, `run_track_log_subprocess` via from swell.utilities.shell_commands import run_track_log_subprocess spits out the error so it made it easier for me to find out about the error. I see that you need to provide an additional kwarg  for `results = output.stdout` to work but it may still work?

OR 

other option is to link or copy log files to your current exp folder but I think that is unnecessary

mranst:
> Ok I figured out the problem. In `JediOopsLogParser` `cycle_dir` is used, as such:
> 
> ```
> class JediOopsLogParser(taskBase):
> 
>     def fgrep_residual_norm(self):
> 
>         cycle_dir = self.cycle_dir()
> ```
> 
> this will not use the `run` folders from the original experiment folders but rather where you created your experiment directory.`cycle_dir` is defined in the base class.
> 
> Logs are located under at `comparison_exp1` and `comparison_exp2` `run`. For instance for my case i need my logs from the run/20210701T120000Z etc.
> Screenshot 2025-07-31 at 2 50 07 PM
> 
> There are many ways to create this but one way might be to utilize `comparison_experiment_paths` by dropping the last two levels and combining `run` and isofolder information:
> 
> `/discover/nobackup/projects/gmao/advda/dardag/sles15/fgat_jra55_01/fgat_jra55_01-suite/experiment.yaml`
> 
> One last suggestion for subprocess, `run_track_log_subprocess` via from swell.utilities.shell_commands import run_track_log_subprocess spits out the error so it made it easier for me to find out about the error. I see that you need to provide an additional kwarg for `results = output.stdout` to work but it may still work?
> 
> OR
> 
> other option is to link or copy log files to your current exp folder but I think that is unnecessary

This might be overcomplicated, but I designed that task to essentially run as part of the existing experiments. That way, we could potentially have JediOopsLogParser run just as part as the 3dvar suite without necessarily being tied down to the comparison suite in particular. The workflow templating will direct that task to run on the specific experiment.yaml in the comparison experiment paths:

image

Dooruk:
> This might be overcomplicated, but I designed that task to essentially run as part of the existing experiments. That way, we could potentially have JediOopsLogParser run just as part as the 3dvar suite without necessarily being tied down to the comparison suite in particular. The workflow templating will direct that task to run on the specific experiment.yaml in the comparison experiment paths
 
 Hmm I see, it does seem a bit complicated. So users have to specify that during the create process? Or would this task/suite automatically engage? 
 
I would think mostly folks would like to compare experiments after they run. For instance when we are testing Tier2 one test would be to compare runs with a static or stable JEDI build against the new one.

mranst:
> Hmm I see, it does seem a bit complicated. So users have to specify that during the create process? Or would this task/suite automatically engage?
> 
> I would think mostly folks would like to compare experiments after they run. For instance when we are testing Tier2 one test would be to compare runs with a static or stable JEDI build against the new one.

As long as the `comparison_experiment_paths` are specified in override.yaml, swell will handle setting up that task in flow.cylc automatically. Trying to alter the experiment paths after the creating the experiment and flow.cylc would be awkward. I can just alter the task to cycle through the paths by itself if that's too complicated

Dooruk:
> > Hmm I see, it does seem a bit complicated. So users have to specify that during the create process? Or would this task/suite automatically engage?
> > I would think mostly folks would like to compare experiments after they run. For instance when we are testing Tier2 one test would be to compare runs with a static or stable JEDI build against the new one.
> 
> As long as the `comparison_experiment_paths` are specified in override.yaml, swell will handle setting up that task in flow.cylc automatically. Trying to alter the experiment paths after the creating the experiment and flow.cylc would be awkward. I can just alter the task to cycle through the paths by itself if that's too complicated

Ok, maybe I'm not running this properly, let me read your description again.

rtodling:
@mranst I have completely followed the discussions here - although I spent sometime this AM reading through them. 

I will have to say that I am rather surprised by the amount of changes this requires ... in my mind, I would think this should require no change to existing DA flows!  

A compare suite should be independent of anything - and it should be flexible for how the user organizes the data -or perhaps less generally, it should allow for the organization of the data as in SWELL, but also as in GEOS-JEDI. I would like for the compare to be something that can be deployed by the average user and the the monitoring group; both will have different purposes - ultimately, not initially.

Dooruk:
@mranst would it be possible to demonstrate this PR and how the comparisons are done at the end of our meeting today?

mranst:
> @mranst would it be possible to demonstrate this PR and how the comparisons are done at the end of our meeting today?

Yep, sounds good

Dooruk:
@mranst this is good for me. If you can add one or two examples on the usage under `Practical Examples` in the documentation that would be great and then we can merge it.

mranst:
I cleaned up the comparison tasks and added one that creates jedi output plot comparisons for two experiments. I also added increment comparisons for geos_atmosphere. You can now just run `compare_variational_marine` and `compare_variational_atmosphere`, rather than `compare_3dvar`.

Dooruk:
i would like to focus on this next, this is some substantial update though. 

This also includes significant flow.cylc changes. Maybe we should have a meeting to discuss them, there is a limited amount of people involved in the flow.cylc modification.

mranst:
> i would like to focus on this next, this is some substantial update though.
> 
> This also includes significant flow.cylc changes. Maybe we should have a meeting to discuss them, there is a limited amount of people involved in the flow.cylc modification.

Sounds good. We're going to have a lot going on at the next swell meeting, it might make sense to schedule an additional meeting in the next week or so

mranst:
@Dooruk @ashiklom This should be ready for another review pass. I've simplified some of the comparison suite, and added another task called `TierTestOutput` which parses jedi log analysis and formats it into a comparison table, which looks like this:
 
image

I've added some documentation for the flow.cylc changes under `docs/examples/templating_workflows.md`, and also some for the comparison workflow under `docs/examples/soca/comparison_workflows.md`

Dooruk:
> @Dooruk @ashiklom This should be ready for another review pass. I've simplified some of the comparison suite, and added another task called `TierTestOutput` which parses jedi log analysis and formats it into a comparison table, which looks like this:
> image
> 
> I've added some documentation for the flow.cylc changes under `docs/examples/templating_workflows.md`, and also some for the comparison workflow under `docs/examples/soca/comparison_workflows.md`

I like this output. Maybe we could add a third column for the difference? Then we could use that output for enforcing zero-diff (well, with a reasonable tolerance) for Tier 2 nightly tests (not necessarily for this PR)?

mranst:
> I like this output. Maybe we could add a third column for the difference? Then we could use that output for enforcing zero-diff (well, with a reasonable tolerance) for Tier 2 nightly tests (not necessarily for this PR)?

👍 Can set tolerances for different values

image

docs/examples/templating_workflows.md
```diff
@@ -0,0 +1,106 @@
...
+
+    # Create the graph section
+    graph_section = self.create_new_section('graph', graph_str)
+
+    return graph_section
+```
+
+The `define_graph_section` task is used to set the graph. There are a few built-in methods used to format the strings into cylc syntax. The `format_cycle` method is used to construct properly indented blocks for cycling intervals. `create_new_section` creates a section object that tracks indentation levels, and can be added to other sections, with properly handled indentation and spacing. Here the `graph` is constructed as one of the sections, since it itself is a sub-member of the `scheduling` block in a cylc graph. Any suite questions in the config can be referenced from `self.experiment_dict`. For example, this is often used to format the model component.
+
+## Tasks and the runtime section
```
Dooruk, line 39: Really appreciate the documentation, it helps me quite a bit to follow the changes.

One thing I'm wondering about how would we be able to add other types of potential entries for runtime tasks with this new approach?

Some examples for `RunGeosExecutable` below:

```cylc
        [[[job]]]
            shell = /bin/csh
```

```cylc
execution_time_limit
```

https://github.com/GEOS-ESM/swell/blob/f9fb2ae6946a82e494bbdbbff250eb2a2a2ae29c/src/swell/suites/3dvar_cycle/flow.cylc#L172

docs/examples/templating_workflows.md
```diff
@@ -0,0 +1,106 @@
...
+@dataclass
+class RunJediConvertStateSoca2ciceExecutable(Task):
+    is_cycling: bool = True
+    is_model: bool = True
+    time_limit: bool = True
+    slurm: dict = mutable_field({'nodes': {'all': 1,
+                                           'nccs_discover_cascade': 2}})
+```
+
+On the `nccs_discover_cascade` platform, `nodes` will be set as 2, but on any other platform it will be 1. User overrides will still work as they did previously.
```
Dooruk, line 106: very cool feature!

src/swell/tasks/jedi_log_comparison.py
```diff
@@ -0,0 +1,154 @@
...
+import re
+import numpy as np
+
+from swell.tasks.base.task_base import taskBase
+
+# --------------------------------------------------------------------------------------------------
+
+comparison_fields = {'Residual norm': {'delimiter': '=', 'dtype': float}}
+
+tolerances = {'Residual norm ( 0)': 0.01,
```
Dooruk, line 21: Number of the Residual norm array is dynamic and defined by the `number_of_iterations`

Dooruk (commented):
This is shaping up great. I think it is very close to merging, I would like others' input on this and I will bring it up during the development meeting.

I will make one majorish comment regarding the log file names. For compare variational, only `jedi_variational_log` is used here. For FGAT, which is another flavor of variational, log name would be `jedi_fgat_log`. One way to handle this might be to have another key in the override such as `comparison_log_type`, which would then dictate this in the EvaComparion tasks or `JediOopsLogParser` or some of the eva YAMLs.

then maybe we could define another comparison workflow within the same suite as such:

```py
    compare_fgat_marine = QuestionList(
        list_name="compare_variational_marine",
        questions=[
            compare_variational,
            qd.comparison_log_type('fgat'),
            qd.model_components(['geos_marine']),
        ]
    )
```

For `localensembleda` the log structure will be quite different so we don't have to worry about that for now, but my suggestion would also help with that going forward. @metdyn might shed some insight for OOPS log outputs for `localensembleda`.

Finally, I have to say, changes in `flow.cylc` are a bit confusing but I understand it will provide us a lot more flexibility in some regards and maybe to the extent we might be able to use another workflow management engine (?). Either way, only a handful of developers would be messing with that part so we can just explain to them, or expand upon the documentation as needed. 

Thanks again for working on this, this really provides substantial amount of changes.

PR #577, opened 2025-06-13
Add task to get observations from GEOS location

### LINKED ISSUE #575: Option to get observations to swell from existing GEOS-JEDI experiment
This is similar to the existing option to get backgrounds from existing GEOS ADAS experiments, SWELL task: 

get_geos_adas_background.py

This option would bypass getting observations from R2D2 and would instead point to a directory such as:

/discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda/20250321T120000Z/geos_atmosphere

and copy everything that's there into the run/20250321T120000Z/geos_atmosphere of a swell suite.

mranst:
## Description

#575 This PR adds a task to copy all files from `/discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda/geos_atmosphere/<datetime>`. I called this `GetGeosJediBackground`, though I can change this if there's a more ideal name. This task requires an entry in `experiment.yaml`:
```yaml
path_to_geos_jedi_background: /discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda
```
It can then be called using:
`swell task GetGeosJediBackground experiment.yaml -m geos_atmosphere -d 20250321T120000Z`.

I didn't add it to any suite's flow.cylc yet, but the machinery is there to add that entry when the task is called from by a workflow


rtodling:
@mranst @ashiklom I think there is a misunderstanding here. What I requested has nothing to do with getting the background. 

The directory I pointed to:  /discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda

has observation files - not background files. The request relates to the GetObservations task.

rtodling (dismissed):
Unfortunately, all the changes here are not correct.  All these have to be wiped out and changes need to be made around - but not in the same task - associated w/ GetObservations.

rtodling:
BTW This task needs to be re-titled: Add task to get observations from GEOS location

rtodling:
The above would have to look something like this:



ioda_locations_not_in_r2d2: /discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda

mranst:
@rtodling Is the behavior of this task what you were expecting (copying all files from that directory into the cycle directory), and the naming is the issue? Or does this need to be completely scrapped and incorporated into a different/existing task (GetObservations)

rtodling:
> @rtodling Is the behavior of this task what you were expecting (copying all files from that directory into the cycle directory), and the naming is the issue? Or does this need to be completely scrapped and incorporated into a different/existing task (GetObservations)

@mranst The behavior is correct, but it needs to be scrapped and placed in the GetObs task - getting observations has nothing to do with/ background - you should not need and have any changes in anything associated with/ the backgrounds - the idea is to have a opt within GetObs that bypasses all that GetObs is doing and simply copying all from the given directory to the run directory.

mranst:
I've updated the language to reflect that these are existing observations, and added a question trigger to the `flow.cylc` of  `3dfgat_atmos` as an example, I can remove/rework it depending on how we want to use this task. Files are now linked instead of copied

Dooruk:
@rtodling, are the observations in this folder going to be used and do they change? @ftgoktas is looking into registering more atmosphere observations into R2D2. So if you standardize the locations we could register files with experiment name as an index (542 or x0051 whichever).

however, I just noticed you separated some observation files into variables. I guess that is to resolve the bug regarding aircraft observations?

rtodling:
> @rtodling, are the observations in this folder going to be used and do they change? @ftgoktas is looking into registering more atmosphere observations into R2D2. So if you standardize the locations we could register files with experiment name as an index (542 or x0051 whichever).
> 
> however, I just noticed you separated some observation files into variables. I guess that is to resolve the bug regarding aircraft observations?

@Dooruk I'd strongly advocate against creating an r2d2 database based on the observations I generated - since things are still changing on my end ... that's way having this PR going is important, since it should not care of what's in the database - simply link/copy all ... I am going to pull this branch today and see how it works for me.

rtodling:
> I've updated the language to reflect that these are existing observations, and added a question trigger to the `flow.cylc` of `3dfgat_atmos` as an example, I can remove/rework it depending on how we want to use this task. Files are now linked instead of copied

@mranst thanks Mike, I will pull this branch today and see how it works for me. I'll provide feedback...

rtodling:
@mranst This now works for me just fine. Very nice!

mranst:
> @mranst This now works for me just fine. Very nice!

Great! If you want to approve the PR I can merge it

Dooruk (dismissed):
Sorry I know this is not related to me but I get email updates for swell PRs and this looks good 👍🏼

mranst:
> Sorry I know this is not related to me but I get email updates for swell PRs and this looks good 👍🏼

Caught a mistake, sorry. Could I get it again? Thanks

Dooruk (approved):

PR #578, opened 2025-06-13
Remove delp from bkg; avoid confusion

rtodling:
## Description

This removes DELP from the background; this not needed for an ak/bk type analysis and it introduces room for confusion in fv3-jedi. Change in results might be seen as roundoff; nothing relevant.

We want to do this to the ensemble as well - but lets do it in a different PR - I believe Yonggang has already a version that removes delp from the ensemble analysis vector (something that never was in Var). We will work on the delp removal and handling examination of how GNSSRO works in the ensemble else where - I believe there is an issue there too.

## Dependencies

None

## Impact

Prep for when handling of 3d-pressure updates change to allow proper use of GNSSRO observations.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/analysis.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/analysis.yaml
@@ -9,7 +9,6 @@ field io names:
   eastward_wind: ua
   northward_wind: va
   air_temperature: t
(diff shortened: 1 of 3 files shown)


Dooruk:
@metdyn please be aware of this change, I will merge it soon.

metdyn:
Thank you, @rtodling and @Dooruk.

Dooruk (approved):

PR #596, opened 2025-07-18
Fix for non-captured CI test output

mranst:
## Description

Makes it so that CI tests properly return non-zero exit status upon failure (this is so simple I can't believe I missed it)


Dooruk (approved):
Please find more simple fixes like this to all of our problems.

Is there a way to test this?

mranst:
> Please find more simple fixes like this to all of our problems.
> 
> Is there a way to test this?

Yes, I've tested it on one of my branches

PR #602, opened 2025-07-28
git-lfs changes with fv3-jedi

metdyn:
## Description

The PR is for 
- https://github.com/GEOS-ESM/swell/issues/600 
related to 
- https://github.com/JCSDA-internal/fv3-jedi/pull/1363

Tested with  fv3-jedi version 2025-07-23.

## Dependencies

- [ ] waiting on GEOS-ESM/repo/pull/<pr_number>

PR #603, opened 2025-07-29
SWELL now controls peripheral files

rtodling:
## Description

This is the permanent solution for what in PR #602  No more references to JEDI source code ... not sure if the ocean part has similar links to the source - if it does, I suggest that gets resolved in way similar to here - but in a separate PR.

Notice that the advantage of us controlling these files allowed me to easily - without the nightmare of approval process on JEDI - to get AKBK files for two new configurations of the model, look under:

/discover/nobackup/projects/gmao/advda/SwellStaticFiles/jedi/interfaces/geos_atmosphere/fv3files/

## Dependencies

None

## Impact

SHouldn't be any.


Dooruk (approved):
I will wait for @metdyn but looks good to me

metdyn:
Thank you, @Dooruk and @rtodling!

PR #604, opened 2025-08-01
Automatically set window length for cycling

### LINKED ISSUE #513: `-P1` does not work in the UTC mode

mranst:
## Description

This sets the UTC cycle point for MoveDaRestart using window_length automatically in flow.cylc

Does this solve #513 ?

### DIFF
--- a/src/swell/suites/3dfgat_cycle/flow.cylc
+++ b/src/swell/suites/3dfgat_cycle/flow.cylc
@@ -68,8 +68,7 @@
             # Model preperation
             # Run the forecast through two windows (need to output restarts at the end of the
             # first window and backgrounds for the second window)
-            # MoveDaRestart-{{model_component}}[-P1D] => PrepGeosRunDir
-            MoveDaRestart-{{model_component}}[-PT6H] => PrepGeosRunDir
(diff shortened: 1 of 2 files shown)


Dooruk:
Wow, it didn't occur to me we can use window_length from the yaml...

Let me test it, hope this is just that simple.

Dooruk (changes requested):
Yep this works, great job!

please repeat this for 3dfgat_cycle and erase the commented out moveDaRestart line then this is good to go.

mranst:
> Yep this works, great job!
> 
> please repeat this for 3dfgat_cycle and erase the commented out moveDaRestart line then this is good to go.

👍

Dooruk (approved):

Issue #558, opened 2025-05-23
Alternative handling of anchor variables


rtodling:
This  is to lay out two issues related to the way things work in SWELL that would more nicely work differently.  

1. Recently, I needed to make a change to one of the yaml files related to the atmosphere analysis. Basically, added something like this to background.yaml

file io names: &file_io_names
  eastward_wind: ua
  long_list_of_var_defs_here ...

This background file, simply defines the var name **file_io_names** - it doesn't use it.  I then edited yaml files, eg. varincrement1.yaml  and added

file io name: *file_io_names

notice this file, varincrement1.yaml does not define the anchor, it was supposed to simply used it. 

Unfortunately, SWELL crates w/ an error complaining that file_io)names is not defined. The issue is that the machinery that replaces the variable name by the actual entries only works when the achor is defined in that peanut yaml file - it's not enough to have it defined once in the beginning.

2. This to me unveil an even more fundamental issue. The problem above would be not be an issue if SWELL let the anchors as such - meaning if it never attempted to replace, say, file_io_names w/ it details.

SOLUTION: What I would like to see if for SWELL to leave the answer all *var_name untouched. So that the final yaml provided to the application will have the definition of the anchor and the *var_name in the final file.

mathomp4:
Swell could use ruamel.yaml to preserve anchors in output. For example:
```python
from ruamel.yaml import YAML
import yaml
from io import StringIO

yaml_str = """
var: &save This long string
var2: *save
"""

# ruamel.yaml preserves anchors
ruamel_yaml = YAML()
data_ruamel = ruamel_yaml.load(StringIO(yaml_str))
output_ruamel = StringIO()
ruamel_yaml.dump(data_ruamel, output_ruamel)

# PyYAML does not preserve anchors
data_pyyaml = yaml.safe_load(yaml_str)
output_pyyaml = yaml.dump(data_pyyaml, default_flow_style=False)

# Display both outputs
print("ruamel.yaml output (anchors preserved):")
print(output_ruamel.getvalue())

print("\nPyYAML output (anchors expanded):")
print(output_pyyaml)
```
The output of this is:
```
❯ python3 test_combined.py
ruamel.yaml output (anchors preserved):
var: &save This long string
var2: *save


PyYAML output (anchors expanded):
var: This long string
var2: This long string
```

ashiklom:
[...]

Issue #568, opened 2025-06-09
SE: General issues to ease GEOS-JEDI transition


rtodling:
The following are items that need to be worked on to facilitate ease of integration and transition to GEOS-JEDI.

Just as background to this issue it should be said that a release of GEOS-JEDI is already available - see https://github.com/GEOS-ESM/GEOSadas/wiki/Y.-GEOS%E2%80%90JEDI . However, there are still a number of issues that make these initial releases not yet ready for primetime.  in this respect, two main issues remain:

0) Allow SWELL to get observations from GEOS-JEDI

1) The observation part of the yaml is still wired to a particular time period.

2) Access to the observations still relies on GSI, its ncdiag output, and their conversion to IODA using SWELL.

Increased flexibility in GEOS-JEDI can be achieved if both issues above are tackled within SWELL. It should be notice that these two issues involve a number of such-issues.

- [ ] 0. **Getting observations from GEOS-JEDI**

see #575 

- [ ] 1.  **Flexible handling of observation yaml's**1a. Implement in SWELL a task that given date/time puts together the observation part of the large yaml normally formed to run any variational application. 

Presently, this machinery is embedded in the run variational task of SWELL; this needs to be extracted from there and made into a task of its own. This part involves looking at the observation database to activate and de-activate satellite channels used at the particular data/time. This needs to be pulled into the new task that would construct the an "obs.yaml" file.

1b. A pending complexity associated with the above is the fact that presently there is no mechanism in SWELL to handle conventional observations. This is a big topic that requires considerable thinking. The details on this need to go into a separate issue of its own. see issue #368 

1c. The obs.yaml file created by this new task should treat the data as being "local to the directory". That is, presently, each of the independent observation yaml file templates refer to the full path of the observation - for example, consider atms_n20.yaml  - it has things like:

obsfile: '{{cycle_dir}}/atms_n20.{{window_begin}}.nc4'

these need to be replaced with simply:

obsfile: 'atms_n20.{{window_begin}}.nc4' 

Since SWELL requires full file paths by construction, perhaps the best thing here would be to give an option to the task such that if the calling program (or user) wants, the yaml would be generated without the full path - this could be done as a post-processing of the obs.yaml file, where a little python function would be created to replaced {{cycle_dir}} with a period.

[...]

PR #605, opened 2025-08-08
JEDI oops and model configuration using python

mranst:
This PR changes all oops and model files (everything under interfaces/<model>/model) to python, instead of yaml. The primary benefit to this is straightforward imports for shared variables across files (addressing #558). A secondary effect is that it eliminates the `jedi_dictionary_iterator` method and jinja2 templating for these files, instead directly setting elements using dictionaries and python logic (except for the observation yamls, which are built in the same way as before). In this method, these configuration files act directly as source code, rather than being template files that are pulled to the experiment cycle directory and filled in later (that may be considered a negative).

Using this method, I was able to easily set up shared variables to be used across different files, eliminating the need for copy-and-pasting the same variables across multiple files when a change needs to be made. For example (note the state variables and field io names):

https://github.com/GEOS-ESM/swell/blob/fff3979d6314bc64c7b147b57059e47c7f0eb060/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.py#L10-L32

This *should* be fully functional, I've made sure that the tier1 tests pass, but there's a lot going on in these files so it would be good to get feedback from affected parties


src/swell/utilities/oops_config.py
```diff
@@ -0,0 +1,83 @@
...
+                 window_type: str,
+                 obs: list,
+                 cycle_time,
+                 jedi_forecast_model: str,
+                 observing_system_records_path: str) -> str:
+
+        self.jedi_rendering = jedi_rendering
+        self.logger = jedi_rendering.logger
+        self.jedi_interface = jedi_rendering.jedi_interface
+        self.template_dict = jedi_rendering.__template_dict__
```
ashiklom, line 43: [PEP 8](https://peps.python.org/pep-0008/) (emphasis mine):

> Even with __all__ set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be **prefixed with a single leading underscore**.

Surrounding double underscores like this should be reserved for Python special methods/attributes.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/analysis.py
```diff
@@ -0,0 +1,28 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+
+def analysis(template_dict: Mapping) -> Mapping:
+    analysis = {
+        'filetype': 'cube sphere history',
+        'provider': 'geos',
+        'datapath': template_dict['cycle_dir'],
+        'filename': '{experiment_id}.analysis.%yyyy%mm%dd_%hh%MM%ssz.nc4'.format(**template_dict),
```
ashiklom: For consistency and clarity (and to raise an `IndexError` if `experiment_id` is not set).
```suggestion
        'filename': f'{template_dict["experiment_id"]}.analysis.%yyyy%mm%dd_%hh%MM%ssz.nc4',
```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background.py
```diff
@@ -0,0 +1,79 @@
...
+def background(template_dict: Mapping) -> Mapping:
+    horizontal_resolution = template_dict['horizontal_resolution']
+
+    background = {
+        'datetime': template_dict['local_background_time_iso'],
+        'filetype': 'cube sphere history',
+        'provider': 'geos',
+        'datapath': template_dict['cycle_dir'],
+        'filenames': [
+            f'bkg.%yyyy%mm%ddT%hh%MM%ssZ.nc4',
```
ashiklom: Python linters should raise warnings for `f` strings with no variables in them.

```suggestion
            'bkg.%yyyy%mm%ddT%hh%MM%ssZ.nc4',
```

src/swell/configuration/jedi/interfaces/geos_marine/model/background_error_diffusion_vt.py
```diff
@@ -0,0 +1,44 @@
...
+                     'vertical': {
+                         'model file': {
+                             'date': template_dict['local_background_time_iso'],
+                             'basename': './',
+                             'ocn_filename': 'calculated_scales.nc'
+                         },
+                         'model variable': 'sea_water_potential_temperature'
+                     },
+                     'write': {
+                         'filepath': 'background_error_model/vt.{local_background_time}'.format(**template_dict)
```
ashiklom:
```suggestion
                         'filepath': f'background_error_model/vt.{template_dict["local_background_time"]}'
```

src/swell/configuration/jedi/interfaces/geos_marine/model/background_error_init.py
```diff
@@ -0,0 +1,51 @@
...
+# --------------------------------------------------------------------------------------------------
+
+
+def background_error_init(template_dict: Mapping) -> Mapping:
+    background_error_init = {
+        'covariance model': 'SocaError',
+        'analysis variables': template_dict['analysis_variables'],
+        'date': template_dict['local_background_time_iso'],
+        'bump': {
+            'io': {'data directory': '{cycle_dir}/background_error_mode'.format(**template_dict)},
```
ashiklom:
```suggestion
            'io': {'data directory': '{template_dict["cycle_dir"]}/background_error_mode',
```

src/swell/configuration/jedi/interfaces/geos_marine/model/getvalues.py
```diff
@@ -0,0 +1,19 @@
...
+
+from collections.abc import Mapping
+
+# --------------------------------------------------------------------------------------------------
+
+
+def getvalues(template_dict: Mapping) -> Mapping:
+    getvalues = None
+
+    return getvalues
```
ashiklom: This should fail type checking (`None` is not a `Mapping`). If we want to be consistent with these always returning `Mapping`s (which is a good idea), then:


```suggestion
def getvalues(_: Mapping) -> Mapping:
    return {}
```

(Also note: Python convention is to use an `_` for an unused variable. Without this, most linters should raise "unused variable" warnings).

src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -180,30 +183,79 @@ def __open_file_render_to_dict__(self, config_file: str) -> dict[Any, Any]:
...
+
+        # Import the module
+        module = import_module(f'swell.configuration.jedi.oops.{config_name}')
+
+        # Check that the module has a proper attribute for configuration
+        if not hasattr(module, config_name):
+            self.logger.abort(f'Config file {config_file} has no attribute {config_file}.')
+
+        # Get the config class
+        config_class = getattr(module, config_name)
```
ashiklom: It's OK for now because this is "idiomatic Swell", but just flagging that idiomatic Python style is less "look before you leap" and tends more towards "easier to ask forgiveness than permission".

https://realpython.com/python-lbyl-vs-eafp/

In this context, this can be written much more compactly as:

```python
module = import_module(f"swell.configuration.jedi.oops.{config_name}")
config_class = getattr(module, config_name)
```

The first statement will raise an informative `ModuleNotFoundError` if `config_name` doesn't exist. The second statement will raise an informative `AttributeError` if `config_name` doesn't exist in `module`.

In a narrow sense, this is functionally equivalent to `self.logger.abort` (both fail with informative error messages). But I would argue this behavior is actually *significantly better* because it gives users of this class the option to explicitly handle these errors. The `sys.exit` in `self.logger.abort` is a brute-force exit from Python that cannot be handled in any way. On the other hand, proper Python exceptions can be handled if appropriate; e.g.,

```python
try:
  result = my_oops.render_oops_file(my_config)
except ModuleNotFoundError as err:
  logger.warn(f"Module {my_config} not found. Assuming {}")
  result = {}
```

...or even re-thrown with more informative errors if we want.

```python
try:
  result = my_oops.render_oops_file(my_config)
except (ModuleNotFoundError, AttributeError) as err:
  logger.exception(f"Failed to load {my_config}. Ensure that a file called {my_config}.py exists and contains a {my_config} function.")
  raise err
```

In general, I would really like to start deprecating the use of `logger.abort` (and eventually to eliminate it altogether) in favor of more idiomatic Python error handling. But that's a longer term project --- I don't want it to hold up this PR.

src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -180,30 +183,79 @@ def __open_file_render_to_dict__(self, config_file: str) -> dict[Any, Any]:
...
+
+            # If the object is a function, call it with the template dictionary, if not set directly
+            if isinstance(config_func, Callable):
+                config_value = config_func(self.__template_dict__)
+            else:
+                config_value = config_func
+        else:
+            self.logger.abort(
+                    f'Interface model file {config_file} does not have an attribute {config_name}.')
+
```
ashiklom: Same comment as above.

src/swell/utilities/render_jedi_interface_files.py
```diff
@@ -180,30 +183,79 @@ def __open_file_render_to_dict__(self, config_file: str) -> dict[Any, Any]:
...
+        # Check that the file has an appropriately name attribute
+        if hasattr(module, config_name):
+            config_func = getattr(module, config_name)
+
+            # If the object is a function, call it with the template dictionary, if not set directly
+            if isinstance(config_func, Callable):
+                config_value = config_func(self.__template_dict__)
+            else:
+                config_value = config_func
```
ashiklom: Are there any instances where `config_func` is a value rather than a function? All the ones I saw above are functions.

To keep things less confusing, I would enforce consistency here. In the spirit of EAPF style (see above comment), just write this as:

```python
config_value = config_func(self.__template_dict__)
```

...and let Python raise an error if `config_func` is not a function.

src/swell/utilities/oops_config.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class OopsConfig(ABC):
+    # Abstract class that contains render information for oops configurations
+
+    def __init__(self,
+                 jedi_rendering,
+                 window_type: str,
+                 obs: list,
+                 cycle_time,
```
ashiklom: Can we assert either a `str` or some kind of date/datetime type here? Time/date variables are a case where it is **especially** useful to be able to communicate whether we are expecting a string representation, a numeric representation (e.g., Unix epoch -- int secs since 1970-01-01), or some kind of date/datetime object.

src/swell/configuration/jedi/interfaces/geos_marine/model/background_diffusion_vt.py
```diff
@@ -0,0 +1,27 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+
+def background_diffusion_vt(template_dict: Mapping) -> Mapping:
+    background_diffusion_vt = {
+        'date': template_dict['local_background_time_iso'],
+        'read_from_file': 1,
+        'basename': './',
+        'ocn_filename': 'MOM6.res.{local_background_time}.nc'.format(**template_dict)
```
ashiklom:
```suggestion
        'ocn_filename': f'MOM6.res.{template_dict["local_background_time"]}.nc'
```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry_inner.py
```diff
@@ -0,0 +1,35 @@
...
+    geometry_inner = {
+        'fms initialization': {
+            'namelist filename': './fv3-jedi/fv3files/fmsmpp.nml',
+            'field table filename': './fv3-jedi/fv3files/field_table_gmao'
+        },
+        'akbk': f'./fv3-jedi/fv3files/akbk{vertical_resolution}.nc4',
+        'layout': [npx_proc, npy_proc],
+        'npx': int(horizontal_resolution),
+        'npy': int(horizontal_resolution),
+        'npz': int(vertical_resolution)
```
ashiklom: This is good, and is a nice example of the advantages of having this live in Python code: We can enforce types early and loudly! (E.g., `int("3.0")` will fail loudly with a clear error).

That said, the "Parse, don't validate" philosophy (see suggested reading below) suggests that we should try to capture such type information as early as possible and carry it around with us. Here, that means that we should consider if we can define `template_dict` more precisely (e.g., as a dataclass with explicit types for all fields) so that all fields are validated and then all objects that work with `template_dict` know exactly what types everything should be.

Let's keep this for now. But flagging this as something to think about in future design.

https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

ashiklom (commented):
I requested a few small fixes and offered a few minor suggestions, but overall, I approve! To me, this is a massive improvement over the old YAML-based system, and opens the door for a lot more robust and flexible configuration. Nice work!

Ping me once you've taken a look at all my comments (one way or another) and I can hit "Approve".

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry_inner.py
ashiklom: On the other hand, just building validation directly into these functions is conceptually much simpler and maybe better for [Locality of Behavior](https://dev.to/ralphcone/new-hot-trend-locality-of-behavior-1g9k). So maybe it makes more sense to keep this type of structure. 🤷‍♂️ Food for thought!

src/swell/configuration/jedi/interfaces/geos_marine/model/getvalues.py
mranst: I added `Optional` to the type hint - in this case the file does need to return None

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry_inner.py
mranst: In this case, the type enforcing was trying to recreate subtleties in the yaml formatting, as I was aiming to have there be zero diff between files generated by the old system and the new system. The value `horizontal_resolution` in the dictionary is a string, but when it was templated into yaml by the `jinja2` templating, it essentially becomes an int (It likely makes no difference, but I wasn't sure if whichever aspect of jedi uses this field would be able to handle a string vs an int). Your comment also made me realize that in some cases horizontal resolution does have to be a string.

mranst:
> I requested a few small fixes and offered a few minor suggestions, but overall, I approve! To me, this is a massive improvement over the old YAML-based system, and opens the door for a lot more robust and flexible configuration. Nice work!
> 
> Ping me once you've taken a look at all my comments (one way or another) and I can hit "Approve".

@ashiklom Thanks! I've addressed most of the comments. One thing I was wondering about - The pycodestyle tests are failing mostly because of the 100 character line limit. In my opinion, this rule makes sense for readability, but for the purposes of these files, which are mostly for configuration purposes, I wonder if it actually makes important information less readable. There's a lot of long filenames being rendered here that would require being broken up into several lines, maybe making it harder to parse than the old yamls

ashiklom:
> 100 character line limit

You have a few options here. I would deploy them as you see fit, where it makes sense:
- Easiest is to tell pycodestyle to ignore those lines by adding a `# noqa` comment at the end of those lines.
- You can reduce the indenting by extracting some blocks into variables
    ```python
    saber_block = {
        'saber block name': 'gsi hybrid covariance',
        'read': {
            'gsi akbk': f'./fv3-jedi/fv3files/akbk{vertical_resolution}.nc4',
            'gsi error covariance file': f'./fv3-jedi/gsibec/gsi-coeffs-gmao-global-l{vertical_resolution}x{gsibec_nlons}y{gsibec_nlats}.nc4',
            'gsi berror namelist file': f'./fv3-jedi/gsibec/{gsibec_configuration}_l{vertical_resolution}x{gsibec_nlons}y{gsibec_nlats}.nml',
            'processor layout x direction': gsibec_npx_proc,
            'processor layout y direction': gsibec_npy_proc,
            'debugging mode': False
        }
    }
    background_error = {
      ...
      "saber central block": saber_block
      ...
    }
    ```
- Use multi-line strings, split in meaningful ways (note: this also uses Python's implicit string concatenation, so you don't actually need to use `+`
    ```python
    'gsi error covariance file': (
        './fv3-jedi/gsibec/gsi-coeffs-gmao-global-'
            f'l{vertical_resolution}'
            f'x{gsibec_nlons}'
            f'y{gsibec_nlats}.nc4'
    ),
    ```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry_inner.py
ashiklom: OK, let's leave this as is for now to keep the zero diff. This is a bigger conversation.

rtodling:
I will say that I like the idea of moving away from jinja2 templating; and yes, providing the support for the ability to use aliasis in the final, grand, yaml file; the changes here seem extremely extensive ... and writing variable names and things like such in actual python programs doesn't seem to take us is the right direction.

Dooruk:
As we discussed on Monday, this major effort looks good for me and thanks again for leaving observations yaml alone. I think the iteration method is probably the most obscure feature of SWELL so I'm glad this addresses that.

I will mention one thing here that I'm hoping now would be easier with the new approach. For most users, changes will happen on the observations section adding new obs types or filtering using different YAML configurations. For a select few users developing different products, there may be different flavors of model yaml entries for GEOS-FP, reanalyis, and S2S. That is something we can address in a future PR though just letting you all know.

I won't have time until next week, but one comparison for peace of mind would be to run 3dvar and 3dvar_atmos suites with the develop and this PR approach and compare to make sure we get the same OOPS log outputs. I would like to give this PR a try anyway so I can tackle that.

There are few other PRs that depend on some JEDI build updates, I would like them to be merged first. However there are some changes happening on the JEDI side, in particular how the bundles are built.

metdyn:
Hi All,

MPAS-Workflow has been extensively used at NCAR DA group, managing MPAS-model input and output, and MPAS-JEDI work with cylc. It has been able to generate nightly regression and weekly cycling since a few years.  Its structure configuration is similar to swell but more explicit.  In addition, the yaml files with anchor and dereference are extensively used, and there is no python replacement for yaml at file input levels, e.g., https://github.com/NCAR/MPAS-Workflow/blob/develop/config/jedi/applications/3denvar.yaml.
Since we are thinking of improving or replacing the iteration method, that is another reference that is worth to consider.

mranst:
> I won't have time until next week, but one comparison for peace of mind would be to run 3dvar and 3dvar_atmos suites with the develop and this PR approach and compare to make sure we get the same OOPS log outputs. I would like to give this PR a try anyway so I can tackle that.

Sounds good, I tested `3dvar` and `3dvar_atmos`, and the residuals norms for each matched between the two versions

mranst:
I added the use of ruamel for interpreting obs files, now anchors will not be expanded in the final config yaml. I also added a config key `check_for_obs`, if you set this to false with `generate_yaml_and_exit` set to true, you can test creating the config in `RunJediExecutable` tasks without running any tasks beforehand

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry.py
```diff
@@ -0,0 +1,46 @@
...
+        'fms initialization': {
+            'namelist filename': './fv3-jedi/fv3files/fmsmpp.nml',
+            'field table filename': './fv3-jedi/fv3files/field_table_gmao'
+        },
+        'akbk': f'./fv3-jedi/fv3files/akbk{vertical_resolution}.nc4',
+        'layout': [npx_proc, npy_proc],
+        'npx': horizontal_resolution,
+        'npy': horizontal_resolution,
+        'npz': vertical_resolution
+    }
```
ashiklom: The inconsistent typing on `horizontal/vertical_resolution` here is brittle and confusing.

Based on [this comment](https://github.com/GEOS-ESM/swell/pull/605#discussion_r2267974059), if there are some cases where horizontal resolution should be a string then that means it should _always_ be a string (at least in this context). So, I would remove the `int` conversion above and just let it be a string.

Strictly speaking, if it should be a precisely formatted string, a good way to validate that is to use regular expressions; e.g., something like:

```python
import re
###
respattern = re.compile(r'^ *(c)?(\d+)(x\d+)? *$')
# This allows values like: 72, 360, c720, 180x180
# We can customize however we want.
if not respattern.match(horizontal_resolution):
  raise ValueError(f"Invalid horizontal resolution: {horizontal_resolution}")
```

The "Parse, don't validate" approach would argue that things like this should be stored as class objects with precisely and specifically typed fields and precise and specific rendering methods (for templating). I'd like to get here eventually, but it's overkill and requires some overhaul of other code, too. Leaving this here as inspiration:

```python
class Resolution:
  def __init__(string: str):
    pat = re.compile(r'^ *(c)?(\d+)(x(\d+))? *$')
    m = pat.match(string)
    if not m:
      raise ValueError(r"Invalid resolution: {string}")
    grps = m.groups()
    self.cube = grps[0] == c  # boolean
    self.xres = int(grps[1])
    if grps[2]:
      self.yres = int(grps[3])
    else:
      self.yres = xres

  def __str__(self):
    return ("c" if self.cube else "") + f"{self.xres}x{self.yres}"
```

mranst:
@rtodling I've pushed some commits aimed at addressing point 1 in #568. The process that generates the observations section of the config yaml has been extracted from the jedi executables tasks and made into its own separate task. The task called `RenderJediObservations` now generates an intermediate `obs.yaml` before running the executable tasks. I've added a config key called `set_obs_as_local`, which will replace `cycle_dir` with `./` in the config yaml

Dooruk:
~@shiklomanov-an did you test this with uv installation by any chance? import ruamel didn't work for me but i will again.~
Doruk bug

shiklomanov-an:
I haven't tested this PR, but in my offline experiments, ruamel is a bit strange. The pip package name appears to be `ruamel-yaml`. When you install it that way, `import ruamel` works for me (though the more common import is `import ruamel.yaml`, to explicitly grab the YAML part).

Something like this should work:

```sh
uv add ruamel-yaml
# ...or, for the lower-level pip interface
uv pip install ruamel-yaml
```

Dooruk:
> I haven't tested this PR, but in my offline experiments, ruamel is a bit strange. The pip package name appears to be `ruamel-yaml`. When you install it that way, `import ruamel` works for me (though the more common import is `import ruamel.yaml`, to explicitly grab the YAML part).
> 
> Something like this should work:
> 
> ```shell
> uv add ruamel-yaml
> # ...or, for the lower-level pip interface
> uv pip install ruamel-yaml
> ```

Ok, I had to explicitly install `ruamel-yaml` to my `uv venv` and that works. I prefer activating the `uv venv` and running `swell create ...` rather than waiting for 6 seconds to load 150+ spack-stack modules..

`swell create` didn't require `ruamel` before that's why I didn't have this problem.

mranst:
> Ok, I had to explicitly install `ruamel-yaml` to my `uv venv` and that works. I prefer activating the `uv venv` and running `swell create ...` rather than waiting for 6 seconds to load 150+ spack-stack modules..
> 
> `swell create` didn't require `ruamel` before that's why I didn't have this problem.

I added it to the requirements for users not using spack-stack

src/swell/utilities/oops_config.py
```diff
@@ -0,0 +1,97 @@
...
+
+class OopsConfig(ABC):
+    """Abstract class that contains render information for JEDI oops configurations.
+
+    Attributes:
+    jedi_rendering: JediConfigRendering object
+    window_type: String of window dimensionality (e.g. '3D' or '4D')
+    cycle_time: ISO datetime (e.g. 20210701T120000Z)
+    cycle_dir: Location of cycle directory in the experiment
+    jedi_forecast_model: forecast model to be used by the experiment (e.g. pseudo-model)
```
mer-a-o:
```suggestion
    jedi_forecast_model: forecast model to be used by the experiment (e.g. pseudo_model)
```

src/swell/configuration/jedi/oops/hofx4D.py
```diff
@@ -0,0 +1,40 @@
...
+            },
+            'geometry': self.interface_model('geometry'),
+            'initial condition': self.interface_model('background'),
+            'observations': {
+                'get values': self.interface_model('getvalues'),
+                'observers': self.special_observations()
+            },
+        }
+
+        if self.window_type == '4D':
```
mer-a-o: since this is `hofx4D.py` do we need this if statement?

src/swell/configuration/jedi/oops/hofx4D.py
```diff
@@ -0,0 +1,40 @@
...
+            oops['model'] = self.interface_model(self.jedi_forecast_model)
+
+        oops['forecast length'] = self.template_dict['window_length']
```
mer-a-o: forecast length can be different from assimilation window length. For example you may want to run a 5 day forecast with 6h window length.

src/swell/configuration/jedi/oops/variational4D.py
```diff
@@ -0,0 +1,70 @@
...
+                'jb evaluation': False,
+                'time window': {
+                    'begin': self.template_dict['window_begin_iso'],
+                    'length': self.template_dict['window_length'],
+                    'bound to include': 'begin'
+                },
+                'geometry': self.interface_model('geometry'),
+                'model': self.interface_model('pseudo_model'),
+                'variable change': {'variable change name': 'Analysis2Model'},
+                'forecast length': self.template_dict['window_length'],
```
mer-a-o: `forecast length` and `window_length` can be different. I suggest using something line `forecast_length` instead.

src/swell/configuration/jedi/oops/variational4D.py
```diff
@@ -0,0 +1,70 @@
...
+            'variational': {
+                'minimizer': {
+                    'algorithm': self.template_dict['minimizer']
+                },
+                'iterations': [{
+                    'geometry': self.interface_model('geometry_inner'),
+                    'gradient norm reduction': float(self.template_dict['gradient_norm_reduction']),
+                    'ninner': str(self.template_dict['number_of_iterations']),
+                    'linear model': {
+                        'name': 'Identity',
```
mer-a-o, line 47: I'm just curious, is there any examples of a 4DVar experiment in SWELL?
If I want to change the linear model from `Identity` to `FV3JEDITLM` in my experiment ([example](https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/4dvar_geos_cf.yaml#L104-L111)) what would be the cleanest solution?

src/swell/tasks/run_jedi_hofx_executable.py
```diff
@@ -37,6 +37,7 @@ def execute(self, ensemble_members: Optional[list] = None) -> None:
         window_length = self.config.window_length()
         background_time_offset = self.config.background_time_offset()
         observations = self.config.observations()
+        background_time_offset = self.config.background_time_offset()
```
mer-a-o: probably added by mistake.
```suggestion

```

src/swell/tasks/run_jedi_hofx_ensemble_executable.py
```diff
@@ -40,7 +38,6 @@ def execute(self) -> None:
         window_type = self.config.window_type()
         window_length = self.config.window_length()
         background_time_offset = self.config.background_time_offset()
-        observations = self.config.observations()
```
mer-a-o, line 43: I'm not familiar with hofx_ensemble application. Is removing of this line itentional?

mer-a-o (commented):
Thank you @mranst. I added a few comments and questions. I will test this branch later this week to better understand the changes.

src/swell/tasks/run_jedi_hofx_ensemble_executable.py
mranst: `observations` is no longer being used since the observation portion of the yaml is being generated in a different task now

src/swell/configuration/jedi/oops/variational4D.py
mranst: I believe `3dfgat_atmos` uses the `variational4D` config. I think the best way would be to replace that with a template, and have logic in the `run_jedi_variational_executable` task that sets it to whatever value you want. I'm not sure if that can be based on an existing key (e.g. window_type), or you can maybe add a key in experiment.yaml that's used to set it

src/swell/configuration/jedi/oops/hofx4D.py
mranst: I adopted this logic from the way swell handles things currently, but yeah it should be safe to remove, I went ahead and did that

src/swell/configuration/jedi/oops/hofx4D.py
mranst: https://github.com/GEOS-ESM/swell/blob/b6f72662e5d005d17319fae4fb3f4e0494a0d1cc/src/swell/configuration/jedi/oops/hofx4D.yaml#L20-L21

This is how it's being done in Swell currently. We could add a key for this, either set manually or automatically in the task. Any thoughts @Dooruk?

src/swell/configuration/jedi/oops/LocalEnsembleDA.py
```diff
@@ -0,0 +1,69 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
```
mer-a-o, line 1: I noticed that you didn't remove `LocalEnsembleDA.yaml`

src/swell/configuration/jedi/oops/variational4D.py
mer-a-o: Right, `3dfgat_atmos` experiment uses `cost type: 4D-Var`. There's a `cost type: 3D-FGAT` available in JEDI, I wonder why that hasn't been used for this experiment. But I guess that is not related to this PR.

src/swell/configuration/jedi/oops/hofx4D.py
mer-a-o: I think user would want to be able to set that.

mer-a-o (commented):
I tested some experiments with this branch and everything works fine. Thanks @mranst.

src/swell/configuration/jedi/oops/LocalEnsembleDA.py
mranst: Left some in by mistake, thanks

rtodling:
@mer-a-o The point about "forecast length" is a good, however it is a bit of a moot point right now since we don't use a single executable between jedi and geos-gcm. We can add the distinction between the two in the swell scripts, but all default should make it equal to "window length".

src/swell/tasks/render_jedi_observations.py
```diff
@@ -0,0 +1,91 @@
...
+        # Set fields for obs files
+        self.jedi_rendering.add_key('window_begin', window_begin)
+        self.jedi_rendering.add_key('background_time', background_time)
+        self.jedi_rendering.add_key('crtm_coeff_dir', crtm_coeff_dir)
+        self.jedi_rendering.add_key('marine_models', marine_models)
+
+        cwd = os.getcwd()
+
+        # Replace cycle_dir with './' if specified
+        if self.config.set_obs_as_local(False):
```
Dooruk, line 55: Is the purpose for `set_obs_as_local` is to create `obs.yaml` locally while testing with `swell task ...` during testing? Sorry, it wasn't clear to me.

src/swell/configuration/jedi/oops/hofx4D.py
Dooruk: @rtodling already responded to this. But so far we only have been running in `PseudoModel` mode. For marine I don't use 4D, SOCA doesn't have a linear code. I do run FGAT but that is a 3D cost function..

I'm not aware if you were able to run GEOS directly within JEDI. I think `in-core` is the right term?

src/swell/tasks/render_jedi_observations.py
mranst: This was requested as point 1c in #568, I think it has to do with making it work with GEOS-JEDI, but I'm not exactly sure what the particulars are

Dooruk:
You already verified the tier 1 tests work and I confirm that. All looks good to me thank you for these changes. I can approve it once the comments are addressed.  It is nice to do away with the iterator method and keeping the order of YAMLs.

With these changes we will be using two different YAML packages, ruamel and pyyaml. Is there a downside to this? `import yaml` is all over the code so that wouldn't be trivial but we could make the change to ruamel gradually.

src/swell/tasks/render_jedi_observations.py
Dooruk: Ah ok, @rtodling FYI

mranst:
> You already verified the tier 1 tests work and I confirm that. All looks good to me thank you for these changes. I can approve it once the comments are addressed. It is nice to do away with the iterator method and keeping the order of YAMLs.
> 
> With these changes we will be using two different YAML packages, ruamel and pyyaml. Is there a downside to this? `import yaml` is all over the code so that wouldn't be trivial but we could make the change to ruamel gradually.

Thanks!

I don't believe there's any reason to keep using pyyaml over ruamel

src/swell/configuration/jedi/oops/hofx4D.py
mranst: Pushed a commit that distinguishes `forecast_length` from `window_length` in the tasks. It will be set to the value of `window_length` currently, but we can add a question for that later when we want to distinguish it

mer-a-o: @Dooruk Correct, we can't run GEOS `in-core`. In your cycling experiments at the forecast step, how do you set the forecast length? Is it hard-coded?

mer-a-o (dismissed):

shiklomanov-an:
> I don't believe there's any reason to keep using pyyaml over ruamel

I enthusiastically endorse switching from py-yaml to ruamel! Among other things, [pyyaml works on YAML 1.1](https://yaml.dev/doc/ruamel.yaml/pyyaml/#top), which has [all kinds of issues](https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell) that have mostly been addressed in YAML 1.2 (the default version used by ruamel).

Dooruk (dismissed):

src/swell/configuration/jedi/oops/hofx4D.py
Dooruk: I set it with the `forecast_length` key in `experiment.yaml`. Then I rewrite `CAP.rc` with it like so:

https://github.com/GEOS-ESM/swell/blob/c6e065f82f230f475b26c45f259876da6716428b/src/swell/tasks/prep_geos_run_dir.py#L500

Dooruk:
This is good to go in for me, I think @shiklomanov-an has some unaddressed comments?

src/swell/configuration/jedi/oops/hofx4D.py
mer-a-o: Thanks. You're using `forecast_duration` here. So perhaps, in a different PR, we can unify these names.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/geometry.py
mranst: The int type casting appears to not be an issue for jedi, I've removed it for this file

mranst:
Thanks everyone. I think this should be okay to go, with the exception of the generic anchor names in the observation files that we've been discussing in #558.

Dooruk (approved):
All good!

PR #610, opened 2025-08-20
Adding cloud seeding option to all-sky radiance yaml files

gmao-yzhu:
The current all-sky microwave yaml files don't have cloud seeding option. UFO codes have been evolving. Based on the July UFO version, it is seen that the cloud seeding option is not needed when geovals are produced from GSI runs, but this option is needed for all-sky radiances when geovals are generated on the fly.  GMI and AMSR2 are used in all-sky conditions in GEOS, so this PR is adding this option to their yaml files.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsr2_gcom-w1.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/amsr2_gcom-w1.yaml
@@ -19,6 +19,7 @@ obs operator:
   Absorbers: [H2O,O3,CO2]
   Clouds: [Water, Ice, Rain, Snow]
   Cloud_Fraction: 1.0
+  Cloud_Seeding: true  
   obs options:
     Sensor_ID: &Sensor_ID amsr2_gcom-w1
     EndianType: little_endian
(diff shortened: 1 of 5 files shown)


Dooruk (dismissed):
Thanks!

mjkagnes123 (approved):
Thank you, Yanqiu!

gmao-wgu:
@gmao-yzhu Observations from MHS are also currently being assimilated under all-sky conditions in GEOS. Do their YAML files need to be updated to reflect this change?

gmao-yzhu:
Thanks Min-Jeong. I will add MHS.

gmao-yzhu:
Thanks Wei. I will add MHS...        Now Done.

gmao-wgu (approved):

rtodling:
Since nobody provided a picture of what this change amounts to I will provide some.

Taking a GMI-only analysis as test-case.

The is the difference in PS-inc for two JEDI runs  (the incs are multiplied by a factor of 5 - since PS inc due to GMI is small) - the top ithe difference of the no-seeding opt w/ GSI and the bottom is the diff of the two JEDI runs:

ps

Here is the similar set of plots for T-skin increment.

ts

and another one for 850-mb T (now multiplied by a factor of 20!):

t850

Yes - it is as it seems; the differences between having and not having cloud-seeding a nearly irrelevant to the results ... now the question is whether this is an expected result or not. I suppose we could say that the match w/ GSI is already so good that it truly doesn't matter. 

Just to give a final sense of the results between GSI and JEDI here is a temperature increment at 850 w/o seeding  - w/ incs blown up by a factor of 20 ... pretty decent match:

t850

However, there are small differences one would have hoped something like the cloud seeding would have addressed ... but that is not the case.

**PLEASE**: next time, attach some figures and an evaluation of how results change with the added feature.

PR #611, opened 2025-08-25
Removing unnecessary parameters from the NSST filter

gmao-wgu:
## Description

*Both obserr_demisf and obserr_dtempf are not parameters related to the NSST filter and should be removed*


rtodling:
@gmao-wgu I'm sorry, I am confused. I thought you had been saying that nothing was wrong in the IR settings!  After I pointed out that some of these filters are doing nothing you are then agreeing w/ me that something are not quite there yet? 

Are your changes zero diff?  Can you pls indicate this  (yes or no) is the prologue  of this PR?

If not zero diff, we'll need to see some figs please.

gmao-wgu:
> @gmao-wgu I'm sorry, I am confused. I thought you had been saying that nothing was wrong in the IR settings! After I pointed out that some of these filters are doing nothing you are then agreeing w/ me that something are not quite there yet?
> 
> Are your changes zero diff? Can you pls indicate this (yes or no) is the prologue of this PR?
> 
> If not zero diff, we'll need to see some figs please.

@rtodling I might not have made myself clear.  This NSST filter works as expected, except that these two parameters are unnecessary. However, even if you include them, the filter still functions correctly. Here’s a comparison before and after:

https://gmao.gsfc.nasa.gov/gmaoftp/wgu/Qc_correct

https://gmao.gsfc.nasa.gov/gmaoftp/wgu/Qc_correct_modified

They are identical.

src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/cris-fsr_n20.yaml
```diff
@@ -244,18 +244,18 @@ obs post filters:
...
 # Surface Temperature Jacobian Check over Land
-- filter: Bounds Check
-  filter variables:
-  - name: brightnessTemperature
-    channels: *cris-fsr_n20_channels
-  where:
-  - variable:
-      name: GeoVaLs/water_area_fraction
-    maxvalue: 0.99
+#- filter: Bounds Check
```
rtodling, line 247: @gmao-wgu why are these lines now commented out? Is this related to the changes associated with getting things to be in sync w/ changes in x0051? Same question for the other cris instruments.

rtodling (approved):
I am fine with these changes - they are consistent w/ the near SST filter.

src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/cris-fsr_n20.yaml
gmao-wgu: @rtodling  This filter removes observations from CrIS-FSR over non-water surfaces when the sensitivity with respect to Tskin >=0.2.  Since observations over non-water surfaces are not currently assimilated in GEOS, so this filter has not impact on the results.

Dooruk:
@gmao-wgu was this related to an open issue in Swell?

PR #613, opened 2025-08-27
(Bugfix) Check for string None in stage_jedi

### LINKED ISSUE #612: Tier1 tests failing
Latest tier1 tests are failing on two different tasks. First one is StageJedi:

Image

I checked the experiment.yaml (`/discover/nobackup/gmao_ci/swell/tier1/17268814774/hofx_FAILED/swell-hofx-17268814774/swell-hofx-17268814774-suite`) and no user static files were defined. It is set to `None` so this looks like something to do with use of `None` in yaml, I think.

The second error is the similar one regarding  "missing" soca executable that fails GenerataBClimatology. @jardizzo already restarted the runners so I'm not sure why this is happening.

mranst:
## Description

#612 Also checks for string = None in stage_jedi

### DIFF
--- a/src/swell/tasks/generate_b_climatology.py
+++ b/src/swell/tasks/generate_b_climatology.py
@@ -244,7 +244,7 @@ class GenerateBClimatology(taskBase):
 
         # Use static_files_user if present in config and contains files
         # -------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Dooruk (approved):
Thanks!

Issue #318, opened 2024-03-29
R2D2_v3 Adaptation and Observation Database Development


Dooruk:
This issue will evolve over time and there will be new steps/issues introduced.
**Feel free to add and/or suggest edits any time. Most of these tasks will involve other GMAO folks.**

# Which files need to be indexed by R2D2 v3: 

These files are currently located in `/discover/nobackup/projects/gmao/advda/R2D2DataStore/Shared`:

### 1) Observations (IODAv3 format also BUFR format for the future):

⚠  Currently, all observations are organized in `PT6H` windows.

  - For `geos_atmosphere` we are using `x0050`:
    - [x] **ncdiag**: These are files which were created by running `convert_ncdiags` suite. 

For atmosphere, at the moment we rely upon previous x00* experiment runs, i.e. phase 1. See Phase 1 and other phases in figure at the bottom of this issue.

   - For `geos_marine`, there are two sources, **the tiered CI tests only use a subset** of these:
     - [x] **odas**: These are local GMAO processed files (**O**cean **DA** **S**ystem)
     - [x] **gdas_marine**: These are the files that are transferred from NOAA/EMC.

### 2)  Bias Coefficients (BCs) (only relevant for `geos_atmosphere`):
  - [ ] **gsi**: These are files which were created by running `convert_ncdiags` suite and `x0050` experiment. Has five different file types:
    - tlapse (txt)
    - satbias (nc4) (Satellite Bias)
    - satbias_cov (nc4) (Satellite Bias)
    - acftbias (nc4) (Aircraft)
    - acftbias_cov (nc4) (Aircraft)

### 3) Experiment Background (only relevant for `geos_marine`):
`geos_atmosphere` uses `GetBackgroundGeosExperiment` task now, used to utilize `geos` folder. 
`geos_marine` utilizes `mom6_cice6_UFS` folder.
### 4) [currently inactive] Ensemble files:
 (currently **none** for either model. `geos_atmosphere` uses `GetEnsembleGeosExperiment`)

---
# Needs from R2D2: 

### Immediate term:

- [x] Indexing `geos_atmosphere` and `geos_marine` `IODAv3` observations that SWELL utilizes within the new R2D2_v3 API context and testing the `search` functionality. 

(dated `20231010T000000Z`, `20231010T060000Z` for `geos_atmosphere`) 
(dated `20210701`, `20210702` for `geos_marine`) 

- [ ] Implementing the new R2D2_v3 API within Swell

### Short-term:
[...]

PR #618, opened 2025-08-28
R2D2 v3 Adaptation 

ftgoktas:
This PR migrates SWELL from r2d2 v1 to r2d2 v3 API for centralized metadata management.
Updates GetObservations, GetBackground, and SaveObsDiags tasks to use new API syntax. 
Adresses #318

**Key Changes:**
- Updated module loading to use `r2d2-client/sles15_0604` instead of `r2d2/sles15_spack19`
- Modified r2d2 configuration to use v3 API parameters (`data_hub`, `data_store`, `compute_host`)
- Updated SWELL tasks to use new r2d2 v3 API syntax:
  - `GetObservations`: Updated fetch parameters (`item='observation'`, `window_start`, etc.)
  - `GetBackground`: Updated for forecast item fetching
  - `SaveObsDiags`: Updated to store as `item='feedback'` with proper parameters
- Added environment variable handling for r2d2 v3 authentication (`R2D2_USER`, `R2D2_API_KEY`)
- Fixed module loading conflicts between JEDI bundles and r2d2 v3
- Created a script for registering observation, background, and bias correction files with r2d2 v3 schema

**Tested with 3dvar and 3dvar_atmos suites** - successfully fetches observations/backgrounds and stores feedback files.

## Dependencies

- r2d2 v3 infrastructure (nccs-gmao data hub, r2d2-experiments-nccs-gmao data store)
- r2d2-client/sles15_0604 module availability on target platform


r2d2_v3_setup.sh
mranst, line 1: Are these shell scripts meant to be run externally? I don't see calls to them from within swell

src/swell/utilities/scripts/ingest_files.py
mranst, line 1: We may want to consider adding this functionality as a basic suite within swell, or maybe it could be a command to the main swell frontend

scripts/register_files.py
```diff
@@ -0,0 +1,259 @@
...
+# Color codes for printing
+RED = "\033[91m"
+GREEN = "\033[92m"
+YELLOW = "\033[93m"
+BLUE = "\033[94m"
+RESET = "\033[0m"
+
+try:
+    import r2d2
+except ImportError as e:
```
mranst: I like the helpful message. You can also handle this by raising an error with a custom message, which provides us with a more consistent way of handling issues
```suggestion
try:
    import r2d2
except ImportError:
    raise ImportError(f'Failed to import r2d2, use \nLoad module: module load r2d2-client/sles15_0604')
```

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -39,13 +39,46 @@ module load other/mepo
...
+
+###################################################
+# R2D2 v3 Environment Variables
+# -----------------------------
+# Source credentials from file if it exists
+if [ -f ~/.swell_r2d2_credentials.sh ]; then
+    source ~/.swell_r2d2_credentials.sh
+fi
+
+# Check if required variables are set and provide helpful error if not
```
mranst: I would personally prefer handling this kind of thing from within python using `os.environ` (others may disagree). Trying to source the the modules file without the swell credentials file caused me to get exited from discover. This could potentially be a swell wrapper around `r2d2`, or a function within the `TaskBase` class that sets the environment variables. We have a directory that we use frequently for config files under `~/.swell`, you may want to consider putting this file there. Instructions for this file should also be added in docs

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -39,13 +39,46 @@ module load other/mepo
...
+    exit 1
+fi
+####################################################
 
 # Load eva and jedi_bundle
 # ------------------------
 module load eva/sles15_milan_1.6.5
 module load jedi_bundle/sles15_skylab9
 
+
```
mranst: Is jedi_bundle loading a version of r2d2?

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -57,3 +90,10 @@ ulimit -S -s unlimited
 ulimit -S -v unlimited
 umask 022
 
+#########
```
mranst: leftover comments

src/swell/tasks/.save_obs_diags.py.swn
mranst, line 1: remove swapfiles

src/swell/tasks/get_background.py
```diff
@@ -143,21 +143,35 @@ def execute(self) -> None:
...
-                    model=r2d2_model_dict[model_component],
-                    file_type=file_type,
-                    fc_date_rendering='analysis',
-                    step=bkg_step,
-                    resolution=horizontal_resolution,
-                    type='fc',
-                    experiment=background_experiment)
+                file_extension=file_type.split('.')[-1] if '.' in file_type else 'nc'
+
+                print("\n\n************************************\n")
```
mranst: In swell we generally prefer to handle this kind of messaging with `self.logger.info(` so we can potentially filter them with messaging levels

src/swell/tasks/get_observation_with_logging.py
mranst, line 1: Is this task being used currently?

src/swell/tasks/get_observations.py
```diff
@@ -231,22 +242,26 @@ def execute(self) -> None:
...
-                          type='bc',
-                          experiment=obs_experiment,
-                          file_type=bias_file_type)
+                    r2d2.fetch(
+                        item='bias_correction',
+                        provider='gsi',
+                        observation_type=observation,
+                        file_extension=bias_file_type.split('.')[-1] if '.' in bias_file_type else bias_file_type,
+                        window_start=background_time.strftime('%Y-%m-%dT%H:%M:%SZ'),
+                        window_length='PT6H',
```
mranst: I imagine we don't want the value for `window_length` to be hard-coded, but I'm not sure what the appropriate config value would be, may need to add it

src/swell/tasks/save_obs_diags.py
```diff
@@ -57,29 +58,78 @@ def execute(self) -> None:
...
+                    observation_type=name,
+                    file_extension=obs_path_file.split('.')[-1],
+                    window_length='PT6H',
+                    window_start=window_begin,
+                    source_file=obs_path_file,
+                    member=-9999,
+                )
+                self.logger.info(f'Successfully stored feedback file for {observation}')
+                
+            except Exception as e:
```
mranst, line 128: If there's a specific error type you're expecting to encounter frequently that's not an issue to the run, I'd try to catch it specifically. We don't necessarily want to let any kind of error slip through

src/swell/tasks/task_questions.py
```diff
@@ -399,6 +399,22 @@ class TaskQuestions(QuestionContainer, Enum):
 
     # --------------------------------------------------------------------------------------------------
 
+    GetObservationsR2d2 = QuestionList(
```
mranst: Looks like this may be left over from a previous task

src/swell/tasks/save_obs_diags.py
```diff
@@ -37,10 +37,11 @@ def execute(self) -> None:
         self.jedi_rendering.add_key('marine_models', self.config.marine_models(None))
 
         # Get window beginning
-        window_begin = self.da_window_params.window_begin(window_offset)
+        #window_begin = self.da_window_params.window_begin(window_offset) #dto
```
mranst: I would reset this section to avoid adding to the git blame unnecessarily

src/swell/tasks/save_obs_diags.py
```diff
@@ -57,29 +58,78 @@ def execute(self) -> None:
...
+                else:
+                    self.logger.info(f'  File does not exist!')
+                    
+                self.logger.info(f'  GetObservations task did not fetch this file successfully')
+                self.logger.info(f'  Skipping {observation}')
                 continue
 
-            # Store observation files
-            # -----------------------
+            # Store diagnostic observation files (OUTPUT from RunJediVariationalExecutable)
```
mranst: This task is used after other jedi executables as well (hofx, localensembleda, etc)

src/swell/tasks/get_observations.py
```diff
@@ -154,14 +154,24 @@ def execute(self) -> None:
...
+                        'window_start': obs_window_begin,         #'2023-10-09T21:00:00Z',   # From filename timestamp
+                        'window_length': obs_window_length,       #'PT6H',                   # From filename
+                        'target_file': target_file                #'./fetched_gmi_gpm.nc4'   # Where to save
+                    }
+                
+                    print(f"Searching for file with criteria: {fetch_criteria}")
+                    try:            
+                        r2d2.fetch(**fetch_criteria)
+                        self.logger.info(f"Successfully fetched {target_file}")
+                    except Exceotion as e:
```
mranst:
```suggestion
                    except Exception as e:
```

r2d2_v3_setup.sh
```diff
@@ -0,0 +1,24 @@
...
+
+unset R2D2_SERVER_HOST
+unset R2D2_SERVER_PORT
+
+export R2D2_USER=
+export R2D2_API_KEY=
+export R2D2_HOST=discover-gmao
+export R2D2_COMPILER=intel
+
+source venv_client/bin/activate
```
mranst: Need a different way of loading the environment if these need to be in swell

mranst (dismissed):
Nice work- It seems like r2d2 v3 requires a lot of environment variables to be set in order to run. @ashiklom or @Dooruk may have a better idea, but maybe a way to handle this would be to have a config file called `~/.swell/r2d2_credentials.yaml` where the user could store this information. There could then be a function within `TaskBase` that parses and sets these variables using `os.environ` (you can see a somewhat similar example to this in `slurm.py`, where we set slurm defaults in `~/.swell/swell-slurm.yaml`). We're also going to need some documentation explaining this, I'm having trouble getting it running with all the necessary environment additions

src/swell/utilities/scripts/ingest_files.py
Dooruk: I like the idea of having this functionality as a suite though it won't be trivial (nothing is in the DA world it seems!). Registering files would be only allowed to a certain subgroup of users. Ideally one or two obs. lead from each group/product (atm, ocn, aero, bgc. etc). Most index values will require model specific entries (e.g, horizontal_resolution for background, provider, step etc.). Also we would be using different file formats and so many naming conventions (BUFR, nc4, zarr?).

Anyways, we can bring this up with our meeting with @jeromebarre and @mer-a-o

r2d2_v3_setup.sh
Dooruk: agreed, we should have a .r2d2* file at $HOME to handle this like Michael suggested.

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -39,13 +39,46 @@ module load other/mepo
 module use -a /discover/nobackup/projects/gmao/advda/JediOpt/modulefiles/core
 module load solo/sles15_skylab9
 module load py-boto3
-module load r2d2/sles15_spack19
+# Ensure to unload old r2d2 v2 modules
+module unload r2d2 2>/dev/null || true
+module load r2d2-client/sles15_0604
```
Dooruk: So I see two different r2d2 installations, r2d2 and r2d2-client, are both of them necessary?

Since R2D2 uses spack-stack modules, the overall approach shouldn't change, other than setting environment varialbes or API key.

src/swell/deployment/platforms/nccs_discover_sles15/modules
Dooruk, line 1: same change needs to happen on nccs_discover_cascade

src/swell/tasks/get_background.py
Dooruk: Agreed

Dooruk (changes requested):
@ftgoktas great job! finally v3 is becoming a reality.

There are Swell specific peculiarities, @mranst pointed out most of them. Also there are built-in functions like logger so you could use them. We should store R2D2 related config and environment variables inside a file like `~/.swell/r2d2-config.yaml`.

 A few additional notes and Q.s:
 
- There is a little bit of separation happening here with the atmosphere side accessing obs file outside of R2D2 context, which is fine during the development.
 
- Once we have something like `~/.swell/r2d2-config.yaml` we need to have a dedicated section in the documentation, maybe pointing at JCSDA-internal/R2D2 documentation and with useful commands.

- Is it possible to request API keys for @mranst and @ashiklom? Or is it possible to request a generic "gmao-user" API key and have special ones for sudo type users? That way we can create a generic `~/.swell/r2d2-config.yaml` if none exists, otherwise I imagine %98 users will be flabbergasted and message one of us without looking at the documentation.

- I probably asked this before but Is linking rather than copying an option with `fetch`?

- (Long term) Let's keep in mind we would like to use R2D2 on AWS or NAS and able to access cloud dataset (e.g., NNJA) with it.

src/swell/tasks/task_questions.py
ftgoktas: Yes, this was for testing. Removing it.

r2d2_v3_setup.sh
ftgoktas: The individual shell scripts are for loading R2D2 to test without running Swell. Included these under /scripts since they are helpful when testing R2D2 file registration and fetching. Swell itself loads R2D2 in modules and doesn't have to use them.

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -39,13 +39,46 @@ module load other/mepo
 module use -a /discover/nobackup/projects/gmao/advda/JediOpt/modulefiles/core
 module load solo/sles15_skylab9
```
Dooruk: Need to eliminate py-boto3 and solo

r2d2_credentials.yaml
```diff
@@ -0,0 +1,13 @@
+# R2D2 v3 credentials file
+# Save this as ~/.swell/r2d2_credentials.yaml
+# Set permissions: chmod 600 ~/.swell/r2d2_credentials.yaml
+
+# R2D2 v3 API credentials
+user: your_username
+api_key: your_api_key_here
+
+# Platform configuration  
+host: discover-gmao
```
Dooruk, line 10: This should be platform specific (which there is already a way to do this in `utilities/r2d2.py`)

src/swell/deployment/platforms/nccs_discover_sles15/modules
ftgoktas: The credentials are now loaded from within Python using `os.environ` in the `load_r2d2_credentials()` function, which is called from `create_r2d2_config()` in the `r2d2` utilities module. The credentials file should be located at `~/.swell/r2d2_credentials.yaml`, and the function safely handles missing files by falling back to existing environment variables, so there's no risk of getting kicked from Discover. I removed all the shell script sourcing from the modules file to avoid those exit issues.

src/swell/deployment/platforms/nccs_discover_sles15/modules
ftgoktas: No, jedi_bundle does not include r2d2. Testing also shows "no module named r2d2" when only jedi_bundle is loaded.

src/swell/tasks/get_observation_with_logging.py
Dooruk, line 1: do we need this file?

src/swell/deployment/platforms/nccs_discover_sles15/r2d2_config.yaml
Dooruk, line 1: Make the same change in `src/swell/deployment/platforms/nccs_discover_cascade/r2d2_config.yaml`

Dooruk:
Well, `3dvar` is working for me, I was able to use my API key. Finally, R2D2v3 is here, this is exciting 🤩 

One minor comment. If you look at the log output of `GetObservations` task, it is less "readable":

For instance, mine is here:
`~/cylc-run/new_r2d2-suite/runN/log/job/20210701T1200Z/GetObservations-geos_marine/NN/job.out`


Screenshot 2025-09-24 at 4 13 33 PM

Currently it looks more readable like so:


Screenshot 2025-09-24 at 4 16 07 PM


Could it be because we are calling r2d2.fetch rather than:

```python
from r2d2 import fetch



fetch(...)
```
?

ftgoktas:
> Well, `3dvar` is working for me, I was able to use my API key. Finally, R2D2v3 is here, this is exciting 🤩
> 
> One minor comment. If you look at the log output of `GetObservations` task, it is less "readable":
> 
> For instance, mine is here: `~/cylc-run/new_r2d2-suite/runN/log/job/20210701T1200Z/GetObservations-geos_marine/NN/job.out`
> 
> Screenshot 2025-09-24 at 4 13 33 PM
> Currently it looks more readable like so:
> 
> Screenshot 2025-09-24 at 4 16 07 PM
> Could it be because we are calling r2d2.fetch rather than:
> 
> ```python
> from r2d2 import fetch
> 
> 
> 
> fetch(...)
> ```
> 
> ?

No, the import style (`import r2d2` vs `from r2d2 import fetch`) doesn't affect the logging output. The difference in logs is because R2D2 v3 has different internal logging compared to the older version, v3's logging is more technical by default.   We can control the logging by setting the `R2D2_LOG_LEVEL` environment variable to `SILENT`, `INFO`, `DEBUG`, or `TRACE` (default is `INFO`). Setting it to `SILENT` will reduce the technical output but to get back the readable format like the old version, we need to add our own custom logging before the `r2d2.fetch()` call.

For comparison:

`R2D2_LOG_LEVEL = “SILENT”`:
￼
Screenshot 2025-09-25 at 3 20 53 PM

`R2D2_LOG_LEVEL = “INFO”` (this is default):
￼
Screenshot 2025-09-25 at 3 27 07 PM

`R2D2_LOG_LEVEL = “DEBUG”`:
￼
Screenshot 2025-09-25 at 3 29 59 PM

`R2D2_LOG_LEVEL = “TRACE”`:

following

mranst:
3dvar works for me too, get_obs for 3dvar_atmos fails with this error:
image

ftgoktas:
> 3dvar works for me too, get_obs for 3dvar_atmos fails with this error: image

All 3dvar atmospheric bias coefficient files have been ingested to R2D2. It should be able to run without any errors now.

Dooruk:
Is this still a draft or ready for testing?

ftgoktas:
Can be tested.

Dooruk:
Ok, so far I received two errors:

1) This one seems more critical and not sure why am I the only one receiving it. This is what I got while running `3dvar_atmos`:
Screenshot 2025-10-27 at 2 04 04 PM

2) This one is a simple fix which is a problem in line 287 in `get_observations.py`, I received it during my `3dvar` run. IMO we don't need the self.logger.info outputs there anyway as r2d2 puts out relevant information:

Screenshot 2025-10-27 at 2 06 23 PM

mranst:
> Ok, so far I received two errors:
> 
>     1. This one seems more critical and not sure why am I the only one receiving it. This is what I got while running `3dvar_atmos`:
> 
> 
> Screenshot 2025-10-27 at 2 04 04 PM


I experienced this as well, maybe a credential issue?

ftgoktas:
Fixed read permissions on the bias_correction data store via gmaoswell credentials. Also updated logging. 
It should run without errors now. 

Screenshot 2025-10-27 at 4 39 47 PM

register_swell_3dvar_experiment.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
mer-a-o: Why there's a need to register only `swell-3dvar` experiment? and why for all the `lifetime` categories? 
IMO this code needs to live under `utilities` and can be made more generic so it can be used to register different experiment names. 

The way it's used in Skylab/Ewok is that experiment `name` is randomly generated when an experiment is created and then the experiment is [registered on R2D2 ](https://github.com/JCSDA-internal/ewok/blob/develop/src/ewok/bin/create_experiment.py#L39)with some [metadata](https://github.com/JCSDA-internal/ewok/blob/develop/src/ewok/r2d2_util.py#L38-L44). By default the `lifetime` is set to `debug` and the experiment (and the associated files) are removed from R2D2 after 2 weeks. 

In SWELL, since we are reusing the same experiment names (and same gmao users) we can register each experiment once with `release` lifetime.

src/swell/tasks/get_observations.py
```diff
@@ -154,14 +166,23 @@ def execute(self) -> None:
...
+                        'item': 'observation',               # Required for r2d2 v3
+                        'provider': obs_provider,            # What we registered with
+                        'observation_type': observation,     # From filename
+                        'file_extension': 'nc4',
+                        'window_start': obs_window_begin,    # From filename timestamp
+                        'window_length': obs_window_length,  # From filename
+                        'target_file': target_file,          # Where to save
+                    }
+
+                    print(f"Searching for file with criteria: {fetch_criteria}")
```
mer-a-o: Does R2D2 already print this information and handle the error?

src/swell/tasks/save_obs_diags.py
```diff
@@ -57,29 +57,78 @@ def execute(self) -> None:
...
+                        dataset = nc.Dataset(input_obs_file, 'r')
+                        dims = {dim_name: dim.size for dim_name, dim in dataset.dimensions.items()}
+                        self.logger.info(f'  File exists but dimensions: {dims}')
+                        dataset.close()
+                    except Exception as e:
+                        self.logger.info(f'  File exists but error reading: {str(e)}')
+                else:
+                    self.logger.info(f'  File does not exist!')
+
+                self.logger.info(f'  GetObservations task did not fetch this file successfully')
```
mer-a-o: There could be other reasons for the feedback (obs_diag) file to not exist. I suggest removing this line.

src/swell/tasks/save_obs_diags.py
```diff
@@ -57,29 +57,78 @@ def execute(self) -> None:
...
+            # Check if observation was used - this checks INPUT file exists and has data
+            input_obs_file = observation_dict['obs space']['obsdatain']['engine']['obsfile']
+            self.logger.info(f'Checking input observation file: {input_obs_file}')
+
             use_obs = check_obs(self.jedi_rendering.observing_system_records_path, observation,
                                 observation_dict, self.cycle_time_dto())
+
+            self.logger.info(f'Checking observation {observation}: use_obs = {use_obs}')
+
             if not use_obs:
```
mer-a-o, line 69: For my understanding, when is this if statement true? Don't we do this check in line 64 with `check_obs`?

src/swell/utilities/scripts/ingest_files.py
mer-a-o: Having ingest suite can make storing obs files on R2D2 very convenient. I'd be happy to help with setting it up.

register_swell_3dvar_experiment.py
ftgoktas: This file should already be removed in the latest version; could you check if your branch is up to date?

src/swell/tasks/get_observations.py
ftgoktas: Correct, removing this.

src/swell/tasks/save_obs_diags.py
ftgoktas: check_obs returns False (so the if statement is True) when the input obs isn’t usable for this cycle. It can be a missing/empty file, a zero location dimension etc.

src/swell/utilities/scripts/ingest_files.py
ftgoktas: Created #646  for this.

src/swell/tasks/get_observations.py
ftgoktas: Will address this here #647

src/swell/tasks/get_observation_with_logging.py
ftgoktas: Removed

src/swell/utilities/scripts/ingest_files.py
Dooruk: > Having ingest suite can make storing obs files on R2D2 very convenient. I'd be happy to help with setting it up.

thanks @mer-a-o !

src/swell/tasks/save_obs_diags.py
mer-a-o: Thanks @ftgoktas. Would this make the if statement in line 72 redudnet? You already check the file dimension with `check_obs`.

src/swell/deployment/platforms/nccs_discover_cascade/r2d2_config.yaml
```diff
@@ -1,22 +1,21 @@
...
 
-# when storing data, in which order should the databases accessed?
-store_order:
-  - ${USER}
+# Optional: credentials file for v3 API authentication
+# credentials_file: ~/.swell/swell-r2d2-credentials.yaml
 
+# Cache settings
 cache_name: ${USER}
+cache_fetch: false
```
Dooruk, line 21: What is the purpose of `cache_fetch` (just wondering)

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -37,15 +37,28 @@ module load other/mepo
...
+
+# Force r2d2 v3 to be first in PYTHONPATH (override any JEDI r2d2 v2)
+# -------------------------------------------------------------------
+R2D2_V3_PATH="/discover/nobackup/projects/gmao/advda/JediOpt/core/r2d2-client/sles15_0604/lib/python3.11/site-packages"
+export PYTHONPATH="${R2D2_V3_PATH}:$PYTHONPATH"
+
+
+# Remove any r2d2 v2 paths from PYTHONPATH to prevent conflicts
+export PYTHONPATH=$(echo $PYTHONPATH | tr ':' '\n' | grep -v "/r2d2/sles15_spack19/" | tr '\n' ':' | sed 's/:$//')
+
```
Dooruk:
```suggestion

```

I really don't think these are necessary.

src/swell/deployment/platforms/nccs_discover_sles15/modules
```diff
@@ -37,15 +37,28 @@ module load other/mepo
 # Load r2d2 modules
 # -----------------
 module use -a /discover/nobackup/projects/gmao/advda/JediOpt/modulefiles/core
-module load solo/sles15_skylab9
-module load py-boto3
-module load r2d2/sles15_spack19
+
+# Ensure to unload old r2d2 v2 modules
+module unload r2d2 2>/dev/null || true
```
Dooruk:
```suggestion

```

I really don't think these are necessary.

src/swell/deployment/platforms/nccs_discover_sles15/r2d2_config.yaml
```diff
@@ -1,22 +1,23 @@
-databases:
+# TODO
```
Dooruk:
```suggestion
```
you already did it :)

src/swell/tasks/base/task_base.py
```diff
@@ -278,11 +278,22 @@ def create_task(
...
+        # Load R2D2 credentials before importing any task modules
+        # -------------------------------------------------------
+        from swell.utilities.config import Config
+        from swell.utilities.r2d2 import load_r2d2_credentials
+        from swell.utilities.logger import get_logger
+
+        # Get platform info from config to load credentials
+        temp_logger = get_logger('R2D2Setup')
+        temp_config = Config(config, temp_logger, task, model)
+        load_r2d2_credentials(temp_logger, temp_config.__platform__)
```
Dooruk, line 290: Could you remind me why this part has to be in the base class and not handled only in R2D2 specific tasks?

src/swell/tasks/get_background.py
```diff
@@ -13,14 +13,13 @@
...
-from r2d2 import fetch
-
+import r2d2
 
 # --------------------------------------------------------------------------------------------------
 
 r2d2_model_dict = {
     'geos_atmosphere': 'geos',
-    'geos_marine': 'mom6_cice6_UFS',
+    'geos_marine': 'mom6',  # 'mom6_cice6_UFS'
```
Dooruk, line 22: One thing for marine is that we deal with different types of models, such as MOM6, CICE6, or WAVEWATCH3 ( even a biogeochemical model) in the future. So perhaps this could be a tuple and depending on the `marine_models` key in `experiment.yaml` only relevant models would be called.

Dooruk: `geos_marine`: ['mom6', 'cice6', 'ww3']

src/swell/tasks/get_observations.py
```diff
@@ -102,6 +108,10 @@ def execute(self) -> None:
         r2d2_local_path = self.config.r2d2_local_path()
         cycling_varbc = self.config.cycling_varbc(None)
 
+        # Get model component
+        model_component = self.get_model()
+        r2d2_model = r2d2_model_dict.get(model_component, model_component)
```
Dooruk, line 113: Is this only relevant for bias coeff and satellite time lapse type files?

src/swell/tasks/get_observations.py
```diff
@@ -205,49 +221,87 @@ def execute(self) -> None:
...
+            if bias_file_type is not None:
+                if fetch_required:
+                    self.logger.info(f'fetching obs_experiment: {obs_experiment}')
+                    self.logger.info(f'fetching model: {r2d2_model}')
+                    self.logger.info(f'fetching provider: gs')
+                    self.logger.info(f'fetching observation_type: {observation}')
+                    self.logger.info(f'fetching file_extension: {bias_file_ext}')
+                    self.logger.info(f'fetching file_type: {bias_file_type}')
+                    self.logger.info(f'fetching date: {background_time_iso}')
+                    self.logger.info(f'fetching target_file: {target_bccoef}')
```
Dooruk:
```suggestion

```

redundant, r2d2 outputs these

src/swell/tasks/get_observations.py
```diff
@@ -205,49 +221,87 @@ def execute(self) -> None:
...
+                    self.logger.info(f'fetching model: {r2d2_model}')  # geos
+                    self.logger.info(f'fetching provider: gsi')  # gsi
+                    # aircraft_temperature
+                    self.logger.info(f'fetching observation_type: {observation}')
+                    self.logger.info(f'fetching file_extension: {bias_file_ext}')
+                    self.logger.info(f'fetching file_type: {bias_file_type}')  # acftbias
+                    # 2023-10-09T15:00:00Z
+                    self.logger.info(f'fetching date: {background_time_iso}')
+                    # /path/to/bias_coef.nc
+                    self.logger.info(f'fetching target_file: {target_bccoef}')
```
Dooruk:
```suggestion

```
redundant, r2d2 outputs these

src/swell/tasks/save_obs_diags.py
```diff
@@ -9,9 +9,9 @@
 
 import os
 from swell.tasks.base.task_base import taskBase
-from r2d2 import store
 from swell.utilities.r2d2 import create_r2d2_config
 from swell.utilities.run_jedi_executables import check_obs
+import r2d2
```
Dooruk: nitpick but can you keep `import` statements together 😄 ?

src/swell/tasks/save_obs_diags.py
Dooruk: I agree with @mer-a-o here, I think we are repeating what `check_obs` does here.

src/swell/tasks/save_obs_diags.py
```diff
@@ -57,29 +57,79 @@ def execute(self) -> None:
...
-                  obs_type=name,
-                  type='ob',
-                  experiment=self.experiment_id())
+            self.logger.info(f'Found diagnostic output file: {obs_path_file}')
+
+            # Store to R2D2
+            # ---------------
+            self.logger.info(f'Storing feedback file {obs_path_file} to r2d2')
+            self.logger.info(f'  item=feedback, observation_type={name}')
+            self.logger.info(f'  experiment={self.experiment_id()}')
```
Dooruk:
```suggestion
```
redundant, r2d2 outputs these

Dooruk (changes requested):
I was able to run three suites successfully!

Most of my comments are taking out redundant logger outputs as `r2d2` already puts out that information. For me this is really close, with three things holding this back for me, other than ingesting more obs. ofcourse:

1) There is an addition to the taskBase for r2d2 but could that be put in `utiltiies/r2d2.py` and only employed during R2D2 relevant tasks?
2) Model types should support more for marine (I briefly explained this in my comment)
3) In terms of storage, what "lifetime" are we choosing? Are people storing their experiments temporarily? Is everyone storing their experiments in the general gmao database?


Also, can you bump up the SWELL version? I would like to make a release before we merge this.

I also noticed a slight increase in the time it takes to complete `getObservations`. Not a huge deal as we can trigger this task at a cycle before any other dependencies but just got me thinking.

Dooruk:
I probably forgot to mention this one but in cycling tasks, there is a need to save backgrounds and/or restarts. `saveObsDiags.py` is saving observation outputs but we would also like to save model fields, could be forecast, analysis, or restarts. Currently `saveRestart.py` does this in a super manual way but it would be nice R2D2v3 to handle this for us. This is a bit more involved with model runs so I can look into that with the new approach but I need to grasp a few more things in terms of "lifetime".

I am not super clear on where we will end up in terms of `experiment_id` (manual setup vs. randomly assigned) and how long the items are stored within R2D2 but I see there are `lifetime name` values ( debug, science, publication, release). I would say only certain users would have access to `permanent` storage, correct?

@mer-a-o do you store all hourly model state files for the backgrounds? Also do you store ensemble files? (you being in Skylab)

src/swell/tasks/base/task_base.py
ftgoktas: If I ran this inside an R2D2-specific task, I tried 🙃, that module would be imported first, `import r2d2` would run at module load, and credentials wouldn’t be set yet and causes initialization errors.
Putting it in the base class guarantees credentials are loaded before any task module. Will take a look at this again to see if it can be any other way.

src/swell/deployment/platforms/nccs_discover_cascade/r2d2_config.yaml
ftgoktas: I don't think r2d2_config.yaml is relevant for v3, we can remove it. Only the credential loading is needed for v3.

src/swell/tasks/save_obs_diags.py
ftgoktas: Agreed, removing it.

src/swell/tasks/base/task_base.py
Dooruk: I see.. Ok, this is not a show stopper so we can also address it later. Just need to make a note if it or create a minor issue.

Dooruk:
So this fails the `save_restart.py` task, which is basically just a copying operation meant to be used with R2D2v1. However, that lead me to thinking further about a potential issue which I described below:

https://github.com/JCSDA-internal/r2d2/issues/839

For now I can deactivate the `save_restart` and move forward with this PR but I would like to ingest more obs before merging.

@mranst would you mind testing this on `gmao_ci` when you get a chance since this will break tests there unless the API key is setup?

mranst:
> @mranst would you mind testing this on `gmao_ci` when you get a chance since this will break tests there unless the API key is setup?

https://github.com/GEOS-ESM/swell/actions/runs/19082156562

I added the `gmao_user` credential file to `gmao_ci`, r2d2 tasks are failing with this error, looks like it's expecting a string true or false and getting a bool type at some point. It's happening for me locally as well

image

ftgoktas:
The issue was caused by a bug in the r2d2 codebase, the r2d2-client module on Discover has been updated with the fix, and tasks should now run without errors. @mranst could you please pull the latest changes and re-run the tests?

mranst:
> The issue was caused by a bug in the r2d2 codebase, the r2d2-client module on Discover has been updated with the fix, and tasks should now run without errors. @mranst could you please pull the latest changes and re-run the tests?

https://github.com/GEOS-ESM/swell/actions/runs/19150912180/job/54740655821

`SaveRestart` is failing for `3dvar_cycle` and `3dfgat_cycle`, but otherwise looks good

ftgoktas:
This will be addressed in #652

Dooruk:
Working towards merging this, some final things:

- I need to turn off `saveRestart` for this PR. I will create a branch off of yours and make a PR for that.
- `storebackground` is not used currently but we need to revive that (rename it to saveBackground) after this goes in. restart and background are different in DA and modeling context. 

Q.s:
- I see saveobsdiags are in `feedback`. Is `feedback` specific item like a IODA output? i couldn't find much information on this.
- I made a tiny edit to your ingest script and actually testing with that it seems to be working. However there seems to be no protection in terms of who can write into `nccs-gmao` data_hub currently or who can remove files from there, is that correct?

ftgoktas:
> Q.s:
> 
> * I see saveobsdiags are in `feedback`. Is `feedback` specific item like a IODA output? i couldn't find much information on this.

Yes, feedback is for IODA observation diagnostic files produced by DA systems, contains original observations plus analysis results. I had confirmed with the R2D2 team that JEDI 3dvar output files (observations + analysis feedback) should use item='feedback', which applies to SaveObsDiags since that task stores those files.

> * I made a tiny edit to your ingest script and actually testing with that it seems to be working. However there seems to be no protection in terms of who can write into `nccs-gmao` data_hub currently or who can remove files from there, is that correct?

The ingest script doesn't enforce permissions; it relies on the R2D2 v3 API server to enforce authentication/authorization, so unauthorized users will get permission errors from the API if they don't have the API key set up in their environment.

Dooruk:
Ok, this is ready to go in after #653 merges. There are still things to be addressed but no need to hold this one back.

I made a release right before this so we should be good.

Dooruk (approved):
Excellent job! Long time coming.

PR #620, opened 2025-09-05
Handle task aborts using exceptions

mranst:
## Description

This PR removes the use of `sys.exit` from the `Logger.abort` method, instead opting to raise exceptions. I personally like this better from the standpoint of allowing the user to catch specific exception types. It can potentially allow more dynamic handling of exception messaging. It's also more direct, being able to natively handle tracebacks.

This PR also fixes a bug where the pinned version test would switch Swell to develop, rather than jedi

`Logger.abort` now has an parameter to throw any type of exception. I chose to have it expand the use of the niche `SwellError` exception

### DIFF
--- a/src/swell/tasks/stage_jedi.py
+++ b/src/swell/tasks/stage_jedi.py
@@ -12,6 +12,7 @@ import os
 
 from swell.tasks.base.task_base import taskBase
 from swell.utilities.filehandler import get_file_handler
+from swell.utilities.exceptions import SwellError
 from swell.utilities.file_system_operations import check_if_files_exist_in_path
 
 
@@ -82,5 +83,5 @@ class StageJedi(taskBase):
                 self.logger.abort('One or more files not ready')
             else:
                 fh.get()
-        except self.SwellError as e:
+        except SwellError as e:
             self.logger.abort(str(e))
--- a/src/swell/tasks/stage_jedi.py
+++ b/src/swell/tasks/stage_jedi.py
@@ -11,8 +11,7 @@
 import os
 
 from swell.tasks.base.task_base import taskBase
-from swell.utilities.filehandler import *
-from swell.utilities.exceptions import *
+from swell.utilities.filehandler import get_file_handler
 from swell.utilities.file_system_operations import check_if_files_exist_in_path
 
 
@@ -83,5 +82,5 @@ class StageJedi(taskBase):
                 self.logger.abort('One or more files not ready')
             else:
                 fh.get()
-        except SWELLError as e:
+        except self.SwellError as e:
             self.logger.abort(str(e))
--- a/src/swell/utilities/filehandler.py
+++ b/src/swell/utilities/filehandler.py
@@ -271,7 +271,7 @@ class StageFileHandler(FileHandler):
                     filelist = glob.glob(src)
 
                     if not filelist and self.strict:
-                        raise Exception('Source inputs not found "'
+                        raise FileNotFound('Source inputs not found "'
                                         + src + '"')
 
                     for srcfile in filelist:
--- a/src/swell/utilities/filehandler.py
+++ b/src/swell/utilities/filehandler.py
@@ -221,7 +221,7 @@ class FileHandler(object):
         if os.path.islink(dst):
             os.remove(dst)
         if os.path.isfile(dst):
-            raise Exception('File exists where link expected: "' + dst + '"')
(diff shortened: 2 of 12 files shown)


src/swell/tasks/stage_jedi.py
```diff
@@ -83,5 +82,5 @@ def execute(self) -> None:
                 self.logger.abort('One or more files not ready')
             else:
                 fh.get()
-        except SWELLError as e:
+        except self.SwellError as e:
```
ashiklom: Why is this an attribute of `self` rather than just a global-scope `SwellError`?

ashiklom: This looks like the only place where we use exceptions as attributes. Elsewhere (including in your code below), we use imported global exceptions.

Unless we have a very good reason, I would opt for using the global imported exceptions. It's arguably less powerful and flexible, but I don't think we need a ton of power/flexibility here, and the global imported exception approach is simpler and more idiomatic.

src/swell/utilities/logger.py
```diff
@@ -111,20 +110,13 @@ def debug(self, msg: str, wrap: bool = True, *args, **kwargs) -> None:
...
     # ----------------------------------------------------------------------------------------------
 
-    def abort(self, msg: str, wrap: bool = True, *args, **kwargs) -> None:
-        msg = red + msg + end
-        msg = self.format_message(msg, wrap, 'ABORT')
-        super().critical(msg)
+    def abort(self, msg: str, wrap: bool = True,
+              exception: Exception = SwellError, *args, **kwargs) -> None:
+        formatted_msg = red + msg + end
+        formatted_msg = self.format_message(formatted_msg, wrap, 'ABORT')
```
ashiklom: While we're tidying things up anyway, I might suggest the following:

1. **Delete the `red` color**. This only renders correctly in specific terminal interfaces (and, notably, do _not_ render correctly in log files), and when it _doesn't_ render correctly, makes it harder and more annoying to find this. Far better to include a grep-able string like `ERROR` or `WARNING` (which `logger` should provide anyway). So I would leave this blank.

2. **Delete the `format_message` method**. I find the the `format_message` method counterproductive. Terminals automatically wrap strings anyway, as do pagers like `less`. Plus, automatically hard-wrapping the lines makes it much harder to copy-paste these messages.

ashiklom: If we drop the `formatted_message` thing then I think we can delete all of the custom logging methods above, leaving only `abort` (which should be soft-deprecated and eventually removed in favor of plain Python error handling).

ashiklom:
Excellent! This will be a nice QOL improvement. See a few minor suggestions/questions in my comments --- otherwise, I think this can be approved and merged.

mranst:
@ashiklom Thanks - I've removed pretty much everything from the logger class, minus the abort. I was going to delete `exceptions.py` entirely, but I kept it in the case of a few of the code tests, which are looking for an error type to be raised. I figure that a custom error type is better suited in this instance

src/swell/utilities/filehandler.py
```diff
@@ -224,17 +216,14 @@ def link(self, src: str, dst: str) -> None:
...
-            if not os.path.isfile(dst):
-                os.symlink(src, dst)
-        except Exception as e:
-            raise SWELLFileError(str(e))
+            raise FileNotFoundError('Source file does not exist: "' + src + '"')
+
+        if os.path.islink(dst):
+            os.remove(dst)
+        if os.path.isfile(dst):
+            raise Exception('File exists where link expected: "' + dst + '"')
```
ashiklom:
```suggestion
            raise ValueError('File exists where link expected: "' + dst + '"')
```

src/swell/utilities/filehandler.py
```diff
@@ -282,8 +271,8 @@ def list(self, force: bool = False) -> list:
                     filelist = glob.glob(src)
 
                     if not filelist and self.strict:
-                        raise SWELLConfigError('Source inputs not found "'
-                                               + src + '"')
+                        raise Exception('Source inputs not found "'
```
ashiklom:
```suggestion
                        raise FileNotFound('Source inputs not found "'
```

ashiklom (approved):
Nice work -- approved!

Dooruk:
I'm getting the following error by running `swell create 3dvar`. Maybe related to `.swell/swell-slurm.yaml`. How come tier1 tests missed this?

Screenshot 2025-09-17 at 4 02 30 PM

PR #622, opened 2025-09-10
set `max allowable geometry difference` to high value to avoid fv3jedi geom code crash

metdyn:
## Description

After FV3 PR:  https://github.com/JCSDA-internal/fv3-jedi/pull/1356, in geos when the resolution of the bkg file (CX1)  is different from that in the swell atmospheric setting (CX2):
models:
  geos_atmosphere:
    horizontal_resolution:  CX2

the fv3jedi geom subroutine will abort the code, because the array fields read into State is a mismatch, which is the right thing to do.   But setting `max allowable geometry difference` to infinity will allow the code to run, despite for example, C360 field (nf, ny=360, nx=360)  was read in C90 field (nf,ny=90,nx=90), at least to my understanding. 

Hence, this PR is a technical bypass, which I think is problematic. 

## Dependencies

## Impact

output:
OOPS_STATS ---------------------------------- Parallel Timing Statistics (   6 MPI tasks) -----------------------------------

OOPS_STATS Run end                                  - Runtime:     68.24 sec,  Memory: total:     9.17 Gb, per task: min =     1.49 Gb, max =     1.56 Gb
Run: Finishing oops::Variational<FV3JEDI, UFO and IODA observations> with status = 0
OOPS Ending   2025-09-10 11:58:58 (UTC-0600)

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background.yaml
@@ -4,6 +4,7 @@ provider: geos
 datapath: '{{cycle_dir}}'
 filenames: ['bkg.%yyyy%mm%ddT%hh%MM%ssZ.nc4',
             'fv3-jedi/bkg/geos.crtmsrf.{{horizontal_resolution}}.nc4']
+max allowable geometry difference:    1.e15
 state variables: &statevars
 - eastward_wind
 - northward_wind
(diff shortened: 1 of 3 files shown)


mjkagnes123 (approved):
Looks good. Tested with the changed background.yaml and it worked. Thank you!

metdyn:
Hi @all, 

After discussion with @rtodling, we share a common standpoint that with the current fv3-jedi code, the max allowable geometry difference should remain 1.e-3.    

Because `JEDI geom` should equal `BKG file resolution` in scientific experiments. 

This PR is a bypass to allow the jedi code to run, but this will generate incorrect scientific results in comparison with original input BKG file.  Hence this PR needs to be closed to avoid confusions.

Issue #625, opened 2025-09-16
Running GEOS within SWELL in an efficient way


Dooruk:
This is meant to be an ongoing discussion rather than having a dedicated meeting since we are all busy. I would appreciate everyone's input here. Feel free to tag additional folks.

## Description

With new people joining and having more types of model runs required (including ensemble applications) I would like to start this discussion. I'm aware there is a `gcm_setup` and `gcm_run` modernization effort underway but in the meantime we should have a solution with the existing scripts.

I currently have a way of running GEOSgcm within SWELL and it involves executing a mpirun command via `subprocess`. This has worked alright so far. However, with the recent model version, v12, and on Milan Nodes I am not able to run the model in coupled mode as it gets stuck in the initialization stage. I tried lots of different options (environment variables, library setup, esma_mpirun, stderr output redirecting) to no avail. During this time I also noticed significant time difference between **initialization** stage via submitting `gcm_run.j` directly or via executing in `subprocess`.

## (Real) Quick primer on `gcm_run.j` and how it is used in SWELL
`gcm_run.j`: https://github.com/GEOS-ESM/GEOSgcm_App/blob/develop/gcm_run.j#L155

As you can see, this script is templated and monolithic. The entries like `@COUPLED` and `@CICE6` indicate those sections will be filled only if COUPLED mode is active and/or CICE6 model is active. It sets up SLURM, handles preprocessing, setting up BCs and ICs via the `RC` folder, runs the model, postprocessing, executes scripts (e.g., `linkbcs`), sets up environment variables, etc. you name it.

Since Cylc is capable of handling SLURM commands, in the current SWELL implementation, I divided `gcm_run.j` into two tasks (scripts) called `PrepGeosRundir` (preprocessing) and `RunGeosExecutable` (actually running the model). I was able to parse resource (or `.rc`) files and manipulate them according to my DA needs.

Instead of the `scratch` folder in SWELL, we have timestamped `forecast` folders, like so:

<img width="186" height="124" alt="Image" src="https://github.com/user-attachments/assets/25b590cc-2cbb-4a57-8e04-e5bbeb0186ff" />

## Implementation ideas (work in progress) 💡 

I will start working on a new prototype for executing GEOS within SWELL. Inspired by Skylab (JCSDA's workflow engine) and some discussions within our group, and after some pondering, I think the best approach would be to have **model specific scripts** for however developers want to handle the `gcm_run.j`.

[...]

PR #626, opened 2025-09-16
Allow for model-differentiated tasks

mranst:
## Description

Simple pr that would allow for tasks to be differentiated by model. Inspired by #625 and recent discussions about decoupling some of the structure that might be burdened by trying to support both models at the same time. 

When running`swell task <task_name> -m <model_component> config`, Swell will try to import `<task_name>` from `src/swell/tasks/<model>`. If it can't find it, it will just try to find one under `src/swell/tasks/`. This means that any tasks that would benefit from having a separate model-specific version can do so, and any tasks that don't can continue to use the general version.

I haven't added any of these tasks yet, but they just need to be put under `src/swell/tasks/<model>` (for example)


src/swell/tasks/base/task_base.py
```diff
@@ -281,8 +281,20 @@ def create_task(
...
-        task_class = getattr(importlib.import_module('swell.tasks.'+task_lower), task)
+        task_class = None
+
+        # Try to use the model-specific task if it exists
+        if model is not None:
+            try:
+                task_class = getattr(importlib.import_module(f'swell.tasks.{model}.{task_lower}'),
+                                     task)
+            except ModuleNotFoundError:
+                pass
```
ashiklom: Maybe a least throw a `logger.warn` or `logger.info` here?

Dooruk:
Hmm, I started writing my tasks with different names (e.g., RunCoupledGeosExecutable) but I'm not sure about how to group them yet, I'm open to discussion. 

Ultimately, some tasks might be more application specific rather than model specific (GEOS-FP, MERRA3, or GEOS-IT would all use `geos_atmosphere`). I do like the concept of reverting to default if a task not available though.

Dooruk:
We briefly touched on this today but this PR seems to be a good first step in the right direction. I began doing this in an inelegant way in my convoluted PR #643

I guess I see two potential issues here:

1) 3-4 tasks/scripts doing different things will have the same name with a generic task at the high `/tasks` level. This is not a problem for seasoned SWELL user but might be confusing for new users. This is already how it is handled in `flow.cylc` after the jinja template is filled with proper values (e.g., `EvaJediLog-geos_marine` or `EvaJediLog-geos_atmosphere`). 

2) The tasks are not necessarily only model-specific but could be (or should be) application specific (coupled reanalysis, coupled seasonal forecast, OSSE etc. I'm just throwing some of them out there). This is where things get out of control but allowing certain tasks/applications to manipulate the `gcm_run.j` script (or it's next-gen equivalent) would allow people to transfer their workflows to ours easier.

I'm open to further discussion @rtodling or Skylab insight @jeromebarre @mer-a-o

mranst:
Since this was brought up again in #677, I've updated this and added messages to the task factory method. Hopefully this will help make it clear which task module is being used, but as noted before this does have the potential to be confusing to newer users

Dooruk:
Revisiting my [comment above](https://github.com/GEOS-ESM/swell/pull/626#issuecomment-3628849614):

1) I'm wondering if we want task names in `/model` folders would this be a simple change via:

```py
task_class = getattr(importlib.import_module(f'swell.tasks.{model}.{task_lower}-{model}'), task)
factory_logger.info(f'Using model-specific version of {task} in <swell.tasks.{model}.{task_lower}-{model}>')
```

So tasks can have different names with `-model`. Not sure if this would run into any Swell specific problems.

2) I thought about this further and I'm not sure if it is practical to have suite specific tasks (and folders) and we don't need to implement now. However GMAO tasks/scripts are more specific to the product type, rather than model, MERRA3, NWP, OSSE, S2S etc.

   However, to facilitate it we could create a hierarchy:
   - Check suite specific tasks
   - Check model specific tasks
   - Resort back to generic task
   
   but this is more code in terms of Swell wrapper code, task factory etc. So more confusion!

mranst:
>        However, to facilitate it we could create a hierarchy:
>        
>        * Check suite specific tasks
>        * Check model specific tasks
>        * Resort back to generic task
>        
>        but this is more code in terms of Swell wrapper code, task factory etc. So more confusion!

Yeah it's a lot to keep track of. If application-specific tasks would provide more utility, it may just be better to ditch the model-specific idea entirely and just have differently named tasks and bake more logic into flow.cylc that selects the appropriate task. If `gcm_run.j` needs different parameters depending on the suite/application etc, we can just add keys to `experiment.yaml` and do something like the following:

```
{% if application == 'example1' %}
Precursor => Example1
{% if application == 'example2' %}
Precursor => Example2
{% endif %}
```

Dooruk:
> Yeah it's a lot to keep track of. If application-specific tasks would provide more utility, it may just be better to ditch the model-specific idea entirely and just have differently named tasks and bake more logic into flow.cylc that selects the appropriate task. If `gcm_run.j` needs different parameters depending on the suite/application etc, we can just add keys to `experiment.yaml` and do something like the following:
> 
> ```
> {% if application == 'example1' %}
> Precursor => Example1
> {% if application == 'example2' %}
> Precursor => Example2
> {% endif %}
> ```

Let's stick with your changes for now and keep things simple. It is a good step towards model specific tasks and we can just expect different applications to name particular tasks under the model folder.

Dooruk (dismissed):

mranst:
I changed the file naming as suggested above so they follow `swell.tasks.{model}.{task_name}-{model}`, otherwise this should be ready

Dooruk (approved):
Thanks, looks good!

Dooruk:
Does this approach support having a single model only task under the model folder? I think the valid tasks currently only count if they are placed under the `/tasks` folder.

There are certain tasks that are only relevant for a particular model, for instance:

https://github.com/GEOS-ESM/swell/blob/953b02879e9206fdd70c88fb16f20faaf6dbbeb9/src/swell/tasks/generate_observing_system_records.py#L27-L31

If this task is called with `-m geos_marine` in `flow.cylc` it will run but do nothing. Which is fine, but I'm wondering if this is the most efficient approach or should we address this minor redundancy?

I was thinking having certain tasks only under the model folders might work but than for coupled workflows this will fail, as there won't be a generic task for `-m geos_model`.

Maybe with #656 more conditionals can take place to handle it?

Sorry, I'm testing implementing this in my PR and just thought about this after it already merged.

PR #582, opened 2025-06-20
Allow for use of 3d-pressures in interface to observers

rtodling:
## Description

These changes allow for use of 3d-pressure in the observation operators - this is particularly relevant for handling GNSSRO. 

This has been tested JEDI 18Jun2025.

The aim is to reset the pinned version to that of 18Jun2025.

## Dependencies

JEDI 18 Jun 2025

- [x] https://github.com/GEOS-ESM/jedi_bundle/pull/77 

In reality there is not true dependence on the PR above - but we should avoid multiple changes ... so it would be best to first approve the JEDI_BUNDLE PR above and build a 18 Jun version of JEDI that is consistent w/ the changes in the PR in the bundle.

## Impact

proper handling of UFO operators that depend on 3d (edge) pressures.

PR #629, opened 2025-09-17
Combine multiple PRs

Dooruk:
Ingests https://github.com/GEOS-ESM/swell/pull/622 and https://github.com/GEOS-ESM/swell/pull/536 and https://github.com/GEOS-ESM/swell/pull/582


Dooruk:
@metdyn and @rtodling this is good to go in, 

As expected it creates difference in variational results. Hopefully with Michael's changes we will have a better handle over this.

PR #630, opened 2025-09-17
remove wrap arguments from logger

mranst:
Sorry about this! This should be the last of them


Dooruk (approved):
All good, thanks for taking care of it!

PR #637, opened 2025-10-01
Update to sync with sfcCorrected UFO changes

rtodling:
## Description

This syncs develop with UFO's change:

https://github.com/JCSDA-internal/ufo/pull/3779

I find this change not to be zero diff; but the differences are roundoff.

Untitled

Yes - there are two curves of each in rhe plot above, but you can't see the diff. So I suppose that's acceptable as far as differences go.

## Dependencies

None

## Impact

It should be noted that the changes in UFO associated w/ this are troubled for some configuration of the yamls - ok for SWELL's configuration for now.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/sfc.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/sfc.yaml
@@ -39,10 +39,10 @@ obs operator:
      - name: virtualTemperature
 
    # Surface pressure is handled using the correction scheme
-   - name: SfcPCorrected
+   - name: SfcCorrected
      variables:
      - name: stationPressure
-     da_psfc_scheme: GSI
+     correction scheme to use: GSL
      geovar_geomz: geopotential_height
      geovar_sfc_geomz: height_above_mean_sea_level_at_surface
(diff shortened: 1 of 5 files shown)


Dooruk:
This is going to need a build after September 16th. We could test the new compare var suite and show zero-diff with a more recent build:

https://github.com/orgs/JCSDA-internal/discussions/186

metdyn:
I donot see a difference using swell task compare_variational (thanks @mranst @Dooruk) before and after the UFO keyword change (Sep-14 vs Oct-14). So the PR is ready to merge once new JEDI build becomes up to date.
Screenshot 2025-10-15 at 1 12 32 PM

metdyn (dismissed):
Works well, in my opinion.

Dooruk:
Thanks so much @metdyn for checking this!

@mranst, without this fix newer (past September 16th) JEDI `fv3-jedi` builds will fail while running `3dvar_atmos`, `3dfgat_atmos` and `hofx` tasks. 

So we should update the pinned JEDI build to `fv3_soca_SLES15_10022025`, I will make the changes in this branch and trigger a Tier 1 test.

Dooruk (approved):

PR #639, opened 2025-10-02
compare_variational suite without flow.cylc changes

mranst:
## Description

This PR adds the variational comparison suite that is part of #540, without the flow.cylc overhaul. We can potentially merge that one later, but I'm feeling like I may want to change some things about that PR and wanted to get this in sooner

It works basically the same way as before, docs are under docs/examples/soca/comparison_workflows.md

### DIFF
--- a/docs/examples/soca/comparison_workflows.md
+++ b/docs/examples/soca/comparison_workflows.md
@@ -1,0 +1,14 @@
+# Running comparison experiments within Swell
+
+Swell currently supports comparing variational experiments.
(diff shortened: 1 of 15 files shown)


Dooruk (approved):
All good, I don't want to hold this back any longer. The logic is straightforward and we can make relevant changes easily. 

I saw you triggered the tests, once they pass it is good to merge 👍🏼

mranst:
Thanks!

PR #643, opened 2025-10-17
A different way of executing GEOS from SWELL

Dooruk:
To address #625 I worked on a different way of executing GEOSgcm in SWELL. I had to make a lot of changes but i will try to explain the essential part of this PR.

Let's consider `gcm_run.j` in 4 stages:

1) SLURM & node assignment
2) Preprocessing
3) Execution
4) Postprocessing

In the current implementation, SWELL handles 2 & 3 via python and `subprocess` and 1 is assumed to be set properly by the user, which caused trouble with the NCCS. For DA purposes 4, postprocessing is explicitly handled by SWELL but that is not the focus of this PR.

In this proposed implementation, the main difference is that we rely on `gcm_run.j` for 2 and 3 by conducting surgical edits via `PrepCoupledGeosRundir` at few locations and running `gcm_run.j` directly from Cylc (which doesn't capture failed exit status):

```cylc
    [[RunGeos]]
        script = "{{experiment_path}}/forecast/gcm_run.j"
        platform = {{platform}}
        [[[job]]]
            shell = /bin/csh
        [[[directives]]]
        {%- for key, value in scheduling["RunGeos"]["directives"]["all"].items() %}
            --{{key}} = {{value}}
        {%- endfor %}
```

I created the `3dfgat_coupled_cycle` suite for testing, should work by default if anyone has time to check it out.

Sorry I'm tagging a lot of people but I would like to get some feedback on this, but not sure if my explanation was sufficient here. Otherwise i can explain it during the next Swell Development Tag-up.

Also it has the idea of model-differentiated tasks (inspired by the Skylab workflow and @mranst has a PR for it already), so perhaps having `tasks/model/` might be better way of organizing the model specific tasks like he proposed.

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_marine/model/background_error.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_marine/model/background_error.yaml
@@ -12,9 +12,12 @@ saber central block:
     - sea_surface_height_above_geoid
   read:
     groups:
-    - variables: [sea_water_potential_temperature, sea_water_salinity]
+    - variables:
+      - sea_water_potential_temperature
+      - sea_water_salinity
       horizontal:
-        filepath: 'background_error_model/hz_rossby'
+        filepath: 'background_error_model/new_corr_1p0'
+        # filepath: 'background_error_model/hz_rossby'
       vertical:
         levels: {{vertical_resolution}}
         filepath: 'background_error_model/vt.{{local_background_time}}'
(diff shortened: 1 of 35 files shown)


DavidSRussell:
Looks like the default `3dfgat_coupled_cycle` ran fine for me. The only obvious differences I see are the location of the forecast directory and the time stamp on `cycle_done`. Not sure how well I understand the pros and cons of the new approach; maybe we can go over it some time?

DavidSRussell:
Just a reminder that we also talked about including time series of the mean observation error, which would be useful for tracking changes in the altimetry processing.

jeromebarre:
Hi, it seems that some commits of the PR address changes that are not part the actual feature proposed.
For example:
https://github.com/GEOS-ESM/swell/commit/7caae0c42cb6cc716cef6044b46ba60defe878ab

would it be possible to do a cleanup (or recreate a new branch with just the correct cherry picked commits) so it's easier to review.

Thanks a lot!

Dooruk:
see #677

PR #644, opened 2025-10-27
Fix for compare suite and rename

mranst:
This addresses an error in JediOopsLogParser and renames the compare suite


Dooruk (approved):
👍🏼

Issue #645, opened 2025-10-27
handling empty observation


mer-a-o:
The [check_obs](https://github.com/GEOS-ESM/swell/blob/develop/src/swell/utilities/run_jedi_executables.py#L23) function in SWELL checks whether an observation file is empty. If it is, the observation is then [removed from the observation list](https://github.com/GEOS-ESM/swell/blob/develop/src/swell/utilities/run_jedi_executables.py#L109-L113)
 to prevent the file from being used downstream.

This behavior causes issues in cases where only one type of observation is being assimilated and some assimilation cycles contain no observations. For example, TEMPO observations are only available during the daytime, so each day includes at least one assimilation window without any observations. In these cases, SWELL builds an incomplete JEDI executable YAML, leading to a runtime failure:

```
...
observations:
  observers: []
...
```

A more robust approach could be to create empty observation files during the ingestion (storage) step in R2D2. These files would then be fetched during assimilation, and the JEDI executable YAML could be configured in the same way as for non-empty files.

JEDI can handle empty observation files by generating empty feedback files and storing them in R2D2. This would simplify the SWELL logic and eliminate the need for special checks and conditional handling.

This approach might also be preferable since the ignore_missing capability [was removed in R2D2 v3](https://github.com/JCSDA-internal/r2d2/pull/309). However, users would need to be aware that empty observation files will not trigger any distinct warning messages, so additional validation or logging may be helpful.

I'm opening this issue so we can discuss and identify the best way moving forward. Please let me know what you think.

Dooruk:
From a general perspective, having empty files is fine but it will be waste of lot of ionodes, especially now the disk space is limited. Also we still don't know R2D2 performance for a lot of local users with thousands of files (scalability).



From marine only perspective, one complication is having to run DA in different window lengths (6-hr, 1-day or 5-days)

I have the following method to combine observations, assuming they are all ingested in 6-hourly chunks:

https://github.com/GEOS-ESM/swell/blob/29ebff8d2b33bc1d8b95751ede4e590a7eb9e8b5/src/swell/tasks/get_observations.py#L420

Some PT6H observations might be missing from the full period, say while combining five 6-hourly observations for a 1-day DA window. That handling is not perfect.

jeromebarre:
I agree with Doruk with the potentail limitation of the ionodes... 
One tradeoff I am seeing is to have one template of empty file stored somewhere protected. 
[...]

PR #651, opened 2025-11-05
Add manual setting for obs providers

### LINKED ISSUE #648: obs name X obs provider in `get_observations.py`
In [get_observations.py](https://github.com/GEOS-ESM/swell/blob/develop/src/swell/tasks/get_observations.py#L150-L178), the code loops over both observations and obs_providers, trying to fetch data for every combination of obs_provider-observation defined in `task_questions.yaml`. This results in the code trying to retrieve observations that don’t exist for certain provider–observation combinations.

For example, suppose `task_questions.yaml` defines the following:

```
obs_provider:
  default_value:
  - nasa
  - esa
  options:
  - nasa
  - esa

observations:
  default_value:
  - tempo_no2_tropo
  - tropomi_s5p_no2_tropo
  options:
  - tempo_no2_tropo
  - tropomi_s5p_no2_tropo
```

In this setup:
`tempo_no2_tropo` should be fetched from `nasa` and `tropomi_s5p_no2_tropo` should be fetched from `esa`. The current logic tries all combinations (e.g., `tempo_no2_tropo` with both `nasa` and `esa`, and the same for `tropomi_s5p_no2_tropo`), leading to failures when non-existent combinations are attempted. 

On approach could be to only specify the provider in observation yaml file (under `interfaces/../observations/obs.yaml`) using a special yaml key such as `_provider` and read that in the `get_observations` task.

mranst:
Doruk edit:

#648 This PR adds the ability to manually set providers for observations in `observation_ioda_names.yaml`:

```yaml
  - ioda name: adt_jason3
    full name: ADT Jason-3
    provider: odas
    inst type: altimeter
```

There is still ambiguity in terms of what does `provider` indicate and what name to choose but for now this is an R2D2 requirement.


mer-a-o:
Thanks @mranst for adding this feature. I tested it and it worked for me. 
I needed to add both providers (`nasa` and `esa`) to my `task_questions.yaml` and in `observation_ioda_names.yaml`. 

One suggestion I have is to use single `provider` instead of plural `providers` in `observation_ioda_names.yaml`. 

In all the examples I have seen the provider name is embedded in the "ioda name" so you don't have multiple provider for one observation type. With the same logic I think having providers list in`task_questions.yaml` can be redundent and confusing for the user (e.g. user may be confused by the order of items in `obs_provider` and `observations`). Are there cases where for a "ioda name" or "observations" user defines multiple providers?

Dooruk:
Oops, i was using old modules, now it is working..

Sorry all, back to development mode.

Dooruk:
Well, another question. Is fetch supposed to fail for non-existing obs? I'm only getting a warning @ftgoktas

ftgoktas:
> Well, another question. Is fetch supposed to fail for non-existing obs? I'm only getting a warning @ftgoktas

Yes, fetch actually raises an exception for non-existing files, but `GetObservations` task catches it and logs as `.info()` instead of `.error()`. If we want stricter behavior, we can replace `logger.info()` with `logger.error()` or remove the try-except (in `get_observations.py`).

Dooruk:
> > Well, another question. Is fetch supposed to fail for non-existing obs? I'm only getting a warning @ftgoktas
> 
> Yes, fetch actually raises an exception for non-existing files, but `GetObservations` task catches it and logs as `.info()` instead of `.error()`. If we want stricter behavior, we can replace `logger.info()` with `logger.error()` or remove the try-except (in `get_observations.py`).

Ah, sounds like you found a way to circumvent #645, but we need to address it intentionally.

jeromebarre:
> Thanks @mranst for adding this feature. I tested it and it worked for me. I needed to add both providers (`nasa` and `esa`) to my `task_questions.yaml` and in `observation_ioda_names.yaml`.
> 
> One suggestion I have is to use single `provider` instead of plural `providers` in `observation_ioda_names.yaml`.
> 
> In all the examples I have seen the provider name is embedded in the "ioda name" so you don't have multiple provider for one observation type. With the same logic I think having providers list in`task_questions.yaml` can be redundent and confusing for the user (e.g. user may be confused by the order of items in `obs_provider` and `observations`). Are there cases where for a "ioda name" or "observations" user defines multiple providers?

I think there could be and there are already cases where we use multiple providers per observation. For the composition observation we do not include the provider in the observaiton name. The nomencalture I chose a few years back was `<instrument>_<variable>_<type> `e.g. `tempo_no2_tropo`. But we could agree to put the provider in the name too. But what is actually provider? We would need to agree on this. Is it the institution that provided the observation product, e.g. `ESA, NASA, etc.`.. or is it the data **stream** it comes from or is used for  e.g.  `odas, gdas_marine, etc`...

In our composition work we also use this key as a way to create test datasets without overwriting the current reference dataset we use... e.g. `tempo_demo, nasa_VCD`, etc...

As r2d2 is currently is, the `provider` key is the only way to store another dataset for the same observation and same date without chaging its name to avoid overwritting...

Dooruk:
This is good to go, could use a second set of eyes. @mranst I can't add you as a "self-reviewer".

@jeromebarre I followed EMC's nomenclature when i first begun using their OBS:

`<variable>_<instrument>`

When I first started a few years back, `provider` wasn't clear to me so hence I named `odas` and `gdas_marine` but going forward having these in the filename might be a safer bet.  For now `provider` key in `ioda_obsnames` should clarify that confusion.

mer-a-o:
I tested this branch with my hofx-geos-cf branch and it works well. Thank you @mranst and @Dooruk . 
I only have one question/comment, If I don't add `provider` for my obs in `observation_ioda_names.yaml` the `GetObservations` task will still pass (green box in cycl tui). But in `job.err` I see the error message that r2d2 fetch failed as `provider` is empty. Later, RunJedi fails because the observation files is not available in the run directory. Is this intentional or should we set the task `GetObservations` status to "failed" in this case?

mranst:
> I tested this branch with my hofx-geos-cf branch and it works well. Thank you @mranst and @Dooruk . I only have one question/comment, If I don't add `provider` for my obs in `observation_ioda_names.yaml` the `GetObservations` task will still pass (green box in cycl tui). But in `job.err` I see the error message that r2d2 fetch failed as `provider` is empty. Later, RunJedi fails because the observation files is not available in the run directory. Is this intentional or should we set the task `GetObservations` status to "failed" in this case?

Right now any exception in `fetch` is being passed through
https://github.com/GEOS-ESM/swell/blob/2717eb0ca73f31b29ad7a24ab7d3fe5b26beb3d6/src/swell/tasks/get_observations.py#L178-L182

Dooruk:
Good catch, I set that to fail when no provider is given.

I'm setting up an R2D2 meeting next week to discuss how to handle empty obs among many other things. Exception handling is ok while we are implementing but can't be a long term solution.

Dooruk (approved):
This is good to go in, I tested with actions.

mer-a-o (approved):

PR #654, opened 2025-11-10
Minor QOL updates

mranst:
## Description

This PR addresses two things that have been bugging me about swell for a while. The first allows for a location to be set for code tests to send the `jedi_bundle`, `GEOS_mksi`, and `output` directories that are created when code testing, rather than being created in the cwd. The second suppresses some of the output from the subprocess commands used to run cylc from within swell, making the error messages a bit easier to parse


Dooruk (approved):
👍🏼

PR #655, opened 2025-11-10
Fixes for comparison tests

mranst:
## Description

This should be the final set of fixes for the comparison tests to work with tier2 tests. The `3dfgat_atmos` comparison still fails because there's apparently something different between jedi built within swell and jedi_bundle. I've tried everything I could think of at the moment to reconcile this difference, including making sure the pinned versions match, using an updated version of jedi_bundle, and playing around with different npx_proc and npy_proc to match the specs in the geos_atmosphere config (not sure if that would make a difference).


src/swell/tasks/prepare_analysis.py
```diff
@@ -126,7 +126,7 @@ def mom6_increment(self, f_rst: str, ana_path: str, incr_path: str) -> None:
         # -----------------------------------------------------------------------
         command = f'ncks -O -v h {f_ana} {h_ana} \n' + \
             f'ncks -A -v h {h_ana} {f_incr} \n' + \
-            f'mv {f_incr} {mom6_incr}'
+            f'cp {f_incr} {mom6_incr}'
```
Dooruk, line 129: Sorry, I didn't think ahead..

Dooruk (approved):

Dooruk:
Guessing this doesn't break anything?

mranst:
> Guessing this doesn't break anything?

Tier1 tests look good 👍

PR #656, opened 2025-11-10
Updated workflow templating

mranst:
This is the most recent version of the concept of updating the workflow templating to use python. This approach should be more straightforward and easier to understand than the previous version, as it uses the same jinja templating in the graph section to make it easier to read (though it doesn't necessarily have to). This pr also merges the task_runtimes and task_questions objects used by the previous pr into one object for simplicity. I'm aiming for this approach to allow for greater independence of the tasks from the suite creation system, to allow generating configs for single tasks and having mock values for code tests for each task, but that's not ready to show off yet. 

I still need to merge develop into this and update the docs, but I wanted to open this up as a draft


mranst:
My most recent commits added a new feature to this pr - the ability to create configs for individual tasks. For example, you can now run `swell create-task-config CloneJedi` or `swell create-task-config GetObservations -m geos_marine -d 20210701T120000Z` (for any tasks), and use the cli or modify the config file to run tasks without generating an entire suite. This will then give you an executable shell file, or in the case of slurm tasks, an `sbatch` file. This obviously won't be useful for most tasks at the moment, given how interconnected they are, but some may find it useful for debugging or for the more independent tasks (such as the `JediLogComparison` task. I plan on expanding this feature to add code tests for individual tasks using mock values.

docs/examples/templating_workflows.md
```diff
@@ -0,0 +1,79 @@
...
+    execution time limit = PT30M
+    [[[directives]]]
+        --job-name = EvaObservations-geos_marine
+        --qos = allnccs
+        --nodes = 1
+        --ntasks-per-node = 64
+        --constraint = mil
+        --no-requeue =
+        --account = <account>
+```
```
ashiklom, line 84: I'm not sure what this code block refers to or how it connects to the earlier code blocks. Is this the rendered output from one of the classes above? If so, suggest being explicit about that; e.g.,

> The output of code that looks approximately like this:
> 
> ```python
> evaobs = EvaObservations()
> evaobs.slurm = {"nodes": 1, "ntasks-per-node": 64, ...}
> evaobs.get_workflow_string()
> ```
> 
> ...is a string that looks like this:
> 
> ```
> [[EvaObservations-geos_marine]]
> # ...
> ```

docs/examples/templating_workflows.md
```diff
@@ -0,0 +1,79 @@
...
+This can be used to set task-specific defaults in `task_attributes.py`, rather than being set in `slurm.py`:
+
+```python
+class RunJediConvertStateSoca2ciceExecutable(Task):
+    def set_attributes(self):
+        self.is_cycling = True
+        self.is_model = True 
+        self.time_limit = True
+        self.slurm = {'nodes': 1}
+```
```
ashiklom: For clarity/completeness, suggest stating exactly which task-specific defaults are being overwritten.


```suggestion
This can be used to set task-specific defaults in `task_attributes.py`, rather than being set in `slurm.py`. For example, the task below defaults to slurm setting `--nodes=1`:

```python
class RunJediConvertStateSoca2ciceExecutable(Task):
    def set_attributes(self):
        self.is_cycling = True
        self.is_model = True 
        self.time_limit = True
        self.slurm = {'nodes': 1}
```
```

File: src/swell/configuration/jedi/interfaces/geos_atmosphere/suite_questions.yaml [hunk 3]
```diff
@@ -2,6 +2,12 @@ cycle_times:
   default_value: ['T00', 'T06', 'T12', 'T18']
   options: ['T00', 'T06', 'T12', 'T18']
 
+cycling_varbc:
+  default_value: false
+  options:
+  - true
+  - false
+
```
ashiklom, line 8: Flagging that this looks like an unrelated change. When we do the final squash-merge, we need to make sure there's a note in the message somewhere that this was also done.

src/swell/configuration/jedi/interfaces/geos_marine/suite_questions.yaml
```diff
@@ -10,3 +10,13 @@ ensemble_hofx_packets:
...
+
+marine_models:
+  default_value:
+  - mom6
+  - cice6
+  options:
+  - mom6
+  - cice6
+  - bgc
+  - ww3
```
ashiklom, line 32: Ditto to my comment above re: unrelated change.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -542,10 +411,10 @@ def ask_a_question(
         if model is not None:
             if 'models' not in self.experiment_dict:
                 self.experiment_dict['models'] = {}
-                self.questions_dict['models'] = f"Configurations for the model components."
```
ashiklom, line 545: Was this just a misnomer before? I.e., It was called `questions_dict`, but effectively used as a registry of comments for the output? Or did we make some changes to *how* the stuff in here is being used?

src/swell/deployment/create_task_config.py
```diff
@@ -0,0 +1,221 @@
...
+        modules_file = os.path.join(task_path, 'modules-csh')
+        shell_type = 'csh'
+    else:
+        # Default to bash
+        template_modules_file(logger, experiment_dict, task_path)
+        create_modules_csh(logger, task_path)
+        logger.info('Shell type not detected, make sure you have the proper modules'
+                    ' loaded before running experiment')
+        modules_file = os.path.join(task_path, 'modules')
+        shell_type = 'bash'
```
ashiklom: This feels fragile. `$SHELL` is an environment variable set by default by the running shell, and I think it influences a of programs. I think the system would have to have some kind of nontrivial configuration error for `$SHELL` to be unset.

My suggestion would be to not assume any defaults here and just throw an informative error like "Failed to deduce the target shell. `$SHELL` is currently set to <value of `shell`>".

src/swell/tasks/base/task_base.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         datetime_input: Optional[str],
         model: str,
         ensemblePacket: Optional[str],
-        task_name: str
+        task_name: str,
```
ashiklom: Unnecessary trailing comma?

src/swell/tasks/base/task_base.py
```diff
@@ -191,9 +191,14 @@ def cycle_dir(self) -> str:
         self.logger.assert_abort(self.__model__ is not None, 'In get_cycle_dir but this ' +
                                  'should not be called if the task does not receive model.')
 
-        # Combine datetime string (directory format) with the model
-        cycle_dir = os.path.join(self.experiment_path(), 'run',
-                                 self.__datetime__.string_directory(), self.__model__)
+        # Check whether to send to cycle dir
+        if self.config.use_cycle_dir(True):
```
ashiklom, line 207: Why do you pass `True` here? (1) That should already be the default value in the dataclass. (2) If we have to pass it, (2a) I suggest adding a comment explaining why, and (2b) pass it as a named argument to make it clear what you're setting (`self.config.use_cycle_dir(default_value=True)`).

src/swell/tasks/base/task_base.py
```diff
@@ -275,7 +280,7 @@ def create_task(
         config: str,
         datetime: Union[str, dt, None],
         model: str,
-        ensemblePacket: Optional[str]
+        ensemblePacket: Optional[str],
```
ashiklom: Extra trailing comma?

src/swell/tasks/base/task_base.py
```diff
@@ -313,7 +319,7 @@ def get_tasks() -> list:
     tasks = []
     for task_file in task_files:
         base_name = os.path.basename(task_file)
-        if '__' not in base_name:
+        if '__' not in base_name and base_name not in ['task_attributes.py']:
```
ashiklom: Why `not in` here? Why not just `base_name != "task_attributes.py"`?

src/swell/tasks/base/task_attributes.py
```diff
@@ -0,0 +1,865 @@
...
+                                             {'datetime': '$CYLC_TASK_CYCLE_POINT',
+                                              'config': '$CYLC_SUITE_DEF_PATH/experiment.yaml'})]  # noqa
+
+    # --------------------------------------------------------------------------------------------------
+
+    class BuildGeos(Task):
+        def set_attributes(self):
+            self.question_list = QuestionList([
+                qd.geos_build_method()
+            ])
```
ashiklom: Can you explain this design choice?

In my view, we should strive for [locality of behavior](https://htmx.org/essays/locality-of-behaviour/). In this context, I would prefer to have all information like this be local to the definition of the Task, not scattered across multiple places. Specifically, it feels more natural for the BuildGeos task attributes to reside somewhere in `.../tasks/build_geos.py`, maybe in an `attributes` method/container in the class.

src/swell/utilities/cylc_formatting.py
```diff
@@ -0,0 +1,96 @@
...
+from collections.abc import Mapping
+import textwrap
+
+INDENT = ' ' * 4
+
+# --------------------------------------------------------------------------------------------------
+
+
+def format_dict(dictionary: Mapping):
+    # Convert a dictionary into a string
```
ashiklom: A good habit to start forming is writing properly formatted Python [docstrings](https://realpython.com/how-to-write-docstrings-in-python/) rather than just comments. These give you access to the documentation via Python's `help()` function (e.g., `help(format_dict)`, object attributes (e.g., `format_dict.__doc__`), and I _think_ in suggestion popups in VSCode and other (properly configured) text editors.

My suggestion would be to use [Google's docstring format](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods), which is a popular choice.

If you want to be a bit lazy, AI tools like ChatGSFC or GitHub Copilot are pretty good at writing these for you (or, at least, creating first drafts).

Example here (generated by AI from the function definition! But with some light editing on my part):

```python
def format_dict(dictionary: Mapping):
    """Convert a dictionary into a string.

    Args:
        dictionary (Mapping): The dictionary to format.

    Returns:
        str: A string representation of the dictionary, with each key-value pair on a new line in the format 'key = value'.

    Examples:
        >>> format_dict({'a': 1, 'b': "test"})
        'a = 1\nb = test\n'

        # NOTE: Strings are not quoted
        >>> print(format_dict({'a': "1", 'b': "test"}))
        a = 1
        b = test

        # NOTE: Nested dictionaries are printed in native dict/JSON format
        >>> print(format_dict({'a': "this", 'b': {"b1": 1, "b2": 2}}))
        a = this
        b = {'b1': 1, 'b2': 2}

        >>> format_dict({})
        ''
    """

  dict_str = ''
  # [...]
```

src/swell/utilities/cylc_formatting.py
```diff
@@ -0,0 +1,96 @@
...
+    else:
+        level -= 1
+
+    content = section.content
+    if isinstance(content, Mapping):
+        content = format_dict(content)
+
+    section_str += indent_lines(content, level+1)
+
+    return section_str
```
ashiklom: Since this function is specific to the `CylcSection` class, it should be a method of that Class.

src/swell/utilities/cylc_formatting.py
ashiklom: You don't need extensive docstrings for _every_ function. And incorrect documentation is always *worse* than no documentation at all! But, even in relatively simple cases like this, it's sometimes useful to be explicit. E.g., It may not be immediately obvious to everyone reading this code that (1) dictionaries with multiple entries will be printed on adjacent new lines; (2) neither keys, nor values, will be quoted, even if they are strings; and (3) nested dictionary entries will be rendered _differently_, in dict/JSON format (with quotes and all).

ashiklom: Here's a more extensive Google docstring example.

https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

Another nice feature of properly formatted docstrings is that they can be automatically converted into documentation websites later by tools like sphinx.

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+
+@dataclass
+class Task:
+
+    '''
+    Contains the basic properties and information needed to format the cylc [runtime] section.
+    '''
```
ashiklom: Echoing my comment above re: properly formatted docstrings. In those links, there are examples of documenting class attributes.

Consider something like this:

```python
@dataclass
class Task:

    """Properties and information needed to format the cylc [runtime] section.

    Attributes:
        base_name (str, optional):
            Name of the task. If `None`, defaults to class name.
        scheduling_name (str, optional):
            Name to use for SLURM `job_name` parameter. If `None`, defaults to value of `base_name`.
        model (str, optional):
            If the task is model-specific, the name of the model. If `None`, <document behavior...>

    """
```

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+    def match_platform(self, content: Union[str, dict], platform: str):
+        # Resolve platform-specific entries in the task object
+
+        if isinstance(content, Mapping):
+            if platform in content.keys():
+                content = content[platform]
+            elif 'all' in content.keys():
+                content = content['all']
+
+        return content
```
ashiklom: The logic in here is a bit awkward. IMHO, a bit clearer to return a dedicated result object:

```python
if not isinstance(content, Mapping):
  return content

if platform in content.keys():
  return content[platform]
elif 'all' in content.keys():
  return content['all']
else:
  raise ValueError(f"Unexpected information in `content` dict: {content}")

```

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+            slurm_dict = update_dict(slurm_dict, slurm_dict['all'])
+            del slurm_dict['all']
+        if self.model in slurm_dict.keys() and isinstance(slurm_dict[self.model], Mapping):
+            slurm_dict = update_dict(slurm_dict, slurm_dict[self.model])
+
+        for model in get_model_components():
+            if model in slurm_dict.keys():
+                del slurm_dict[model]
+
+        return slurm_dict
```
ashiklom: Similar comment here. You have to be a little careful with Python dicts because they are passed by reference, not by value.

```
>>> a = {"x": 3}
>>> b = a
>>> b["y"] = 7
>>> print(a)
{'x': 3, 'y': 7}
```

A safer pattern here might be explicitly create and populate a _new_ output object; something like:

```python
result = update_dict(slurm_dict, slurm_dict["all"]
del result["all"]
# Etc...
```

As long as we always call this function as something like `slurmdict = obj.resolve_model(slurmdict)` then it shouldn't really matter. But it may be safer to explicitly create a new object to avoid confusion related to mutating the argument.

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+        slurm_dict = {'job-name': self.scheduling_name,
+                      **self.resolve_model(slurm_globals),
+                      **self.resolve_model(slurm_dict),
+                      **self.resolve_model(slurm_task)}
+
+        return slurm_dict
+
+    # --------------------------------------------------------------------------------------------------
+
+    def runtime_string(self, experiment_dict: Mapping, slurm_external: Mapping):
```
ashiklom: Add typehint for expected output.

```suggestion
    def runtime_string(self, experiment_dict: Mapping, slurm_external: Mapping) -> str:
```

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+        ''' Return the runtime section for the given task. '''
```
ashiklom: Since this is a moderately complex "workhorse" function, I suggest a longer docstring here describing (briefly) the major things this function does.

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+
+            runtime_section.add_subsection(directive_section)
+
+        # Append additional sections to runtime
+        for section in self.additional_sections:
+            runtime_section.add_subsection(section)
+
+        # Check slurm messaging parameters
+        events = []
+        events = self.mail_events
```
ashiklom: The first line here is unnecessary since it's instantly overridden by the second line?

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # Add messaging section
+        settings_file = os.path.expanduser(os.path.join('~', '.swell', 'swell-settings.yaml'))
+        if os.path.exists(settings_file) and len(events) > 0:
+            with open(settings_file, 'r') as f:
+                settings_dict = yaml.safe_load(f)
```
ashiklom: This feels conceptually out of place here. Ideally, we should parse all the various settings information (including workflow YAML, overrides, and user/global config files) that Swell uses once relatively early in a Swell job, resolve them into a settings object, and then use that settings object consistently throughout the workflow, with no further reference to things like the user swell settings file.

src/swell/utilities/cylc_runtime.py
```diff
@@ -0,0 +1,243 @@
...
+        elif self.time_limit:
+            time_limit = self.match_platform(self.time_limit, platform)
+            runtime_dict['execution time limit'] = time_limit
+
+        # Set the retry if this task needs it
+        if self.retry:
+            if self.retry is True:
+                retry = '2*PT1M'
+            else:
+                retry = self.match_platform(self.retry, platform)
```
ashiklom: Is there a better place we could put these defaults and logic? If possible, I would prefer this function do basically 2 things: Put the sections in the right order and then convert them to a string. If it *also* sets some defaults, it's starting to strain the [Single Responsibility Principle (SRP)](https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html).

It's easy to go overboard with SRP, like many of the SOLID principles. But a simpler way to think about this is: If I asked a new Swell developer where they might expect to find a default value, would you expect them to look in the function/method whose job it is to render a configuration into a string?

src/swell/suites/base/cylc_workflow.py
```diff
@@ -0,0 +1,89 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+
+# --------------------------------------------------------------------------------------------------
+
+from typing import Tuple
```
ashiklom, line 10: Even as of Python 3.9+ (which is past EOL; as of today, Python 3.10 is the earliest supported version) you can just use built-in typing generics; i.e., just things like `tuple[list, dict]`, without having to import `typing.Tuple`. So you can drop this import and simplify this wherever it appears.

src/swell/suites/base/cylc_workflow.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # --------------------------------------------------------------------------------------------------
+
+    @abstractmethod
+    def set_tasks(self) -> None:
+        pass
+
+    # --------------------------------------------------------------------------------------------------
+
+    def get_independent_and_model_tasks(self) -> Tuple[list, dict]:
```
ashiklom, line 64: See comment above. You can drop the `typing.Tuple` import and just do:

```suggestion
    def get_independent_and_model_tasks(self) -> tuple[list, dict]:
```

src/swell/utilities/question_defaults.py
```diff
@@ -9,7 +9,7 @@
 
 
 from dataclasses import dataclass
-from typing import List, Dict
+from typing import List, Dict, Union
```
ashiklom: FYI: As of Python 3.10 (the earliest supported Python version now that Python 3.9 is EOL), we can use the `|` operator for more concise versions of this.

E.g.,

```python
default_value: str | dict = "something"
```

src/swell/utilities/question_defaults.py
```diff
@@ -1262,6 +1284,16 @@ class swell_static_files_user(TaskQuestion):
 
     # --------------------------------------------------------------------------------------------------
 
+    @dataclass
+    class task_email_parameters(TaskQuestion):
+        default_value: Union[str, dict] = "auto"
```
ashiklom: Here and elsewhere: If there are only very specific strings with specific meanings (like `"auto"`), a better type annotation uses `typing.Literal`:

```python
from typing import Literal
# ...
default_value: Union[Literal["auto"], dict] = "auto"
```

This will throw a type error for any string value that isn't *precisely* the string `"auto"`.

src/swell/utilities/task_specification.py
```diff
@@ -0,0 +1,251 @@
...
+
+from swell.utilities.cylc_formatting import CylcSection, indent_lines
+from swell.utilities.suite_utils import get_model_components
+from swell.utilities.dictionary import update_dict
+from swell.utilities.swell_questions import QuestionList
+
+# --------------------------------------------------------------------------------------------------
+
+
+class Task:
```
ashiklom: Lot of overlap here with `cylc_runtime.py`. Duplication is not necessarily a bad thing if we expect these two things to evolve quasi-independently. But if we fully expect all the rendering in both places to be the same, better to define it once and reuse implementations somehow...unless that's really awkward for some reason, in which case it may be OK to keep these duplicated for now.

ashiklom (changes requested):
@mranst See comments above. Some things are minor aesthetic/documentation things, and some of those can be considered optional (or punted to future PRs).

The biggest non-aesthetic thing to think about is the design of `task_attributes.py`. Ideally, I would love to have as much task-specific information as possible --- including the stuff in this file --- live in (or next to) the task definitions. If there are technical reasons that's hard or undesirable, it would be good to know. Let me know if you'd like to discuss.

Sorry this is a lot of comments. But, with a change of this size, complexity, and importance, I figured it's important to try to get as much right as we can the first time, to avoid another big change in the future.

src/swell/utilities/cylc_runtime.py
mranst: 👍 Moved to post_init

src/swell/utilities/task_specification.py
mranst: I thought I had removed `cylc_runtime.py` in the refactor, sorry about that. This file is actually the one being used now, though the logic should still be the same.

src/swell/utilities/cylc_runtime.py
mranst: I've added email address to the questions, with an additional `defer_to_user` key that's parsed in the experiment creation. We could potentially expand the functionality of this to essentially act as an override file

mranst:
> The biggest non-aesthetic thing to think about is the design of `task_attributes.py`. Ideally, I would love to have as much task-specific information as possible --- including the stuff in this file --- live in (or next to) the task definitions. If there are technical reasons that's hard or undesirable, it would be good to know. Let me know if you'd like to discuss.

Thanks for the thorough review, completely agree since this changes so much. I've experimented with storing the task attributes in the individual files, and while it's doable it does introduce some messiness of its own (all of which is addressable). The biggest issue is the way modules are handled by the tasks right now. Modules like `eva` and `r2d2` are loaded at the top of the task files currently, so these would need to be handled differently if we want to store the attributes in those files.  These are not loaded at the time of experiment creation because those installations are platform-specific, so you'll get an import error when trying to reference those files. The solution to this is probably quite simple, but I didn't necessarily want to touch that in this pr yet. The other (very minor) issue is that you need some way of referencing the task attributes object within the workflow. As far as I can tell there are two ways of handling this if you have the attributes in each file:

1. Dynamically import the modules by globbing the task files and linking them to an object
2. Manually link to the objects in `task_attributes.py`

One thing I like about the current `task_attributes` structure is that you have direct references to each task. It's a small thing but I like being able to see the connection in an IDE. In the first point, you handle things in the same way, you just lose the ability to see the direct link to the object. The second maintains that feature but is a little tedious. There may be another way, and this point is so minor I'm happy to just go with point 1, but I did want to put that out there

docs/examples/templating_workflows.md
```diff
@@ -0,0 +1,79 @@
...
+        self.questions = [
+            background_crtm_obs,
+            qd.marine_models(),
+            qd.observing_system_records_path(),
+            qd.window_offset(),
+            qd.marine_models(),
+        ]
+```
+
+Attributes are set by override the `set_attributes` method in `TaskSetup`. This has been combined with the previously-used `task_questions.py` for simplicity. Here, the tags `is_cycling` and `is_model` are used to specify what tags the task needs to be appended with in the runtime section. These are set to `False` by default. Tasks with a specified `slurm` dictionary (rather than set to null, as by default) will use their contents to build the `directives` section. For the task specification above for `EvaObservations`, the runtime section will be renderend as the following:
```
ghost: Suggested improvement for clarity.

Also, confirm the order of precedence between task-specific overrides here and user-specific overrides in the user slurm config file.

```suggestion
Attributes are set by override the `set_attributes` method in `TaskSetup`. This has been combined with the previously-used `task_questions.py` for simplicity. 

The tags `is_cycling` and `is_model` (both `False` by default) modify the script command (`swell task <task name> $config`):

- `is_cycling = True` adds `-d $datetime` for cycling tasks
- `is_model = True` adds `-m {model}` to indicate model-specific tasks.

The `slurm` attribute determines where or not the task requires Slurm and provides a way to set task-specific overrides:

- `slurm = None` means the task is not a Slurm task, so no `[[[directives]]]` section will be written.
- `slurm = {}` means the task *is* a Slurm task, so the `[[[directives]]]` will be populated according to the platform's default slurm settings (in `src/swell/deployment/platforms`) along with user-specific overrides.
- `slurm = {<arguments>}` will optionally override the platform defaults with task-specific ones (but note that *user-configured overrides always have the highest priority*).

For the task specification above for `EvaObservations`, the runtime section will be renderend as the following:
```

src/swell/deployment/create_task_config.py
```diff
@@ -0,0 +1,225 @@
...
+    if task.slurm is not None:
+        task_slurm_dict = task.generate_task_slurm_dict(slurm_external_dict, platform)
+
+        time_limit = task.get_time_limit()
+        if time_limit is not None:
+            time_limit_dto = isodate.parse_duration(time_limit)
+            task_slurm_dict['time'] = isodate.strftime(time_limit_dto, '%H:%M:%S')
+
+    else:
+        task_slurm_dict = {}
```
ghost: Two minor nits:

1. `task_slurm_dict = None` is an easier and clearer default to check for later, rather than always having to do `iflen(task_slurm_dict) > 0`. Unless this breaks `template_string_jinja2`?
2. Since `slurm_external_dict` is only used by `generate_task_slurm_dict`, I would move it inside the `if task.slurm is not None` block, to more clearly signal the intent.
3. I personally find it (very slightly) clearer and more concise to remove the `else` block and just do:

```python
task_slurm_dict = None
if task.slurm is not None:
  slurm_external_dict = prepare_slurm_defaults(...)
  task_slurm_dict = task.generate(...)
```

shiklomanov-an:
> Modules like eva and r2d2 are loaded at the top of the task files currently, so these would need to be handled differently if we want to store the attributes in those files. These are not loaded at the time of experiment creation because those installations are platform-specific, so you'll get an import error when trying to reference those files.

While somewhat less common than loading everything at the module root, it's perfectly reasonable for these kinds of modules to be imported locally. For example, `xarray` doesn't import `h5netcdf` until the relevant class is initialized (https://github.com/pydata/xarray/blob/main/xarray/backends/h5netcdf_.py#L133).

> The other (very minor) issue is that you need some way of referencing the task attributes object within the workflow. As far as I can tell there are two ways of handling this if you have the attributes in each file:
>
>    1. Dynamically import the modules by globbing the task files and linking them to an object
>    2. Manually link to the objects in task_attributes.py

There are a few other patterns you can consider (with some AI help from Google Gemini!):

3. You can first define a central "registry" in `task_attributes.py` and then register the classes in that registry when they are created.

```python
# task_attributes.py
class TaskAttributes:
    def __init__(self):
        self._classes = {}

    def register(self, name):
        def wrapper(cls):
            self._classes[name] = cls
            return cls
        return wrapper

    def get(self, name):
        return self._classes.get(name)

task_attributes = TaskAttributes()

# some_task.py
from .task_attributes import task_attributes
# NOTE: ^^ The instance, not the class

@task_attributes.register("SomeTask")
class SomeTask:
    pass
    
# usage
from task_attributes import task_attributes
obj = task_attributes.get("SomeTask")
```

4. Somewhat related: You can use the `__init_subclass__` method to treat the parent class as the registry. If you add a method like this to the `TaskSetup` base class...

```python
class TaskSetup:
    _registry = {}

    # [...]
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        # Store class by its name (or a custom attribute)
        cls._registry[cls.__name__.lower()] = cls
```

...then you can get a dictionary of all instances of that class with `TaskSetup._registry["SomeTask"]`.

5. You can also use the `__subclasses__` method to get a list of all subclasses of a parent class (https://stackoverflow.com/questions/3862310/how-to-find-all-the-subclasses-of-a-class-given-its-name). So `TaskSetup.__subclasses__` contains a list of all the currently defined sub-classes (at the time `__subclasses__` is called). You can iterate over that to build a dict of classes accessible via their names. Note however that this only gives a list of direct children of the class, not "grandchildren" (sub-subclasses).

Of these approaches, I think the best one is probably 4 (`__init_subclass__`) because it works for all descendant classes (not just direct children) and gives maximum control over the behavior (e.g., we can construct exactly the object we want).

A catch to these approaches is that for them to work you have to first import the modules where the classes are defined to run the definitions and corresponding "hooks". A simple script that satisfies this looks something like this (AI-generated):

```python
import pkgutil
import importlib
import plugins  # The package containing your classes
from framework.registry import manager

def discover_plugins(package):
    """Dynamically imports all modules in a package."""
    for loader, module_name, is_pkg in pkgutil.walk_packages(package.__path__):
        # Construct the full path (e.g., 'plugins.apple')
        full_module_name = f"{package.__name__}.{module_name}"
        
        # Import the module. This executes the code in the file, 
        # which triggers the @manager.register decorator.
        importlib.import_module(full_module_name)
        print(f"Loaded: {full_module_name}")

# 1. Run discovery
discover_plugins(plugins)

# 2. Now the registry is full!
print(f"Registered classes: {list(manager._classes.keys())}")
```

mranst:
Setup declarations have been put in the main task files. I went with method 3 as that seemed to work best. Method 4 seemed to register the inherited classes under each instance that TaskSetup was imported under, so it didn't work to centrally track them. Method 5 worked but there wasn't a great way to robustly handle the naming of the classes without initializing them. I moved the external imports (eva, r2d2, jedi_bundle) into the execute methods of the tasks that need them, but there may be a better way of handling this

src/swell/suites/base/all_suites.py
```diff
@@ -0,0 +1,109 @@
...
+                        import_module(f'swell.suites.{suite}.workflow'), f'Workflow_{suite}')
+
+                workflow_dict[suite] = workflow
+
+        self.workflow_dict = workflow_dict
+
+    def get_workflow(self, suite: str) -> CylcWorkflow:
+        return self.workflow_dict[suite]
+
+    def all_workflows(self) -> list:
```
ghost:
```suggestion
    def all_workflows(self) -> dict_keys:
```

`dict.keys()` doesn't return a `list` but rather a special iterable `dict_keys`.

src/swell/suites/base/all_suites.py
```diff
@@ -0,0 +1,109 @@
...
+            workflow_path = os.path.join(get_swell_path(), 'suites', suite, 'workflow.py')
+            if os.path.exists(workflow_path):
+                workflow = getattr(
+                        import_module(f'swell.suites.{suite}.workflow'), f'Workflow_{suite}')
+
+                workflow_dict[suite] = workflow
+
+        self.workflow_dict = workflow_dict
+
+    def get_workflow(self, suite: str) -> CylcWorkflow:
```
ghost:
```suggestion
    def get_workflow(self, suite: str) -> type[CylcWorkflow]:
```

src/swell/deployment/create_task_config.py
```diff
@@ -0,0 +1,223 @@
...
+        override = {}
+
+    if model is not None:
+        override['model_components'] = [model]
+    else:
+        override['model_components'] = []
+
+    # Build in current working directory
+    if cwd:
+        override['experiment_root'] = os.getcwd()
```
ghost: This will fail if you pass any string argument to `override` (as annotated in the function definition) because then `override` is not `None` (so the `override = {}` will not be run), but subsequent calls to `override['model_components']`, etc. will not work if `override` is a string.

(Note that I caught this thanks to the type checker ([ty](https://docs.astral.sh/ty/)) in my editor. A good example of types being useful for catching errors!)

src/swell/swell.py
```diff
@@ -115,6 +127,40 @@ def create(
...
+
+@swell_driver.command()
+@click.argument('task', type=click.Choice(get_tasks()))
+@click.option('-p', '--platform', 'platform', default='nccs_discover_sles15',
+              type=click.Choice(get_platforms()), help=platform_help)
+@click.option('-d', '--datetime', 'datetime', default=None, help=datetime_help)
+@click.option('-m', '--model', 'model', default=None, help=model_help)
+@click.option('-i', '--input_method', 'input_method', default='defaults',
+              type=click.Choice(['defaults', 'cli']), help=input_method_help)
+@click.option('-o', '--override', 'override', default=None, help=override_help)
```
ghost, line 152: `override` here is presumably a path to an override YAML file, right? If so, I suggest reading the YAML file into a dict in this file and then changing the `task_config_wrapper` to something like `override_dict[dict]` to make the distinction between a path to an override file and an override dict extremely clear.

(Ideal Python practice would be something like `override_struct[Override]`, where `Override` is a (data)class that has an `__init__` method that reads a YAML file, validates it, and then returns strongly typed attributes along with other helper methods as appopriate...but that's probably overkill for now).

src/swell/deployment/create_experiment.py
```diff
@@ -93,68 +93,98 @@ def prepare_config(
...
+    workflow = workflow_class(suite_dict, slurm_dict)
+
+    # Get the list of tasks from the workflow's graph
+    # -----------------------------------------------
+    model_ind_tasks, model_dep_tasks = workflow.get_independent_and_model_tasks()
+
+    # Set the tasks to be used in preparing the suite
+    # -----------------------------------------------
+    prepare_config_and_suite.model_independent_tasks = model_ind_tasks
+    prepare_config_and_suite.model_dependent_tasks = model_dep_tasks
```
ghost, line 185: Type checker complains about `unresolved attributes` here because these attributes are not instantiated in the `__init__` method of this class.

I suggest modifying `PrepareExperimentConfigAndSuite.__init__` to include something like this:

```python
# These get populated by <method/function>
self.model_independent_tasks = []
self.model_dependent_tasks = []
```

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -78,212 +78,192 @@ def __init__(
...
+        # --------------------------------------------
+        self.prepare_suite_question_dictionary()
+        self.override_with_defaults(QuestionType.SUITE)
+        self.override_with_external(QuestionType.SUITE)
+        self.ask_questions_and_configure(QuestionType.SUITE)
 
     # ----------------------------------------------------------------------------------------------
 
-    def prepare_question_dictionaries(self) -> None:
+    def configure_and_ask_task_questions(self) -> None:
```
ghost:
```suggestion
    def configure_and_ask_task_questions(self) -> tuple[dict, dict]:

File: src/swell/deployment/create_experiment.py (hunk 33 above)
Comment (ghost) [reply src/swell/deployment/create_experiment.py:185]: ...or use `{}` for these if they're supposed to be `dict`s.

File: src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py [hunk 35]
```diff
@@ -359,173 +344,70 @@ def override_with_external(self) -> None:
...
 
-        8. Ask the new task questions that do not actually depend on the model..
+        return out_dict
 
-        9.1 Build a list of tasks for each model component.
+    # ----------------------------------------------------------------------------------------------
 
-        9.2 Iterate over the model_dep dictionary and ask task questions.
-        """
+    def ask_questions_and_configure(self, suite_task: QuestionType) -> Tuple[dict, dict]:
```
ghost:
```suggestion

File: src/swell/tasks/base/task_attributes.py [hunk 36]
```diff
@@ -0,0 +1,75 @@
...
+        self.script = False
+        self.pre_script = "source $CYLC_SUITE_DEF_PATH/modules"
+        self.additional_sections = [self.create_new_section('environment',
+                                                            {'datetime': '$CYLC_TASK_CYCLE_POINT',
+                                                             'config': '$CYLC_SUITE_DEF_PATH/experiment.yaml'})]  # noqa
+
+
+class sync_point(TaskSetup):
+    def set_attributes(self):
+        self.script = "true"
```
ghost, line 53: What's up with the type instability of `self.script` here (and in `root` above)? In `TaskSetup`, it's either `None` or a string corresponding to the script. But here, we introduce boolean values (`False` above) and here it's a string `"true"` (is there a program called `true` that this is supposed to run...?). At a minimum, this needs a comment explaining what's going on. But ideally, we should think about the underlying design choices to avoid these kinds of kludges.

src/swell/tasks/base/task_attributes.py
```diff
@@ -0,0 +1,75 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+import importlib
+import pkgutil
+
+import swell.tasks
+from swell.tasks.base.task_setup import TaskSetup
+
+# --------------------------------------------------------------------------------------------------
```
ghost, line 14: I suggest adding some documentation (ideally, via file-level docstring, though just comments are fine too) on how this works and what it does. I can mostly follow the `TaskAttributes` class's behavior and the use of the `@task_attributes.register()` decorator, but it's worth documenting for future developers how these things work. I also suggest documenting that `root` and `sync_point` are swell-wide defaults for the corresponding fields in cylc (but which can be overridden by tasks).

src/swell/suites/localensembleda/flow.cylc
```diff
@@ -0,0 +1,238 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
```
ghost: Presumably, this was included by mistake, since we have a corresponding `workflow.py` file for this?

src/swell/tasks/base/task_base.py
```diff
@@ -48,6 +47,8 @@ def __init__(
         # ---------------------
         self.logger = get_logger(task_name)
 
+        from swell.utilities.config import Config
```
ghost, line 50: To help future developers (e.g., to make sure an overzealous new dev doesn't move this import into the module header), suggest adding a comment about why the local import is needed here.

src/swell/tasks/base/task_setup.py
```diff
@@ -0,0 +1,352 @@
...
+    pre_script: cylc setting for scripts run before the main script
+    script: string of shell code to be run by cylc for the task
+    retry: times * time interval cylc should retry the task, e.g. 2*PT10s
+    time_limit: execution time limit for slurm
+    slurm: dictionary of slurm parameters
+    mail events: list of events for email messaging through cylc
+    question_list: list of questions keys used by the task
+    additional_sections: list of additional CylcSection objects to append to the runtime section
+    '''
+
```
ghost, line 48: Given how important this class is to Swell machinery, and how useful type checkers are at catching unexpected behavior early, I suggest adding some class-level type annotations for the attributes here (for the ones that aren't arguments to `__init__`). You can do that with code like this:

```python
class TaskSetup(ABC):
    '''
    <documentation...>
    '''
    base_name: str | None
    scheduling_name: str | None

    is_cycling: bool
    is_model: bool

    pre_script: str | Literal[False]

    # etc...
```

This should allow type checkers to flag code that violates these types.

src/swell/tasks/base/task_setup.py
```diff
@@ -0,0 +1,352 @@
...
+        self.set_attributes()
+        self.post_init()
+
+    # --------------------------------------------------------------------------------------------------
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+
+        cls._registry[cls.__name__.lower()] = cls
```
ghost: Definitely add a (brief) docstring to this method explaining what it does for the benefit of future developers.

src/swell/tasks/build_geos.py
```diff
@@ -27,15 +42,19 @@ def execute(self) -> None:
         swell_exp_path = self.experiment_path()
         geos_gcm_path = os.path.join(swell_exp_path, 'GEOSgcm')
 
+        from swell.utilities.build import build_and_source_dirs
```
ghost, line 45: Add comment explaining why local import is needed here.

src/swell/tasks/build_geos_by_linking.py
```diff
@@ -26,6 +43,8 @@ def execute(self) -> None:
         swell_exp_path = self.experiment_path()
         geos_gcm_path = os.path.join(swell_exp_path, 'GEOSgcm')
 
+        from swell.utilities.build import build_and_source_dirs, link_path
```
ghost, line 46: Ditto here.

src/swell/tasks/build_jedi.py
```diff
@@ -22,11 +38,15 @@ class BuildJedi(taskBase):
 
     def execute(self) -> None:
 
+        from jedi_bundle.bin.jedi_bundle import execute_tasks, get_bundles
```
ghost, line 41: ...and here (and throughout, wherever local imports are used).

src/swell/tasks/jedi_c_test.py
```diff
@@ -1,52 +0,0 @@
-# (C) Copyright 2021- United States Government as represented by the Administrator of the
```
ghost: Did we decide to get rid of this? Fine if yes (less code is almost always a good thing!) -- just wanted to confirm.

shiklomanov-an (changes requested):
Nicely done! I really like how this is looking. I think having the task parameters in the same files as the task definitions will be very useful.

I flagged only a few (hopefully minor) things.

* Some of these are type annotation issues flagged by my type checker.
* There's also what looks like a bug related to `override` as a file path vs. a dictionary.
* Finally, there are a few documentation requests for implementation details and design decisions, to help future developers (since we're using slightly esoteric Python machinery to get this working).

(Re: Local import documentation --- if we're just using local imports for consistency across modules, that's totally acceptable and you can disregard my comments about those...but maybe add a line to the documentation somewhere about why we're using that general pattern, with reference to specific cases like `eva` and `r2d2`).

@Dooruk @rtodling  -- Take a look the new task definition structure (any of the files under `tasks`) and let me know what you think. Don't worry too much about the underlying machinery; I'm more interested on what you think of this overall structure (compared to having all the task questions sit in a single giant YAML file / Python script like we had before; https://github.com/GEOS-ESM/swell/issues/314).

src/swell/suites/base/all_suites.py
ghost: ...or just convert the output of the command to a `list`.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
ghost: Whoops -- malformed suggestion. Just check the type annotation here -- I think the function should return `None`.

src/swell/tasks/base/task_base.py
ghost: ...or, per my comment in the overall review, just document once in some shared location why we generally prefer local imports for this. Then, you can resolve this and all related comments below.

src/swell/swell.py
mranst: Agreed, the override methodology should probably be restructured going forward

src/swell/tasks/base/task_attributes.py
mranst: `sync_point` is essentially a blank task that's treated as a dependency node in the cylc graph. Since swell enforces `allow implicit tasks = False`, it needs to have a runtime entry, so it just passes `true` to the shell as a placeholder

src/swell/tasks/base/task_base.py
mranst: Added lines explaining this for all imports, I wasn't sure where to put this in the docs

src/swell/tasks/jedi_c_test.py
mranst: Yes, it's not being used

Dooruk:
Wow, this is some impressive amount of change. Really appreciate the documentation, it would be quite confusing without it.

When I first started using SWELL, I was able to include config variables inside the task scripts without the `task_questions.yaml` and `suite_questions.yaml` interference. Explaining that concept to the new users has always been challenging. Overall, I'm happy with the vision and design decisions here. Previous method made it hard to introduce new developers, hope this will simplify the SWELL entry point.

Overall, I'm happy with these changes. In the previous workflow templating iteration `flow.cylc` handling was really confusing, and this approach makes it more straightforward.

Abandoning single task_questions and suite_questions is a welcome change. Users would also see `QuestionDefaults` class is coming from `utilities` so that multi-file ambiguity is inherently handled. 

Since we have a limited number of seasoned users, I would be curious to hear from Yonggang and Maryam too, once this is ready for that stage.

Some minor questions/details/nitpicks:

1) Is this ready for testing yet?

2) How can `workflow.py`  handle non `\tasks` specified scripts? Is this PR limiting us with python-only scripts such as: 
- (More of a general comment but going forward) Could this new method handle, say, bash scripts?
- Or calling outside scripts directly, like so:
```
    [[RunGeos]]
        script = "{{experiment_path}}/GEOSgcm/forecast/gcm_run.j"
        platform = {{platform}}
        [[[job]]]
            shell = /bin/csh
```

-  Model specific tasks organized under `\tasks\model`, following Michael's PR.


3) Documentation states that `[scheduling]` section of  `flow.cylc` remains the same, that means that part is still meant to handle conditional tasks and loops since jinja templating parses it?

4) What are "messaging parameters"?

5) I'm going to sound very pedantic and annoying, but `is_model`  switch might be better named `model_specific` or `model_dep`? Considering how widespread this switch is I can live without this change :) In a similar vein, `time_limit` -> `task_time_limit`.

mranst:
@Dooruk Thanks! To answer your questions:

1) Yes, this should be ready for testing. It's been a week or two but this PR has passed tier1 and tier2 previously
2) The runtime section generation is geared to swell tasks for convenience, but you can easily run any kind of script by overriding the `TaskSetup` object. Your example would look like this: (you can just put this in the workflow.py file and reference it there, it doesn't need to be it's own task)
```python
class GcmRunJ(TaskSetup):
    def set_attributes:
        self.base_name = "RunGeos"
        self.script = "{{experiment_path}}/GEOSgcm/forecast/gcm_run.j"
        self.slurm = {}
        self.additional_sections.append(
             self.create_new_section('job', {'shell': 'csh'})
        )

class Workflow:
...

    def set_tasks(self) -> list:
        self.tasks.append(GcmRunJ)
```
For development purposes, there's also no reason you couldn't just add the section in plain text to the scheduling
https://github.com/GEOS-ESM/swell/blob/dabe9b2c33e89775b69b7dd3642649dc340d5e58/src/swell/suites/3dvar/workflow.py#L101-L107

^ right here, for instance.

3) Basically nothing is different in the scheduling section, cylc still handles it in the exact same way once `flow.cylc` is generated.

In all workflows I maintained that the scheduling section is templated using jinja, like so:
https://github.com/GEOS-ESM/swell/blob/dabe9b2c33e89775b69b7dd3642649dc340d5e58/src/swell/suites/3dvar/workflow.py#L111-L122

 So that section of file construction is still handled in the same jinja logic. But technically, you can set the `set_workfow_string` method in any way you want, so there's room to get more advanced here if we wish

4) Messaging parameters are the events for which cylc will email the user if the suite is launched with the `-m` flag. The relevant ones are `failed` and `submit-failed`. The ones that are designed to fail normally only have `submit-failed` as a option to not annoy the user. Suites can have `succeeded` which might be relevant as well.

5) Good call, I'll rename them

mranst:
I refactored some machinery with the configs I've come to dislike recently, it's implemented similarly to the task_attributes in this PR in the sense that it requires registry for the lists, but it's more dynamic and doesn't depend as much on parsing filenames and the general naming of things. Recent commits also add easy support for setting suite-dependent slurm parameters 

This is ready for testing, and all tier1 tests are passing for this. I'd like to get this in soon if possible, as merging develop into this can be an extensive process due to how many files this touches, but we can wait for PRs like #716 and #677 to merge before shaking things up

mranst:
Bringing in a few more eyes to this PR now that we have some more people familiar with how Swell works. This is a complex PR that changes a lot of how swell is constructed, so please let me know if you have any questions or want a more detailed description on anything. I wrote some docs explaining how the new workflow templating works: https://github.com/GEOS-ESM/swell/blob/se/mranst/workflow_templating_redux2/docs/examples/templating_workflows.md.

To summarize the advantages introduced by this PR:

1. More flexibility in constructing workflows by allowing more dynamic python logic within each suite's setup
2. With the above comes easier implementation of features to swell, for example this PR also introduces email functionality for tasks and the ability to set tasks to pause on
3. Less rigid construction. This PR moves task question assignments from `task_questions.py` to individual task files.

Should we choose to merge this PR, I know it is bound to affect incoming PR's that are based on the old system. I am happy to assist in merging branches with this system to make this as painless as possible

Dooruk:
Hi all, could you take an ~moment~ hour to review this PR? @mranst put in a lot of effort into this and he is having to modify it with every new tasks/suites added to SWELL. I will do the same 😄

jeromebarre:
I will try to review it over the next few days.

src/swell/suites/3dvar_cf/workflow.py
```diff
@@ -88,72 +101,52 @@
...
-        script = "swell task CleanCycle $config -d $datetime -m {{model_component}}"
-    {% endfor %}
+
+'''  # noqa
+
+# --------------------------------------------------------------------------------------------------
+
+
+@workflows.register('3dvar_cf')
+class Workflow_hofx_cf(CylcWorkflow):
```
jeromebarre: it should be `Workflow_3dvar_cf` here and not `Workflow_hofx_cf`

jeromebarre:
which suites have been tested with this branch? I will go over the CF ones. Has the suite CI been run on this?

src/swell/suites/r2d2_ingest/workflow.py
```diff
@@ -56,47 +63,45 @@
...
-        execution time limit = PT10M
-
-    {% endfor %}
+
+'''  # noqa
+
+# --------------------------------------------------------------------------------------------------
+
+
+@workflows.register('r2d2_ingest')
```
jeromebarre, line 72: there is also a mismatch here between the new and old name of this suite

src/swell/suites/r2d2_ingest/workflow.py
```diff
@@ -56,47 +63,45 @@
...
+class Workflow_ingest_obs(CylcWorkflow):
```
jeromebarre: it should be 
```suggestion
class Workflow_r2d2_ingest(CylcWorkflow):
```

src/swell/suites/hofx/workflow.py
```diff
@@ -94,79 +105,55 @@
...
+
+        return workflow_str
+
+    def set_tasks(self) -> list:
+
+        self.tasks.append(ta.root())
+        self.tasks.append(ta.CloneJedi())
+        self.tasks.append(ta.BuildJedi())
+        self.tasks.append(ta.BuildJediByLinking())
+        self.tasks.append(ta.CloneGeosMksi())
```
jeromebarre: this line appears in the hofx suite but not in the 3dvar suites. I think it should be removed. As there is 
`self.tasks.append(ta.CloneGeosMksi(model=model))`
below anyway

src/swell/suites/forecast_coupled_geos/workflow.py
```diff
@@ -0,0 +1,127 @@
...
+    def set_tasks(self) -> list:
+
+        self.tasks.append(ta.root())
+        self.tasks.append(ta.CloneGeos())
+        self.tasks.append(ta.BuildGeosByLinking())
+        self.tasks.append(ta.BuildGeos())
+        self.tasks.append(ta.PrepCoupledGeosRunDir())
+        self.tasks.append(ta.GetCoupledGeosRestart())
+        self.tasks.append(ta.MoveForecastRestart())
+        self.tasks.append(ta.SaveRestart())
```
jeromebarre: this is not referenced anywhere in the flow cylc graph. either make it appear in the flow cylc graph or remove this line

src/swell/tasks/eva_observations.py
```diff
@@ -25,12 +26,39 @@
...
+        self.task_time_limit = True
+        self.is_cycling = True
+        self.model_dep = True
+        self.slurm = {}
+        self.questions = [
+            qd.background_time_offset(),
+            qd.crtm_coeff_dir(),
+            qd.observations(),
+            qd.observing_system_records_path(),
+            qd.marine_models(),
```
jeromebarre: duplicate question with line 57

src/swell/tasks/eva_observations.py
```diff
@@ -25,12 +26,39 @@
...
+            qd.observing_system_records_path(),
```
jeromebarre: duplicate question as well

src/swell/tasks/get_background.py
```diff
@@ -9,11 +9,37 @@
...
+@task_attributes.register(task_name)
+class Setup(TaskSetup):
+    def set_defaults(self):
+        self.base_name = task_name
+        self.is_cycling = True
+        self.model_dep = True
+        self.questions = [
+            qd.window_length(),
+            qd.window_type(),
+            qd.window_length(),
```
jeromebarre, line 36: duplicate question

src/swell/tasks/get_observations.py
```diff
@@ -8,22 +8,46 @@
...
+        self.is_cycling = True
+        self.model_dep = True
+        self.questions = [
+            qd.background_time_offset(),
+            qd.crtm_coeff_dir(),
+            qd.observations(),
+            qd.observing_system_records_path(),
+            qd.cycling_varbc(),
+            qd.obs_experiment(),
+            qd.observing_system_records_path(),
```
jeromebarre, line 44: duplicate question

src/swell/tasks/run_jedi_obsfilters_executable.py
```diff
@@ -10,19 +10,59 @@
...
+            qd.window_length(),
+            qd.window_type(),
+            qd.background_time_offset(),
+            qd.crtm_coeff_dir(),
+            qd.observations(),
+            qd.observing_system_records_path(),
+            qd.background_frequency(),
+            qd.generate_yaml_and_exit(),
+            qd.jedi_forecast_model(),
+            qd.observing_system_records_path(),
```
jeromebarre: duplicate question

mranst:
> which suites have been tested with this branch? I will go over the CF ones. Has the suite CI been run on this?

The tier1 tests have been run on recent versions of this and passed. At various points in working on this I've run all of the suites in swell, though I think it would be good for everyone to take a close look at the suites they run frequently

jeromebarre (approved):
Just tested the cf cycle suite and this worked. All good for me. Thanks for the massive amount of work.

Dooruk (approved):
Thanks, I reviewed this before I think so I don't want to hold it back any longer.

It would be nice if https://github.com/GEOS-ESM/swell/pull/855 can go in first but I don't mind either way.

PR #658, opened 2025-11-13
Automatically calculate window_offset from window_length

mranst:
## Description

This PR eliminates the `window_offset` and `analysis_forecast_window_offset` keys from experiment.yaml. These parameters are now calculated in `data_assimilation_window_params.py` using `window_length`. 

`background_time_offset` can also potentially be calculated from window_length but I wasn't sure of the relationship. From what I could tell they should have the following values by default:

3dvar:
window_length: P1D
background_time_offset: PT18H

3dvar_atmos:
window_length: PT6H
background_time_offset: PT9H

If it's just window_offset + 6 I can add that too if someone can confirm

Tier1 tests pass for this

### DIFF
--- a/docs/adding_a_suite.md
+++ b/docs/adding_a_suite.md
@@ -364,7 +364,6 @@ class SuiteConfig(QuestionContainer, Enum):
             qd.cycle_times(['T12']),
             qd.marine_models(['mom6']),
             qd.window_length("P1D"),
-            qd.window_offset("PT12H"),
             qd.horizontal_resolution("72x36"),
             qd.vertical_resolution("50"),
             qd.total_processors(6),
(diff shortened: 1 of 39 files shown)


Dooruk (commented):
This was a long term annoyance of mine, thanks!

I see some usage and `config` access in `utilities/check_da_params.py` but since that is not tested maybe it didn't trigger a warning?

`background_time_offset` seem wonky, must have been artifacts from the old R2D2. I can work that in my PR as it will require some restructuring of others tasks and making sure background files are there..

mranst:
> I see some usage and `config` access in `utilities/check_da_params.py` but since that is not tested maybe it didn't trigger a warning?

Good catch, thanks!

Dooruk (approved):
👍🏼

PR #662, opened 2025-11-25
Add sp to modules

### LINKED ISSUE #661: JEDI built by swell doesn't match JEDI built by jedi_bundle
This became clear when testing comparisons to be added to nightly tier2 tests. `3dfgat_atmos` has fairly significant discrepancies when run on a jedi build created by swell using the `build_jedi` suite (as is done in the tier2 tests) vs. a version built by `jedi_bundle` on pinned versions matching that date. This does appear to be limited to `3dfgat_atmos`, `3dfgat_cycle` does not feature this issue.

experiment 0: Experiment run on September 9 version built by `jedi_bundle`
experiment 1: Experiment run on version of jedi built 11/18 by `build_jedi` suite

Image

Image

To be clear, this doesn't appear to be due to an update to JEDI, I ran `3dfgat_atmos` on a version created by `jedi_bundle` for 11/18 and got the same results. 

Some things I've tried to fix this include updating the jedi_bundle used by swell, ensuring that all bundles are cloned correctly, making sure `CMakeLists.txt` matches between the two versions.

mranst:
This fixes the discrepancy between JEDI builds (#661) by adding sp to the modules so that Swell builds jedi correctly. Thanks to @mer-a-o for figuring out this issue.

This also matches the number of iterations properly in `JediLogComparison`.

### DIFF
--- a/src/swell/deployment/platforms/nccs_discover_cascade/modules
+++ b/src/swell/deployment/platforms/nccs_discover_cascade/modules
@@ -44,6 +44,10 @@ module load r2d2-client/112025
 module load eva/sles15_skylab9
 module load jedi_bundle/sles15_skylab9
 
+# Load sp
+# -------
+module load sp/2.5.0
+
 # Set the swell paths
 # -------------------
 PATH={{swell_bin_path}}:$PATH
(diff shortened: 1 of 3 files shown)


Dooruk (approved):
@mer-a-o for the W 👍🏼

mer-a-o (approved):
Changes look good to me. Thanks @mranst

PR #660, opened 2025-11-15
Adding geos_cf hofx experiment

mer-a-o:
## Description

Currently a draft PR. 
This PR adds the first GEOS-CF experiment, H(x) application, to SWELL.

Summary of major changes made in this PR:

* Added `geos_cf` to the interfaces, including the relevant input YAML files needed to run a geos_cf experiment and the observation/observation operator information under `interfaces/geos_cf`. The suite configuration is also added here.

* Added the `suites/hofx_cf` suite and all the relevant files needed to run this suite.

* Updated `tasks/get_observations.py`: The code now fetches a generic empty observation file that I ingested into R2D2 if r2d2.fetch fails for a particular observation/date. We need a more detailed discussion to determine whether this is the desired behavior.

* Updated `tasks/save_obs_diags.py`: I believe we no longer write feedback files with `_0000`, so there is no need to check for files with that suffix. Additionally, by passing `input_and_output=True to check_obs`, we can detect whether feedback and input files are missing or have zero dimensions and skip saving the feedback file accordingly. There is no need to check file dimensions again in this code.

* Added `npx` and `npy` instead of reading them from `horizontal_resolution`. The reason for this change is that `horizontal_resolution` is used for storing/fetching background files in R2D2 and may differ from `npx` and `npy`. For example, for `geos_cf`, we may set `horizontal_resolution = c90`, while `npx` and `npy` are set to 91.

## Dependencies
The PR depends on these issues:
https://github.com/GEOS-ESM/swell/issues/648 and PR: https://github.com/GEOS-ESM/swell/pull/651: MERGED
https://github.com/GEOS-ESM/swell/issues/645

## Impact
None

PR #668, opened 2025-12-09
breaking down PR/660 into smaller pieces -- part1

mer-a-o:
## Description

Add `npx` and `npy` to the list of questions. See PR description https://github.com/GEOS-ESM/swell/pull/660 for more details.

I don't think if there's any way to test the changes prior to adding `geos-cf` model SWELL. Code tests pass. 

## Dependencies

None

## Impact

None

### DIFF
--- a/src/swell/tasks/task_questions.py
+++ b/src/swell/tasks/task_questions.py
@@ -40,6 +40,8 @@ class TaskQuestions(QuestionContainer, Enum):
         questions=[
             qd.npx_proc(),
             qd.npy_proc(),
+            qd.npx(),
+            qd.npy(),
             qd.horizontal_resolution(),
             qd.vertical_resolution()
         ]
--- a/src/swell/utilities/question_defaults.py
+++ b/src/swell/utilities/question_defaults.py
@@ -1005,6 +1005,32 @@ class QuestionDefaults():
     # --------------------------------------------------------------------------------------------------
 
     @dataclass
+    class npx(TaskQuestion):
+        default_value: str = "defer_to_model"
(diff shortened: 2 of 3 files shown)


Dooruk:
Does how `geos_atmosphere` use `npx_proc` versus `geos_cf` use `npx` differ?

jeromebarre:
> Does how `geos_atmosphere` use `npx_proc` versus `geos_cf` use `npx` differ?

npx, npy is the number of cells +1 you have on a side on the edge of the cube and npx_proc, npy_proc is the number of procs you assign per cube edges.

jeromebarre:
In the process of rearranging this. I am leaning toward keeping `horizontal_resoluiton` to define `npx` and `npy` for now and have simply another variable that defines a `grid_name` for r2d2 storing purposes. We are discussing this with Maryam right now, to make sure this has no impact on the weather suite and functionalities.

mer-a-o:
@Dooruk `npx_proc` defines the `layout` for the mpi application which is different from horizontal resolution or number of grids. GEOS-CF will use `npx_proc` the same way as atmosphere applicaiton.

jeromebarre:
Tested this on the 3DVar atmos. This has no impact. It is just a matter on how we want to use the `horizontal_resolution` key. This might be ok to differ slightly between applications but could create confusion down the line.

jeromebarre (approved):

Dooruk (approved):
`defer_to_model` part will expect corresponding `nx` and `ny` in `geos_cf/task_questions.yaml` but since this is a multipart approach it is fine for now.

PR #669, opened 2025-12-10
breaking down PR/660 into smaller pieces -- part2, empty obs

mer-a-o:
## Description

See https://github.com/GEOS-ESM/swell/pull/660 for more details.

In this PR:
* Updated `tasks/get_observations.py`: The code now fetches a generic empty observation file that I ingested into R2D2 when r2d2.fetch fails for a particular observation/date. If needed we can add a key to `observation_ioda_names.yaml` to mark mandatory observations (the workflow should fail if the mandatory observation file do not exist). 

* Updated `tasks/save_obs_diags.py`: I believe we no longer write feedback files with `_0000`, so there is no need to check for files with that suffix. Additionally, by passing `input_and_output=True to check_obs`, we can detect whether feedback and input files are missing or have zero dimensions and skip saving the feedback file accordingly. There is no need to check file dimensions again in this code.

## Dependencies
None

## Impact
None


Dooruk:
Quick question on the `_0000` files. I haven't seen them in any context since I started this project, are they completely irrelevant? `eva_observations.py` also looks for those..

mer-a-o:
@Dooruk I believe so. You can write one feedback filesper mpi task using:

```
      io pool:
        max pool size: 6
        write multiple files: true
```

But by default one feedback file will be written per observation.

Dooruk:
So to repeat logic here, if a file is missing an empty file is created and `get_observations` doesn't fail. This has another impact, which is not obvious because of the `check_obs` method, the empty observers are not included in the final `jedi_*.yaml`. This happens in this obscure `iterator` method that @mranst addressed in #605: 

https://github.com/GEOS-ESM/swell/blob/c6e065f82f230f475b26c45f259876da6716428b/src/swell/utilities/run_jedi_executables.py#L100-L114

With this, we are not utilizing the `missing file action` key and hoping monitoring would pick up the problems. However if a particular observation type is imperative, say radiosondes, that can be set with `missing file action: error` ideally. 

What do you set the following in your geos_cf observation YAMLs, warn or error?:
```yml
missing file action: warn
```

src/swell/tasks/get_observations.py
```diff
@@ -182,6 +182,18 @@ def execute(self) -> None:
                     self.logger.info(f"Successfully fetched {target_file}")
                 except Exception as e:
                     self.logger.info(f"Failed to fetch {target_file}: {str(e)}")
+                    self.logger.info("Fetch empty observation instead.")
```
mranst: Minor suggestion for clarity
```suggestion
                    self.logger.info(f"Failed to fetch {target_file}. Fetch empty observation instead.")
```

mranst:
This passes tier1 tests for all suites except `3dvar`, which has a failure in `GetObs`

https://github.com/GEOS-ESM/swell/actions/runs/20822240523/job/59813801689#logs

image

mer-a-o:
Looking closer at the error, it looks like combining failed because out of 5 files for `sst_abi_g16_l3c.*.nc4` only one has data and the rest are empty files. Is this a common case?
I can change the empty file to include all the groups in the ioda format and my guess is that it will solve the issue.

mer-a-o:
@Dooruk, @mranst, I modified `read_and_combine` in get observation task to handle the cases where some files of specific observation type are missing. The `3dvar` case passes now.

mer-a-o:
> So to repeat logic here, if a file is missing an empty file is created and `get_observations` doesn't fail. This has another impact, which is not obvious because of the `check_obs` method, the empty observers are not included in the final `jedi_*.yaml`. This happens in this obscure `iterator` method that @mranst addressed in #605:
> 
> https://github.com/GEOS-ESM/swell/blob/c6e065f82f230f475b26c45f259876da6716428b/src/swell/utilities/run_jedi_executables.py#L100-L114
> 
> With this, we are not utilizing the `missing file action` key and hoping monitoring would pick up the problems. However if a particular observation type is imperative, say radiosondes, that can be set with `missing file action: error` ideally.
> 
> What do you set the following in your geos_cf observation YAMLs, warn or error?:
> 
> ```yaml
> missing file action: warn
> ```

I plan to make this change https://github.com/GEOS-ESM/swell/pull/660/files#diff-abd0eef9d7a487fe0025a5970de886cb116d5c08d822d6d4f125de8a21018cbc (I think it's better to be done in this PR), simplify the logic to write all the obs listed in the experiment.yaml in jedi.yaml.  

I didn't know about `missing file action: warn`. The default behavior is warning. I tested setting it to `error` for the 3dvar case with missing `sst_gmi_l3u` obsrvation. It still only gave me a warning in the case of empty observation existing:

```
WARNING: Input file /discover/nobackup/mabdiosk/SwellExperiments/swell-3dvar/run/20210701T120000Z/geos_marine/sst_gmi_l3u.20210701T000000Z.nc4 contains zero observations
sst_gmi_l3u: read database from /discover/nobackup/mabdiosk/SwellExperiments/swell-3dvar/run/20210701T120000Z/geos_marine/sst_gmi_l3u.20210701T000000Z.nc4 (io pool size: 6)
```

Dooruk:
I would like to get this one in next, all tier 1 tests are passing currently.

Ok, I read the [description](https://github.com/JCSDA-internal/jedi-docs/blob/develop/docs/inside/jedi-components/ioda/file-formats.rst#missing-file-action) for the `missing file` action. It says `warn` assigns empty ob space, but I don't think an output file is created, maybe we should check that. I would still like to discuss this with @rtodling, @elakkraoui, @gmao-yzhu, @gmao-wgu as well but with this PR, the strategy seems to be:

1) Get_obs: Check if obs exist, if not pull an empty IODA file from R2D2
2) Run_jedi_executable (this is now handled in RenderJediObs after Michael's PR): Add the observer YAML to the full obs.yaml no matter what (`missing files` flag doesn't impact anything since now we have an empty obs file)
3) Save_obs: only relevant if output obs exist and dimension not 0 
4) EVA obs task (not sure if you touched on this): only relevant if output obs exist and dimension not 0 

For me this is fine, with two caveats
- For NWP or NRT app.s, we miss the Cylc's retry option for getObservations. Frankly I'm not sure how common this is but an obs might be "late" in the data intake pipeline and could still make it in the second or third iteration
- For FP or Reanalysis, @elakkraoui mentioned a `strict check` and `non-strict check` option. We could either implement this in SWELL, or we could expand upon `missing file action` in JEDI/IODA.

We don't need to address these in this PR, I'm just sharing my thoughts.

> Looking closer at the error, it looks like combining failed because out of 5 files for `sst_abi_g16_l3c.*.nc4` only one has data and the rest are empty files. Is this a common case? I can change the empty file to include all the groups in the ioda format and my guess is that it will solve the issue.

Yeah the combining part is a bit complicated and I encountered some edge cases during my long term testing as well.

The reason I have this combination part is that we might have 1-day, 5-day (current S2S) or even longer (1 month for seasonal to decadal prediction) DA windows in marine/landice context. However files are strictly organized in 6-hourly chunks in R2D2 (or any other database) for consistency.

I don't know how common it is in other domains but for marine, some files (like L3 SST you found out about) especially ARGO profilers within this 1 to 5 day window some instruments might contain temperature, salinity, some might have only temperature. And some observations might be missing all together. So for a 1 day window it could look something like this:

To be combined, 5 obs:

Obs 1) T
Obs 2) T, S
Obs 3) T
Obs 4) missing
Obs 5) T,S

I tried addressing this in my restructuring geos_gcm PR but an easier option might be to just separate ARGO profiles into temperature and salinity.

Again, we don't need to address these in this PR, I'm just sharing my thoughts and why are they relevant. However some of your changes need to merge with Michael's latest PR...

jeromebarre:
> The reason I have this combination part is that we might have 1-day, 5-day (current S2S) or even longer (1 month for seasonal to decadal prediction) DA windows in marine/landice context. However files are strictly organized in 6-hourly chunks in R2D2 (or any other database) for consistency.

Ideally this should be a feature that is part of R2D2 but since we are using netcdf aggregating of sectioning the dataset is challenging. I am seeing the appeal for atmospheric DA, of an intermediate functionality where obs could be strored in hourly chunks in R2D2 then the necessary files are retrieved and stitched together. This is kind of off topic for this PR but deserves further discussion to polish this functionality if we want to use it more.

The PR is good to go on my end.

jeromebarre (approved):

Dooruk:
This good to go on my end too, my comments were more broad discussion points. I will approve once it is merged with develop.

mranst:
I believe the only conflict with develop is that `jedi_dictionary_iterator` was deleted, which shouldn't affect this PR. Once that is deleted it should be good to go

Dooruk (approved):

PR #673, opened 2025-12-11
Entries needed when using IODA that comes from bfr converter

rtodling:
## Description

I found that 'error parameter vector' needs to be present in BoundIR check when IODA input file comes from bufr converter.  Something in UFO is not quite correct and the lack of this entry causes BoundIR to reject all obs. 

When IODA files come from GSI (ncdiag) the changes here are zero-diff.

## Dependencies

None

## Impact

None


src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/iasi_metop-c.yaml
gmao-wgu, line 1: It is interesting because we already have this filter in place to ensure that ObsError is assigned to hold the initial observation errors: 
```
# assign ObsError <---- ObsErrorData (Assigned in Yaml)
- filter: Variable Assignment
  assignments:
  - name: ObsError/brightnessTemperature
    channels: *iasi_metop-b_channels
    type: float
    function:
      name: ObsFunction/Arithmetic
      options:
        variables:
        - name: ObsErrorData/brightnessTemperature
          channels: *iasi_metop-b_channels
```

rtodling: Well, I didn't add a filter - I just added the obs error assignment ... and yes, in the default case this doesn't do anything, but when the ioda file comes from bufr this is not how it works - the problem is I believe in how UFO is not properly handling DerivedObsError; instead of spending a more time than I already have trying to figure out what the problem in UFO is, the changes here address the problem more simply.

gmao-wgu: If you review the code, you can see that the line you added in the YAML is assigned to the ObsError array rather than the DerivedObsError array. However, ObsError has already been assigned in the YAML from ObsErrorData through the filter I showed in the previous comment. I have not figured out yet why this assignment filter is not working here when using bufr converted IODA obs file.

gmao-yzhu:
As Wei said, the variable coming from converter is radiance data, It seems the obs error for brightness temperature is somehow applied to radiances. To test if this is the problem, an ioda file could be generated by bufr-query with the radiance to Tb conversion.

Dooruk:
@gmao-yzhu and @gmao-wgu I tested this branch with the develop version and the OOPS results are identical between the two. Do you think it is ok if we merge this?

gmao-wgu:
> @gmao-yzhu and @gmao-wgu I tested this branch with the develop version and the OOPS results are identical between the two. Do you think it is ok if we merge this?

Adding this will introduce an extra 3 × 616 lines for IASI in the YAML file when we run the system, which is my only concern. In the future, if the conversion from radiance to brightness temperature is implemented in bufr-query, we can revert this change.

gmao-wgu (approved):
This looks good to me in the current situation.

Dooruk (approved):

Dooruk:
> 
> Adding this will introduce an extra 3 × 616 lines for IASI in the YAML file when we run the system, which is my only concern. In the future, if the conversion from radiance to brightness temperature is implemented in bufr-query, we can revert this change.

Understood, thanks @gmao-wgu

PR #675, opened 2025-12-11
Implement R2D2 Ingest Suite

### LINKED ISSUE #646: Implement R2D2 Ingestion Suite
Create an ingestion suite that stores observations, bias corrections, and backgrounds to R2D2 v3.

ftgoktas:
## Description

This PR introduces a new observation ingestion suite for Swell, enabling automated ingestion of observation and background data into R2D2 v3 similar to Skylab/Ewok architecture.

**Key Features:**
- **Modular YAML configuration**: Each observation type (e.g., `adt_cryosat2n`, `adt_sentinel6a`) has its own standalone YAML file with retrieval method and metadata.
- **R2D2 duplicate detection**: Automatically checks if observations already exist in R2D2 before ingestion to avoid unnecessary copies.
- **Dry-run mode**: Test ingestion workflows without actually storing files to R2D2.

**Usage:**
`swell create ingest_obs_marine`
`swell launch <experiment_path>` 

The suite automatically ingests all observations listed in `obs_to_ingest` for each cycle point, skipping files already present in R2D2.


Dooruk:
Thanks @ftgoktas, two suggestions right of the bat:

1) I would drop anything `ingest_backgrounds` related. We won't use that approach any time soon so let's focus on the obs for now.
2) i know it's a draft but retrieval method could be used in a "factory pattern" way. I used that for this task here:
https://github.com/GEOS-ESM/swell/blob/5ed0d26d99c5fd2a97c78aa31d0b9f5774fe0cb8/src/swell/tasks/generate_b_climatology.py#L50-L61

That way we can have as many retrieval method we want and simply add it as a separate method/function, obs_proc then can then direct between method A,B or C. It's a super simple concept while I was looking into it I watched [this video](https://www.youtube.com/watch?v=s_4ZrtQs8Do&t=38s) which made it crystal clear for me. 

I do like the one-off scripts you create for R2D2 under utilities. I'm wondering if there is a better way to organize them under somewhere else like `extras` `useful scripts` so other users are also aware? Looking for suggestions from the others here..

metdyn:
Hi,
Thanks for the PR. Apology that I am not running the script today, and I
want a stand alone function such as

Swell R2D2getobs

Yaml:  amsua-n19 ,
            Date1
            Date2
            (Interval)
             Actually I need at least 2020-12-14 21Z.

So that I get the IODA file for this obs.  Thanks.

Dooruk:
> Hi, Thanks for the PR. Apology that I am not running the script today, and I want a stand alone function such as Swell R2D2getobs Yaml: amsua-n19 , Date1 Date2 (Interval) Actually I need at least 2020-12-14 21Z. So that I get the IODA file for this obs. Thanks.

Yonggang, that is the end goal here. Your suggestion assumes that the conversion (BUFR -> IODA) happened offline outside of Swell context and stored somewhere on Discover. See this line for instance, I organized some observations under here:

https://github.com/GEOS-ESM/swell/pull/675/changes#diff-6fdfd3deb1e3f3932856bf26bdf31d0e9cb54263c134c0a0267db878183206aaR8

So `cp` method assumes files are locally ready, similar to what you are suggesting, then Swell will cycle one by one and `ingest` those files to R2D2. **We have to be mindful of the limited disk space and that this is still development stage**. Going forward most obs files will-be read-only, except the experimental ones. 

Practically, we are doing a very lightweight approach similar to here:

https://github.com/JCSDA-internal/skylab/tree/develop/obs/ingest

Except we don't need plotting and we are not doing conversion (netcdf -> IODA) online, at least with this PR.

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+
+from swell.tasks.base.task_base import taskBase
+from swell.utilities.r2d2 import create_r2d2_config
+from swell.swell_path import get_swell_path
+import r2d2
+
+
+class IngestBackground(taskBase):
+    """
+    Task to ingest backgrounds/forecasts to R2D2 v3 using modular YAML configuration files.
```
ashiklom: Is the YAML configuration here an existing practice (e.g., this is how R2D2 / JEDI already works) or did we create something new?

If the latter, I'm going to strongly suggest we create these as Python objects instead, for reasons that I've articulated many times (more flexibility, more robustness), at least for configurations that we store inside of Swell itself. YAML is a decent data format for _users_ of Swell to interact with. But for internal storage, I think it makes more sense to use Python objects (e.g., dataclasses, or maybe Pydantic data models).

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+    
+    - 'backgrounds_to_ingest' list comes from experiment.yaml (e.g., ['geos_restart', 'mom6_forecast'])
+    - Separate YAML file for each background type defining retrieval method and metadata.
```
ashiklom: Echoing comments I made on Michael's big Python workflow PR, I'm going to suggest we start being a bit more rigorous about our docstrings.

Specifically, I'm going to suggest we start using [Google-style Python docstrings](https://google.github.io/styleguide/pyguide.html#s3.8-comments-and-docstrings) and try to make these a bit more comprehensive, with all arguments documented and 1-2 usage examples.

If you're lazy, in my experience, AI is pretty good at drafting docstrings from source code.

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+        self.logger.info("INGESTION SUMMARY")
+        self.logger.info("="*60)
+        if dry_run:
+            self.logger.info(f"Would ingest: {total_ingested} files")
+            self.logger.info(f"Would fail: {total_failed} files")
+        else:
+            self.logger.info(f"Successfully ingested: {total_ingested} files")
+            self.logger.info(f"Failed: {total_failed} files")
+
+    def process_background_config(self, config, bg_name, date, dry_run):
```
ashiklom: Add type hints here for arguments and output.

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+        resolution = bg_metadata.get('resolution')
+        file_type = bg_metadata.get('file_type')
+        step = bg_metadata.get('step', 'P0H')  # Default to analysis time
+        
+        retrieval_method = config.get('retrieval_method')
+        source_pattern = config.get(f'{retrieval_method}_source')
+        
+        if not source_pattern:
+            self.logger.error(f"No source pattern found for method '{retrieval_method}' in {bg_name}.yaml")
+            return ingested, [(bg_name, "Missing source pattern")]
```
ashiklom: Should this be a task-ending error in most cases? If so, we should raise an informative exception here. If not -- i.e., if it's OK that source patterns are missing sometimes --- this should probably be changed to a `logger.warn`.

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+        
+        # Handle wildcards if present (glob)
+        if '*' in expected_file:
+            files_found = glob.glob(expected_file)
+            if not files_found:
+                self.logger.warning(f"No files matched pattern: {expected_file}")
+                return ingested, [(bg_name, "No files found")]
+            target_file = files_found[0]
+        else:
+            target_file = expected_file
```
ashiklom: A glob is a glob, whether or not it contains a wildcard. Given a directory like:

```
mydir/
  file1
  file2
```

...globs work like this:

```
In [1]: import glob

In [2]: list(glob.glob("mydir/"))
Out[2]: ['mydir/']

# NOTE: Incomplete matches are not matched
In [4]: list(glob.glob("myd"))
Out[4]: []

In [5]: list(glob.glob("mydir/file"))
Out[5]: []

# You *need* to have wildcards to match multiple things
In [6]: list(glob.glob("mydir/file*"))
Out[6]: ['mydir/file2', 'mydir/file1']

# There are other wildcards besides `*`
In [7]: list(glob.glob("mydir/file?"))
Out[7]: ['mydir/file2', 'mydir/file1']
```

If we accept globs in these patterns, then we don't need to condition on the presence of wildcards (which is unreliable -- case in point, here, you miss at least one other kind of wildcard, the `?` single-character wildcard) --- you can just run everything through `glob.glob` and, in the absence of wildcards, it will only return exact matches. That lets us simplify the logic here by removing the conditional.

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+                    resolution=resolution,
+                    step=step,
+                    date=date,
+                    file_type=file_type,
+                    source_file=target_file
+                )
+
+                ingested.append(target_file)
+                self.logger.info(f"Successfully ingested {bg_name}")
+            except Exception as e:
```
ashiklom: In general, you should not catch generic `Exception` unless we immediately re-raise it (see [Google Python style guide](https://google.github.io/styleguide/pyguide.html#24-exceptions)). This will also catch exceptions like user interrupts (`Ctrl-C`). If there are common exceptions in `r2d2.store` that we can safely ignore through this mechanism, we should except those specific exception classes explicitly:

```python
try:
  # whatever
except (ValueError, TypeError, KeyError, NetworkError) as e:
  # handler...
# All *other* exception classes will be raised immediately.
```

src/swell/tasks/ingest_background.py
ashiklom: In particular, the output type hints for functions like this are very useful. It's helpful for users to know that this always returns two lists -- a list of ingested files (`list[str]`?) and a list of failures (`list[tuple[str, str]]`?).

So, the final type hint would be: `tuple[list[str], list[tuple[str, str]]]`.

src/swell/tasks/ingest_obs.py
```diff
@@ -0,0 +1,200 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
```
ashiklom, line 1: A lot of my comments from `ingest_background` apply here as well --- type hints, docstrings, glob handling, catching generic `Exception`, etc.

ashiklom (changes requested):
See my comments. TLDR:
* Glob logic can be simplified.
* Add type hints
* Don't catch generic `Exception` if we don't immediately re-raise it.

src/swell/tasks/ingest_background.py
ftgoktas: This YAML pattern is borrowed from the existing [Ewok/Skylab's ingest](https://github.com/JCSDA-internal/skylab/blob/develop/obs/ingest/g17_abi_ingest.yaml) approach for consistency. Each observation type has a standalone YAML defining retrieval method and source paths. I agree Python objects would be more robust. These YAMLs are very simple (just retrieval_method + source_path), but we could wrap them in a validation dataclass.

ashiklom: If we're following Skylab's ingest structure, then let's keep these as YAML! The most important thing is to be consistent --- having them use YAML and us use Python objects would be confusing.

Down the line, might be nice to have some validation, but not necessary for this PR.

Dooruk: A little correction, we are just getting "inspired" by their structure so YAML is not a strict requirement.

mer-a-o:
Thanks @ftgoktas for adding this feature. Some general comments before getting into details:

- I don't understand the point of `dry_run`. Is it more like a search functionality? If yes, can we change the name to something else as `dry_run` in the context of cycling means something else. 
- I suggest adding observation ingest in a separate PR. How would you add running ioda-converter step to this task?
- Please add a clear description about each of the options in the yaml.

ftgoktas:
> Thanks @ftgoktas for adding this feature. Some general comments before getting into details:
> 
> * I don't understand the point of `dry_run`. Is it more like a search functionality? If yes, can we change the name to something else as `dry_run` in the context of cycling means something else.
> * I suggest adding observation ingest in a separate PR. How would you add running ioda-converter step to this task?
> * Please add a clear description about each of the options in the yaml.

1) `dry_run` mode validates files exist and logs what would be ingested, but skips `r2d2.store()`. It's for testing ingestion logic before committing data to R2D2. 
It's not a search functionality, it still runs the full task logic (file pattern matching, existence checks, metadata lookup), just without the final storage step. We can rename to preview_mode or test_ingest if dry_run is confusing in the Cylc context.


2) In Ewok/Skylab's pattern, ingestion is a separate task (`storeObservations`) that runs after conversion (`convertObservations`). Swell already has conversion tasks (`BufrToIoda`), so `IngestObs` will complete the workflow by handling the R2D2 storage step.

ashiklom:
> I don't understand the point of `dry_run`. Is it more like a search functionality? If yes, can we change the name to something else as `dry_run` in the context of cycling means something else.

FWIW, `dry-run` is a pretty common flag in software engineering with a widely understood definition of "print what the command _would_ have done but don't actually do it". E.g., `rsync`, `aws s3 sync`, and lots of other commands that do big/many file modifications have a `--dry-run` option to first check that the files are going to end up where you expect them to before you actually move/copy/delete anything. So my vote would be to keep this name!

In the context of rsync: Can you predict exactly what each of these commands will do? They do different things, but are you 100% sure you know the result?

```
rsync -av path/to/source path/to/destination
rsync -av path/to/source path/to/destination/
rsync -av path/to/source/ path/to/destination
rsync -av path/to/source/ path/to/destination/
```

It's very easy to mess up a directory or even to delete files if you accidentally use the wrong combination of trailing slashes. `rsync --dry-run` is handy because it will print each individual copy operation it would do (and raise any errors/warnings it might encounter) but will not actually do the operations, so you can quickly confirm that you're doing what you intend.

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
```diff
@@ -0,0 +1,15 @@
+# Data availability period (for reference only):
+# available_period:
+#   time_begin: '2020-01-01T00:00:00Z'
+#   time_end: '2021-12-31T23:59:59Z'
+
+bg_to_ingest:
+  item_type: 'background'
```
mer-a-o: `item_type` is not an R2D2 key and `file_type` alone can describe the file. Is there a reason to have `item_type` and separate `background` vs `forecast`?

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
```diff
@@ -0,0 +1,15 @@
...
+  model: 'geos'
+  experiment: 'c180_o1440'
+  resolution: 'C180'
```
mer-a-o: In general, I suggest being consistent with `C180` and `c180`. In this case for experiment name probably doesn't matter. In the jcsda data_store lower case is used e.g. `c180`.

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+        ingested = []
+        failed = []
+        
+        # Extract metadata
+        bg_metadata = config.get('bg_to_ingest', {})
+        model = bg_metadata.get('model')
+        experiment = bg_metadata.get('experiment')
+        resolution = bg_metadata.get('resolution')
+        file_type = bg_metadata.get('file_type')
+        step = bg_metadata.get('step', 'P0H')  # Default to analysis time
```
mer-a-o: `P0H` should be `PT0H`?
I think we need a more detailed discussion about using`PT0H` as the default value, especially for the restart files. In Skylab/geos-cf we keep the information about the lead time of restart files when we are storing them on r2d2.

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
```diff
@@ -0,0 +1,15 @@
...
+  file_type: 'restart'
```
mer-a-o: `step` for restart files? See my comment in src/swell/tasks/ingest_background.py

src/swell/configuration/jedi/interfaces/geos_marine/ingest_observations/adt_cryosat2n.yaml
```diff
@@ -0,0 +1,6 @@
+# Data availability period (for reference only):
+# time_begin: '2023-07-01T00:00:00Z'
+# time_end: '2999-12-31T23:59:59Z'
+retrieval_method: 'cp'
+# cp_source: '/discover/nobackup/projects/gmao/soca/obs/ioda/ocean/adt_cryosat2n/YYYY/MM/'
```
mer-a-o: remove this line

src/swell/configuration/jedi/interfaces/geos_marine/ingest_observations/adt_sentinel6a.yaml
```diff
@@ -0,0 +1,8 @@
+# Data availability period (for reference only):
+# available_period:
+# time_begin: '2023-07-01T00:00:00Z'
+# time_end: '2999-12-31T23:59:59Z'
+
+retrieval_method: 'cp'
+# cp_source: '/discover/nobackup/projects/gmao/soca/obs/'
```
mer-a-o: remove this line

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+retrieval_method: 'cp'
+cp_source: '/discover/nobackup/projects/gmao/soca/data/GEOS_restarts/YYYY/MM/DD/geos.bkg.YYYYMMDDHH.nc4'
```
mer-a-o: I can't view `/discover/nobackup/projects/gmao/soca/data` directory. Is this a permission issue?

src/swell/tasks/ingest_obs.py
```diff
@@ -0,0 +1,200 @@
...
+        """Check if observation already exists in R2D2."""
+        try:
+            # Query R2D2 to see if this observation already exists
+            results = r2d2.fetch(
+                item='observation',
+                provider=provider,
+                observation_type=obs_name,
+                window_start=window_start,
+                window_length=window_length,
+                type='fetch'  # Just check, don't download
```
mer-a-o: Is there a `type` key for fetching observations in r2d2? Why not use r2d2.search instead of r2d2.fetch?

src/swell/tasks/ingest_background.py
```diff
@@ -0,0 +1,169 @@
...
+            self.logger.info(f"    Model: {model}")
+            self.logger.info(f"    Experiment: {experiment}")
+            self.logger.info(f"    Resolution: {resolution}")
+            self.logger.info(f"    Method: {retrieval_method}")
+            self.logger.info(f"    Source: {target_file}")
+            ingested.append(target_file)
+        else:
+            try:
+                # Store to R2D2
+                r2d2.store(
```
mer-a-o: Do we want to search to see if the files are already stored on R2D2?

src/swell/utilities/scripts/search_ingested.py
```diff
@@ -0,0 +1,31 @@
+import r2d2
+from datetime import datetime
+
+# providers = ['gdas']
+# Search for all observations
+results = r2d2.search(
```
mer-a-o, line 5: Can you make this functionality more generic so it can be used in ingest_background and ingest_obs instead of `check_already_in_r2d2`?

mer-a-o:
Thanks @ftgoktas and @ashiklom for the clarificaiton. `dry-run` makes sense now.

mer-a-o:
> 2\. In Ewok/Skylab's pattern, ingestion is a separate task (`storeObservations`) that runs after conversion (`convertObservations`). Swell already has conversion tasks (`BufrToIoda`), so `IngestObs` will complete the workflow by handling the R2D2 storage step.

@ftgoktas, In skylab, the observation ingest suite includes downloading (or copying) observations, running the ioda converter which is unique for each observations type (we don't always use `BufrToIoda` converter), and then storing to r2d2. In Swell, I don't see the conversion being part of the `ingest_obs/suite_config.py` right now. 
 
Having the conversion task as part of the observation ingest suite would streamline the whole process. We can skip conversion if the files are already in ioda format. I assume you are going to add downloading the files from S3 buckets of ftp later. That would be a good place to add the conversion step to the suite.

src/swell/tasks/ingest_background.py
ftgoktas: Thank you, made it more explicit

src/swell/tasks/ingest_background.py
ftgoktas: Added doctrines for ingest_obs suite, will be removing ingest_background for now as it will be a separate PR.

src/swell/tasks/ingest_background.py
ftgoktas: Correct, fixed it.

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
ftgoktas: Thanks, these yamls were outdated and background suite will probably be a separate PR. Update it and now including R2D2 metadata (model, experiment, resolution, file_type) and retrieval info.

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
ftgoktas: Could you elaborate? The code defaults to `PT0H` when not specified in the YAML.

src/swell/configuration/jedi/interfaces/geos_marine/ingest_backgrounds/geos_restart.yaml
Dooruk: No, there is just no such folder for backgrounds.

src/swell/tasks/ingest_obs.py
ftgoktas: Good catch - there's no type key for fetch; changed to `r2d2.search()`. The search call adds an API round-trip, but it avoids re-ingesting files that R2D2 would just overwrite anyway. We can remove this check if the overhead isn't worth it.

src/swell/tasks/ingest_background.py
ftgoktas: Added `r2d2.search()` check to ingest_obs.py. For backgrounds, I added the same check - it's one extra API call per file but avoids redundant file copies when re-running. It's worth it for large batch ingestions; can be removed if overhead is a concern.

mer-a-o:
@ftgoktas I tested your branch and was able to ingest the obs you have in this PR in "dry-run" mode. 
I tried adding an observation that I know exists on R2D2 in dry-run=False and got this failure:

```
TaskFactory: Could not find model-specific version of IngestObs. Looking for generic version.
TaskFactory: Using module swell.tasks.ingest_obs
IngestObs:   Initializing task with the following parameters:
IngestObs:   Task name:     IngestObs
IngestObs:   Configuration: /home/mabdiosk/cylc-run/swell-ingest_obs-suite/run5/experiment.yaml
IngestObs:   Date and time: 20230808T1200Z

IngestObs: Using R2D2 configuration for platform                     'nccs_discover_sles15': host=discover-gmao,                     compiler=intel
IngestObs: Loading R2D2 v3 credentials from /home/mabdiosk/.swell/r2d2_credentials.yaml
IngestObs: R2D2 v3 credentials loaded successfully
IngestObs: Preparing to ingest: tempo_no2_tropo
IngestObs: Successfully ingested tempo_no2_tropo
IngestObs: ============================================================
IngestObs: INGESTION SUMMARY
IngestObs: ============================================================
IngestObs: Successfully ingested: 1 files
Traceback (most recent call last):
  File "/gpfsm/dnb33/mabdiosk/SWELL_uv/swellvenv/bin/swell", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/gpfsm/dnb33/mabdiosk/SWELL_uv/swell/src/swell/swell.py", line 297, in main
    swell_driver()
  File "/gpfsm/dswdev/jcsda/spack-stack/scu17/spack-stack-1.9.0/envs/ue-intel-2021.10.0/install/intel/2021.10.0/py-click-8.1.7-jv64c2h/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dswdev/jcsda/spack-stack/scu17/spack-stack-1.9.0/envs/ue-intel-2021.10.0/install/intel/2021.10.0/py-click-8.1.7-jv64c2h/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/gpfsm/dswdev/jcsda/spack-stack/scu17/spack-stack-1.9.0/envs/ue-intel-2021.10.0/install/intel/2021.10.0/py-click-8.1.7-jv64c2h/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dswdev/jcsda/spack-stack/scu17/spack-stack-1.9.0/envs/ue-intel-2021.10.0/install/intel/2021.10.0/py-click-8.1.7-jv64c2h/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dswdev/jcsda/spack-stack/scu17/spack-stack-1.9.0/envs/ue-intel-2021.10.0/install/intel/2021.10.0/py-click-8.1.7-jv64c2h/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dnb33/mabdiosk/SWELL_uv/swell/src/swell/swell.py", line 204, in task
    task_wrapper(task, config, datetime, model, ensemblePacket)
  File "/gpfsm/dnb33/mabdiosk/SWELL_uv/swell/src/swell/tasks/base/task_base.py", line 361, in task_wrapper
    task_object.execute()
  File "/gpfsm/dnb33/mabdiosk/SWELL_uv/swell/src/swell/tasks/ingest_obs.py", line 143, in execute
    self.logger.info(f"Skipped (already exist): {total_skipped} files")
                                                 ^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'total_skipped' where it is not associated with a value
2026-02-03T01:44:14Z CRITICAL - failed/ERR
```
Do you still plan to keep the feature that avoids ingesting the obs if it's already on R2D2?
I also noticed that currently `window_start` isn't set to the beginning of the assimilation window (3Z,9Z, 15Z, 21Z) values, but the middle of the window (0Z, 6Z, 12Z, 18Z) values.

I'm tagging @jeromebarre so he can review the delete functionality you are adding here, as he's working on it too.

src/swell/tasks/ingest_obs.py
```diff
@@ -0,0 +1,223 @@
...
+        else:
+            try:
+                # Store to R2D2
+                r2d2.store(
+                    item='observation',
+                    provider=provider,
+                    observation_type=obs_name,
+                    file_extension=os.path.splitext(
+                        target_file)[1][1:],  # 'nc' from '.nc'
+                    window_start=window_start,
```
mer-a-o, line 210: `window_start` needs to be set to the beggining of the assimilation windows. Right now, I believe, is set to the middle of the window.

ftgoktas: Fixed, the `window_start` now uses the beginning of the DA window (03Z, 09Z, etc.).

I removed the duplicate check for now. Adding a `search()` call before every `store()` doubles the API calls and if someone reruns ingestion for the same window, they likely want to refresh the data anyway. 

If that would be still useful, I can do it in the next PR. We can have the `check_if_already_in_r2d2` function back to check first and skip if already exists but it will add some latency.

Screenshot 2026-02-03 at 4 06 41 PM

mer-a-o (dismissed):
Great work! Thanks @ftgoktas.

jeromebarre:
Nice first PR for an ingest suite. Thank you for that.
An ingest suite than can be used routinely does 3 main things for a given date or assimilation window span: 
1. raw/original data aquisition retrieval, whether it is bufr on a server or in the local HPC archive or some granules files in a remote server , etc
2. ioda conversion: can be bufr to ioda or netcdf/hdf5 to ioda or else 
3. storage and registration in r2d2: what this PR does.

Do we have plans to implement 1 and 2 in the future?

ftgoktas:
> Nice first PR for an ingest suite. Thank you for that. An ingest suite than can be used routinely does 3 main things for a given date or assimilation window span:
> 
> 1. raw/original data aquisition retrieval, whether it is bufr on a server or in the local HPC archive or some granules files in a remote server , etc
> 2. ioda conversion: can be bufr to ioda or netcdf/hdf5 to ioda or else
> 3. storage and registration in r2d2: what this PR does.
> 
> Do we have plans to implement 1 and 2 in the future?

Yes, the plan is to support 1 and 2 in a separate PR. Swell already has conversion tasks (BufrToIoda), so the idea would be to wire them into the ingest suite with the option to skip conversion when files are already in IODA format. I’ll track that as follow-up work once this PR is merged.

src/swell/tasks/ingest_obs.py
```diff
@@ -0,0 +1,224 @@
...
+        total_failed = 0
+
+        # Iterate over the simple list of names
+        for obs_name in obs_to_ingest:
+            self.logger.info(f"Preparing to ingest: {obs_name}")
+
+            # Locate the configuration file
+            # Look in the JEDI config directory and get obs yaml file from
+            # obs_name (e.g. adt_cryosat2n.yaml)
+            config_path = os.path.join(
```
mranst, line 110: Are the origin filepaths for ingests expected to be static, or will users want to set their own locations? If so, I would personally suggest these not being hardwired into the source code. If users want to make a change, those using static installations will need to run a re-install of swell for it take effect. I think the easiest solution to this would be to keep the file structure, but instead of directly referencing the source code, have swell copy these files to the experiment directory, and reference them from there. We handle Eva configs and (some parts of, as of #605) JEDI files this way.

https://github.com/GEOS-ESM/swell/blob/953b02879e9206fdd70c88fb16f20faaf6dbbeb9/src/swell/deployment/create_experiment.py#L278-L294

This will allow users to modify these files in their individual experiments before launching, without having to change the source code (though this will still work too).

Just a thought, I don't necessarily want to hold it up over this

mranst (commented):
Great work! It'll be really nice to have this as a feature in swell. I have one suggestion regarding the file configuration but from my perspective it looks good

src/swell/configuration/jedi/interfaces/geos_marine/ingest_observations/adt_cryosat2n.yaml
Dooruk, line 1: We don't need to address this in this PR but the general trend in SWELL is to move away from YAML for internal config. So going forward we could change this to pydataclass?

https://github.com/GEOS-ESM/swell/issues/314#issuecomment-2198857656

src/swell/suites/ingest_obs/README.md
Dooruk, line 1: This is in the `\docs` part now so can be erased.

src/swell/tasks/ingest_obs.py
```diff
@@ -0,0 +1,224 @@
...
+        window_start = self.da_window_params.window_begin_iso(window_length)
+
+        # Used for file path construction
+        cycle_time = self.cycle_time()
+
+        # Check for dry-run mode (default True for safety)
+        dry_run = self.config.dry_run(True)
+
+        if dry_run:
+            self.logger.info("=" * 60)
```
Dooruk: Am I the only annoying guy that gets bothered by these `=` lines 🤔

Dooruk (changes requested):
This is great first step for ingesting obs in SWELL!

The following is kind of what we wanted to replicate in terms of the design, but not with YAML, this is what Jerome and Maryam are familiar too:
https://github.com/JCSDA-internal/skylab/tree/develop/obs/ingest

In this sense we can activate different retrieval methods, iodaconverters, and perhaps utilize some external (e.g., another GEOS-ESM) repo obs processing scripts if need be.

src/swell/tasks/ingest_obs.py
ftgoktas: Just removed

src/swell/tasks/ingest_obs.py
ftgoktas: Great point, I just changed this. Now users can modify the observation YAMLs in their experiment directory without touching the source code.

mer-a-o (approved):

Dooruk (approved):

mranst (approved):

PR #677, opened 2025-12-15
Execute GEOS via calling gcm_run.j + rename marine suites -- part 1, 2, & 3

Dooruk:
This is ready to go in. CI-workflows need to be modified to handle new suite names like so:

https://github.com/GEOS-ESM/CI-workflows/pull/32

## Main change 1:

SOCA (marine) suites are renamed:

3dvar -> 3dvar_marine
3dvar_cycle -> 3dvar_marine_cycle
3dfgat_cycle -> 3dfgat_marine_cycle

## Main change 2:

Cylc calls `gcm_run.j` directly in `flow.cylc`.

With this new approach, SWELL can point to an existing GEOS experiment folder (the `experiment.yaml` key for that is `geos_homdir`) and the `forecast` folder is now located under experiment `GEOSgcm/forecast` directory. It is possible to hotstart. With this new approach, forecast directory is not erased and MAPL history outputs can be accumulated under there. I updated docs a bit but might add on more for GEOSgcm execution.

### For those who stumbled upon this PR, more details on change 2 below:

The main thing happening here is that Cylc (flow.cylc) now calls `gcm_run.j` directly. To facilitate this a `forecast` directory was created under `{swell_exp_dir}/GEOSgcm/forecast`. This `forecast` folder is a replication of a GEOS experiment folder, with only a few changes regarding where HOMDIR, EXPDIR are defined. Model execution happens under `{swell_exp_dir}/GEOSgcm/forecast/scratch` similar to typical GEOS model runs.

Why was this change necessary:

- We want SWELL to be less involved in GEOS model execution task(s). The previous method required lots of file manipulation (in particular due to the boundary condition files, AKA `/RC` files) in the forecast directory. This creates incompatibility while running/testing different GEOSgcm versions. Between multiple products and update frequencies, this is an important requirement.
- In Cylc templating  `forecast` dir can't be updated in `flow.cylc` if it is templated in a time dependent way.
- `subprocess` simply couldn't run GEOSv12 on Milan nodes. I tried many combinations, it didn't pass beyond the initialization stage.
- Defining sufficient nodes, MPI layouts etc. is handled by `gcm_run.j`. If users make mistake in terms of requesting sufficient SLURM nodes, GEOS tries submitting hundreds of instances to compensate lack of compute resources, then NCCS will yell at you. 
- (long term relevancy) `gcm_run.j` and `gcm_setup.j` scripts are being or will be modernized. This is work underway but might take a long time (especially `gcm_run.j`).

⚠️ Which means to use a `gcm_run.j` in SWELL, some parts should be erased or commented out. Or, my idea is that there could be conditional sections in `gcm_run.j` say `SWELL_active`, then `gcm_run.j` can skip those sections, which are mainly `postprocessing` anyway.

More details in below comment: https://github.com/GEOS-ESM/swell/pull/677#issuecomment-3855447115

## Finally, little primer on `gcm_run.j`
Let's consider `gcm_run.j` in 4 stages:

1) SLURM & node assignment
2) Preprocessing
3) Execution
4) Postprocessing

In the current implementation, SWELL handles 2 & 3 via python and `subprocess` and 1 is assumed to be set properly by the user, which caused trouble with the NCCS. For DA purposes 4, postprocessing is explicitly handled by SWELL but that is not the focus of this PR.

In this proposed implementation, the main difference is that we rely on `gcm_run.j` for 2 and 3 by conducting surgical edits via `PrepCoupledGeosRundir` at few locations and running `gcm_run.j` directly from Cylc (which doesn't capture failed exit status):

```cylc
    [[RunGeos]]
        script = "{{experiment_path}}/forecast/gcm_run.j"
        platform = {{platform}}
        [[[job]]]
            shell = /bin/csh
        [[[directives]]]
        {%- for key, value in scheduling["RunGeos"]["directives"]["all"].items() %}
            --{{key}} = {{value}}
        {%- endfor %}
```

I created the `3dfgat_coupled_cycle` suite for testing, should work by default if anyone has time to check it out.


src/swell/tasks/base/task_base.py
```diff
@@ -202,25 +203,23 @@ def cycle_dir(self) -> str:
 
     def forecast_dir(self, paths: Union[str, list[str]] = []) -> Optional[str]:
 
+        '''
+        Method to provide "forecast" directory to geos class
+        If paths are provided, it is combined with the forecast directory and returned
+        '''
```
ashiklom, line 209: Better! But, FWIW, where I'd like to get with these is something like this:


```suggestion
        """Set structure of 'forecast' directory for GEOS

        Args:
            paths: One or more paths to concatenate into a nested forecast directory structure.

        Returns:
            The full forecast directory path, appended to `self.cycle_forecast_dir`, as a single string.

        Examples:
            For example, if `obj.cycle_forecast_dir = "/path/to/cycle"` then:

            >>> obj.forecast_dir("forecast")
            "/path/to/cycle/forecast"

            >>> obj.forecast_dir(["nested", "dir", "structure"])
            "/path/to/cycle/nested/dir/structure"
        """
```

ashiklom: For more examples of Google-style docstrings, see here:

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Also, you don't have to change it here (because we use `forecast_dir` a lot, and this PR touches a lot of things!), but just flagging that it's best to avoid functions that can take multiple types whenever possible. In this case, this should either _always_ take a string and literally concatenate it, or (perhaps, better) it should _always_ take a list of strings (even if there's just one path). That requires much less code with, IMHO, much clearer and more consistent behavior.

```python
def forecast_dir(self, paths: list[str]) -> str:
  os.makedirs(self.cycle_forecast_dir, 0o755, exist_ok=true)
  return os.path.join(self.cycle_forecast_dir, *paths)
```

src/swell/tasks/generate_b_climatology.py
```diff
@@ -18,6 +18,81 @@
...
+        # ----------
+        self.jedi_rendering.add_key('local_background_time', local_background_time)
+        self.jedi_rendering.add_key('local_background_time_iso', local_background_time_iso)
+
+        model_component_meta = self.jedi_rendering.render_interface_meta()
+        self.jedi_interface = model_component_meta['jedi_interface']
+
+        # Compute number of processors
+        # ----------------------------
+        self.np = eval(str(model_component_meta['total_processors']))
```
ashiklom: I know we use this pattern throughout Swell, but in general, you should avoid `eval`. For one, it's unsafe:

```python
s = "print('uh oh, arbitrary code execution')"
result = eval(str(s))
# uh oh, arbitrary code execution
print(result)
# None
```

But more practically, it doesn't guarantee that the number of processors is an integer, though we know it has to be:

```python
s = "9."
eval(str(s))
# 9.0
```

You can achieve the same thing far more safely _and_ clearly with `int`.

```suggestion
        self.np = int(model_component_meta['total_processors'])
```

src/swell/tasks/get_coupled_geos_restart.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # If expdir is set to be different to homdir, create a link to expdir and fail if it does not exist
+        if expdir_different:
+            self.expdir = self.config.geos_expdir()
+
+        if expdir_different and os.path.exists(self.geos_expdir):
+            geos_expdir_path = os.path.join(swell_exp_path, 'GEOSgcm', 'GEOS_expdir')
+            # Check if link already exists, unlink first if it does
+            if os.path.lexists(geos_expdir_path):
+                os.unlink(geos_expdir_path)
```
ashiklom, line 81: Seems like a good place for an info/warning message about deleting previous experiments.

src/swell/tasks/get_coupled_geos_restart.py
```diff
@@ -0,0 +1,135 @@
...
+            self.logger.info('Linking GEOS EXPDIR to ' + geos_expdir_path)
+            os.symlink(self.expdir, geos_expdir_path)
+        elif expdir_different and not os.path.exists(self.expdir):
+            self.logger.abort('GEOS_expdir directory does not exist: ' + self.expdir)
+
+        # Create forecast_dir and RESTART
+        # ----------------------------
+        if not os.path.exists(self.forecast_dir('RESTART')):
+            os.makedirs(self.forecast_dir('RESTART'), 0o755, exist_ok=True)
```
ashiklom: Unnecessary `if` since `exist_ok` should already do this.

src/swell/tasks/link_coupled_geos_output.py
```diff
@@ -0,0 +1,252 @@
...
+        # Since history already has the aggregated variables, we just need to rename
+        # the dimensions and variables to match SOCA requirements
+        ds = xr.open_dataset(src_history)
+
+        # rename the dimensions to xaxis_1 and yaxis_1 and rename the variables
+        ds = ds.rename({'ni': 'xaxis_1', 'nj': 'yaxis_1'})
+        ds = ds.rename({'aice': 'aice_h', 'hi': 'hi_h', 'hs': 'hs_h'})
+
+        # Save as a new file
+        ds.to_netcdf(dst_history, mode='w')
```
ashiklom: Suggest being explicit about format here.

```suggestion
        ds.to_netcdf(dst_history, mode='w', format='NETCDF4')
```

src/swell/tasks/move_erase_da_restart.py
```diff
@@ -34,49 +36,32 @@ def execute(self) -> None:
...
-    # ----------------------------------------------------------------------------------------------
-
-    def at_next_fcst_dir(self, paths: Union[str, list]) -> str:
-
-        # Ensure what we have is a list (paths should be a list)
-        # ------------------------------------------------------
-        if isinstance(paths, str):
-            paths = [paths]
+        if not os.path.exists(self.forecast_dir('RESTART')):
+            os.makedirs(self.forecast_dir('RESTART'), 0o755, exist_ok=True)
```
ashiklom: With `exist_ok`, can drop the `if`.

src/swell/tasks/link_coupled_geos_output.py
```diff
@@ -0,0 +1,252 @@
...
+    def cice6_history_formatter(self,
+                                src_date: dt,
+                                hour_prefix: str,
+                                ) -> str:
+
+        # Extract the date part and calculate seconds for the src_history format
+        date_str = src_date.strftime('%Y-%m-%d')
+        seconds = src_date.hour * 3600 + src_date.minute * 60 + src_date.second
+
+        return self.forecast_dir(f'scratch/iceh_{hour_prefix}.{date_str}-{seconds:05d}.nc')
```
ashiklom, line 128: General comment: We are inconsistent in our use of `nc` vs. `nc4` extensions for NetCDF4 files throughout Swell. E.g., `move_da_restart.py` and `gsi_ncdiag_to_ioda.py` create `nc4` files, and it seems like most of our obs come in `nc4`.

I suggest we consistently embrace the `.nc4` extension to make it unambiguous that we're using NetCDF4 throughout and to make it easier to craft globs, etc.

src/swell/tasks/prepare_analysis.py
```diff
@@ -94,6 +100,9 @@ def execute(self) -> None:
     # ----------------------------------------------------------------------------------------
 
     def at_cycledir(self, paths: Union[list, str] = []) -> str:
+        '''
+        Get the absolute path to the model cycle directory for the given relative paths.
+        '''
```
ashiklom: Suggest following the [Google structure](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) (more examples [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)).

src/swell/utilities/file_system_operations.py
```diff
@@ -28,8 +28,8 @@ def copy_to_dst_dir(logger: Logger, src: str, dst_dir: str) -> None:
             logger.info(' Copying file: '+src)
             shutil.copy(src, dst_dir)
 
-    except Exception:
-        logger.abort('Copying failed, see if source files exists')
+    except Exception as e:
```
ashiklom: In general, avoid catch-all `except Exception`.

If `shutil` is most likely to throw an `OSError`, we can catch that explicitly and let all other errors bubble up naturally.

```python
except OSError as e:
  logger.abort("...")
```

Better yet, in general, we should start avoiding this pattern altogether and just let exceptions get raised natively by Python and bubble up through the call stack.

src/swell/utilities/file_system_operations.py
```diff
@@ -153,7 +153,7 @@ def move_files(logger: Logger, src_dir: str, dst_dir: str) -> None:
         logger.info(' Moving file(s) from: '+src_dir)
         shutil.move(src_dir, dst_dir)
 
-    except Exception:
-        logger.abort('Moving failed, see if source files exist')
+    except Exception as e:
```
ashiklom: Ditto to above.

src/swell/utilities/question_defaults.py
```diff
@@ -553,12 +553,49 @@ class geos_build_method(TaskQuestion):
...
+
+    @dataclass
+    class geos_expdir_different(TaskQuestion):
+        default_value: str = False
+        question_name: str = "geos_expdir_different"
+        ask_question: bool = True
+        options: List[bool] = mutable_field([
+            True,
+            False
+        ])
```
ashiklom, line 598: What's up with this data structure?

If we always expect exactly 2 things in here, this should be a `tuple`, not a `list`. That said, if these are specific order-dependent options, that's brittle and not very user-friendly because you have to remember which one comes first and which comes second. If these have specific meanings, it's much better to use a `dict` with named keys to make it clear what you're using (or, better yet, a small `dataclass` specific to this set of options).

src/swell/suites/3dfgat_coupled_cycle/suite_config.py
```diff
@@ -0,0 +1,100 @@
...
+                "sea_water_salinity",
+                "sea_water_potential_temperature",
+                "sea_surface_height_above_geoid",
+                "sea_water_cell_thickness",
+                "sea_ice_area_fraction",
+                "sea_ice_thickness",
+                "sea_ice_snow_thickness"
+            ]),
+            qd.window_length("PT6H"),
+            qd.window_offset("PT3H"),
```
mranst: The window_offset causes failures since its been removed from swell

src/swell/utilities/question_defaults.py
mranst: The options are a list of acceptable answers to the question, in this case it can be either True or False. We could change these to tuples, bypassing the need for the mutable field constructor. I kept it as lists when converting the questions since that's the type they would be when read from yaml, as done previously, and we also still populate options from yaml in the `defer_to_model` questions

ashiklom: Ah, got it. In that case, let's leave this as is for now.

src/swell/tasks/link_coupled_geos_output.py
Dooruk: Yeah, this is a model restriction unfortunately. This file is an history output from CICE6, so we don't have control over the output file extension.
Same with MOM6. See below locations:

`/discover/nobackup/projects/gmao/soca/dardag/GEOS_FORWARD/GEOS_v12_rc20/coupled_025_l91_EMC/CICE_Output`
`/discover/nobackup/projects/gmao/soca/dardag/GEOS_FORWARD/GEOS_v12_rc20/coupled_025_l91_EMC/MOM_Output`
`/discover/nobackup/projects/gmao/soca/dardag/GEOS_FORWARD/GEOS_v12_rc20/coupled_025_l91_EMC/RESTART`

I try using MAPL History collections for .nc4 putput consistency but some history outputs and restarts only has `.nc`

src/swell/tasks/link_coupled_geos_output.py
Dooruk: This would write a *.nc4 file, correct?

src/swell/utilities/file_system_operations.py
Dooruk: I understand OSError is a subset of Exception. Are you suggesting also to avoid `as e` part?

src/swell/tasks/prepare_analysis.py
Dooruk: Agreed, we should have a consistent format.

src/swell/suites/3dfgat_coupled_cycle/suite_config.py
Dooruk: Yeah, I had that change in my full PR. I don't think this PR will work without the remainder of the parts.

src/swell/utilities/file_system_operations.py
ashiklom: Yes, `OSError` is a subclass of `Exception`, but that's exactly the point: It's bad engineering practice to be overly indiscriminate in handling exceptions (i.e., `except Exception`). Basically, the problem is that `SystemExit` (triggered by some system error calls), `KeyboardInterrupt` (the exception triggered by pressing `Control-C`), and lots of other things are *all* rolled up under `Exception`. In most cases, you do not want `SystemExit` and `KeyboardInterrupt` to be handled in the same ways as other things (and usually, you don't want these handled at all but raised immediately).

As a specific example: As written right now, if you try to interrupt this code with `Control-C` while it happens to be inside this `try-except` block, you will not interrupt immediately but rather keep executing via the `logger.abort` statement.

Here is a more detailed writeup: https://jerrynsh.com/stop-using-exceptions-like-this-in-python/ 

All that said: `logger.abort` will just re-raise the same exception with a logging message, so the practical consequences of this are relatively minor. But we could simplify a lot of our code by just not bothering with `logger.abort` in the first place and just letting exceptions raise themselves --- Python is perfectly good at printing out backtraces and other useful information when it raises errors without it needing to be wrapped in a `logger`.

src/swell/tasks/link_coupled_geos_output.py
ashiklom: Ah, OK. That's irritating, but it is what it is.

src/swell/tasks/link_coupled_geos_output.py
ashiklom: `format="NETCDF4"` has nothing to do with the file extension. It will write to whatever the full value of `dst_history` is. If `dst_history = "somefile.crazy"`, it will write a file called `somefile.crazy`.

The point here is to (1) make it clear to other/future programmers that we're writing NetCDF4 files; and (2) make this code agnostic to the current dependencies installed and to `xarray.to_netcdf`'s internal rules about how it decides what kinds of "netcdf" files to write by default (e.g., right now, I think the rule is something like: Use NetCDF4 if h5netcdf or netcdf4 is installed; otherwise, fall back to netcdf3...but that might change in the future).

src/swell/utilities/file_system_operations.py
Dooruk: Can you check if what I have now is what you had in mind?

Dooruk:
I really would like to get this PR go in soon and I'm aware other PRs are piling up. For it to function and past code tests etc it will require a part2, but I will work on that once I address potential reviewer comments here. 

This particular PR aims to achieve a few main things, so if you are reviewing please focus on those aspects mainly:

- Introduce `RunGeos`; which might be summarized as a "transient" task that will only live under suites that runs GEOS.  This task calls `gcm_run.j` directly from Cylc.
- Make GEOS running tasks more application specific, e.g. `PrepCoupledGeosRunDir` would only be used by coupled suites.  
- `GetCoupledRestarts`, new task, can use Discover restarts or R2D2 (this is a placeholder option for now) 
- Switch default to GEOSv12rc**, the rumor mill says an official release is imminent. SWELL based ADAS and ODAS will use GEOSv12+. I know this is not necessarily the case for GEOS-CF. 
- Start using model specific tasks

With the risk of repeating myself (for more details, read the description of this PR), Ricardo, Maryam, and I had a long discussion last week on executing GEOS in SWELL. It boils down to these two ways:

1) (This PR) Copy an existing GEOS experiment folder, make minimal changes, and call `gcm_run.j` directly
2) Current approach, point to an external GEOS experiment folder, copy everything, break down `gcm_run.j` into Python methods, as it is done in `prep_geos_run_dir.py` and execute GEOSgcm.x using Python `subprocess` in `RunGeosExecutable`.

I am exclusively switching to method 1, @mer-a-o  voiced desire to implement some type of method 2 (of which she has a Skylab application). @rtodling is not fully convinced either way, but leaning towards 1.

I'm leaving the second option intact for now to help @mer-a-o choose between 1 & 2. Unfortunately, I'm not able to run GEOS v12rc** using method 2 on Milan nodes so approach 2 won't work for GEOSv12** for now (most likely due to using `mpirun` in `subprocess` approach). FWIW, GEOS-CF currently uses GEOSv10, I don't know what is the target version for a JEDI based release. @mer-a-o  is looking into `gcm_setup` and such at the moment so she can provide her input.

Anyways, happy to have a dedicated meeting on this, maybe even include @tclune. There are just so many different modes of running `gcm_run.j so IMO method 2 makes it hard for SWELL to support multiple versions of GEOSgcm.

src/swell/suites/forecast_geos/flow.cylc
```diff
@@ -92,20 +87,17 @@
...
-        script = "swell task MoveForecastRestart $config -d $datetime"
+    [[MoveEraseForecastRestart]]
+        script = "swell task MoveEraseForecastRestart $config -d $datetime"
 
     [[SaveRestart]]
         script = "swell task SaveRestart $config -d $datetime"
 
     [[RunGeosExecutable]]
-        script = "swell task RunGeosExecutable $config -d $datetime"
+        script = "{{experiment_path}}/forecast/gcm_run.j"
```
mer-a-o: Earlier you have: `"{{experiment_path}}/GEOSgcm/forecast/gcm_run.j"`
Does it make sense to keep it consistent or it's OK like this?

tclune:
@Dooruk Happy to meet when convenient, but lets try to involve Matt and Shayon as they have much more experience with the details here.

mer-a-o:
@Dooruk I tried running `3dvar_cycle` experiment and got this error:


```
PrepGeosRunDir:   Swell called ABORT: In config class, trying to get variable 'geos_experiment_directory' but this variable was not created. Ensure that the variable is in the experiment configuration and specified in the question list for the task (task_questions.py).
```

Dooruk:
> @Dooruk I tried running `3dvar_cycle` experiment and got this error:
> 
> ```
> PrepGeosRunDir:   Swell called ABORT: In config class, trying to get variable 'geos_experiment_directory' but this variable was not created. Ensure that the variable is in the experiment configuration and specified in the question list for the task (task_questions.py).
> ```

Sorry, can you try with this branch: `feature/exec_geos_direct_part2` (which will fail during initilzation:)

> @Dooruk Happy to meet when convenient, but lets try to involve Matt and Shayon as they have much more experience with the details here.

Sure, I will bring this up in our dev meeting.

src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml
```diff
@@ -19,12 +19,18 @@ existing_jedi_source_directory:
...
+  default_value: /discover/nobackup/projects/gmao/advda/SwellStaticFiles/geos/homdirs/coupled_5deg
+
 geos_experiment_directory:
   default_value: 5deg_0701
 
 geos_restarts_directory:
   default_value: restarts_20210701_210000_5deg
 
+initial_restarts_method:
+  default_value: directory
```
mer-a-o: Is `r2d2` also an option or that is not yet implemented?

src/swell/utilities/geos.py
```diff
@@ -388,18 +388,36 @@ def rc_to_bool(self, rcdict: dict) -> dict:
...
+        """
+        # Validate directory exists
+        if not os.path.exists(forecast_dir):
+            self.logger.abort(f'Directory does not exist: {forecast_dir}')
+
+        self.logger.info(f'Renaming *_checkpoint files to *_rst in {forecast_dir}')
+
+        # Use glob to find checkpoint files instead of relying on shell commands
+        checkpoint_pattern = os.path.join(forecast_dir, '*_checkpoint')
+        checkpoint_files = glob.glob(checkpoint_pattern)
```
mer-a-o, line 405: Do you need to import glob first?

src/swell/tasks/geos_marine/move_erase_da_restart-geos_marine.py
```diff
@@ -0,0 +1,175 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+
+# --------------------------------------------------------------------------------------------------
```
mer-a-o, line 8: What is the difference between `tasks/move_erase_da_restart.py` and this file? Do we need both?

src/swell/tasks/move_forecast_restart.py
```diff
@@ -32,65 +32,54 @@ def execute(self) -> None:
...
-    def cycling_restarts(self) -> None:
+    def move_restarts(self) -> None:
 
         # Move restarts (checkpoints) in the current cycle dir
         # ------------------------------------------------------
         self.logger.info('GEOS restarts are being moved to the next forecast dir')
         self.logger.info('Finding _checkpoint restarts')
 
-        src = self.forecast_dir('*_checkpoint')
+        src = self.forecast_dir('scratch', '*_checkpoint')
```
mer-a-o: can `forecast_dir` take two arguments?

src/swell/tasks/prepare_analysis.py
mer-a-o: Here should be `ana_pth`

src/swell/tasks/prepare_analysis.py
```diff
@@ -37,19 +44,18 @@ def execute(self) -> None:
...
         self.cc_dto = self.cycle_time_dto()
 
         window_length = self.config.window_length()
 
         # GEOS restarts have seconds in their filename
-        # --------------------------------------------
+        # We want to use rst_dto at the beginning of the DA window (window offset is negative)
+        # ---------------------------------------------------------
         an_fcst_offset = self.da_window_params.analysis_forecast_window_offset(window_length)
-        rst_dto = self.geos.adjacent_cycle(an_fcst_offset, return_date=True)
```
mer-a-o, line 52: `adjacent_cycle` is used in other places like `move_erase_da_restart.py`. Do you need to update that there too? Do we need to keep `adjacet_cycle` functionality if it's not used any more?

src/swell/utilities/file_system_operations.py
```diff
@@ -20,16 +20,13 @@ def copy_to_dst_dir(logger: Logger, src: str, dst_dir: str) -> None:
...
     """
 
-    try:
-        if not os.path.isfile(src):
-            logger.info(' Copying files from: '+src)
-            shutil.copytree(src, dst_dir, dirs_exist_ok=True)
-        else:
-            logger.info(' Copying file: '+src)
-            shutil.copy(src, dst_dir)
+    if not os.path.isfile(src):
```
mer-a-o, line 23: Why try and Expection were removed?

src/swell/tasks/get_coupled_geos_restart.py
```diff
@@ -0,0 +1,172 @@
...
+        if expdir_different and os.path.exists(self.geos_expdir):
+            geos_expdir_path = os.path.join(swell_exp_path, 'GEOSgcm', 'GEOS_expdir')
+            # Check if link already exists, unlink first if it does
+            if os.path.lexists(geos_expdir_path):
+                self.logger.info('Unlinking existing GEOS EXPDIR link: ' + geos_expdir_path)
+                os.unlink(geos_expdir_path)
+            self.logger.info('Linking GEOS EXPDIR to ' + geos_expdir_path)
+            os.symlink(self.expdir, geos_expdir_path)
+        elif expdir_different and not os.path.exists(self.expdir):
+            self.logger.abort('GEOS_expdir directory does not exist: ' + self.expdir)
```
mer-a-o: Can these if statements lines 75-93 be simplified?

src/swell/suites/3dfgat_marine_cycle/flow.cylc
```diff
@@ -0,0 +1,263 @@
...
+    [[EvaObservations-{{model_component}}]]
+        script = "swell task EvaObservations $config -d $datetime -m {{model_component}}"
+        platform = {{platform}}
+        execution time limit = {{scheduling["EvaObservations"]["execution_time_limit"]}}
+        [[[directives]]]
+        {%- for key, value in scheduling["EvaObservations"]["directives"][model_component].items() %}
+            --{{key}} = {{value}}
+        {%- endfor %}
+
+    [[SaveRestart-{{model_component}}]]
```
mer-a-o, line 244: `SaveRestart` is already defined.

mer-a-o (commented):
Sorry the review took so long. I tried my best to go over most of the changes and also used Claude Code to help me review this PR.

src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml
Dooruk: Not implemented in this PR. We have no restarts stored anyway.

I do envision though, get_*_restart task could/should be model/application specific, especially the R2D2 part.

src/swell/suites/forecast_geos/flow.cylc
Dooruk: Good catch. I will address it.

I didn't finalize if relevant suites are working, this part1 PR is meant to represent both execution approaches.

src/swell/tasks/prepare_analysis.py
Dooruk: Nope, we can get rid of it.

src/swell/tasks/move_forecast_restart.py
Dooruk: I see this is confusing. Alexey also suggested `forecast_dir` should only take one type of input, I will make its input `list` only in another PR:

https://github.com/GEOS-ESM/swell/pull/677#discussion_r2622837351

but to answer your question, yes

src/swell/utilities/geos.py
Dooruk: Yep, that is the case in `part2`

src/swell/tasks/base/task_base.py
Dooruk: Good point, I will address it in a future PR.

For documenting, I will implement that going forward and here in `part2`.

src/swell/utilities/file_system_operations.py
Dooruk: I tried addressing @shiklomanov-an's concern here:

https://github.com/GEOS-ESM/swell/pull/677#discussion_r2623045493

Not sure if this is what he had in mind.

ghost: I think this is better! If any of these operations fails, it will throw a precisely typed exception with a precise error message.

Previously, if this operation failed, we would immediately throw an error anyway, except that we would drop all useful error information and replace it with a generic "Copying failed..." message.

`try: ... except: ...` blocks only make sense if we actually _do_ something to _handle_ the exception (e.g., if copying fails, try looking for the file in a different place, or create a blank file, or do some additional analysis for common problems and then raise a precise exception with the results of that analysis).

src/swell/tasks/geos_marine/move_erase_da_restart-geos_marine.py
Dooruk: This is just to demonstrate new capability Michael added. Swell checks for model-specific tasks under `/tasks/model` folder first, if it can't find it uses the generic task under `/tasks`. 

So currently, generic one and model-specific one is identical, but I will erase marine specific section of this task in `part2`.

Dooruk:
I addressed all of the concerns here, part2 https://github.com/GEOS-ESM/swell/pull/703 has more changes that makes this particular PR merge ready. I am still working on that. In the meantime happy to address any additional concerns.

For part2 I need @mer-a-o's input. As we discussed before currently I'm proposing two ways to execute GEOS. I don't want to put you on the spot but I need to know which way you are leaning towards in terms for forecast between below 3 options. Sounds like I'm convincing @rtodling towards option 1 but no pressure 😄.:

1) Adapt to executing gcm_run.j directly. I'm happy to work with ADAS or CF group in terms of adaptation/explanation. This is reasonable with the way `gcm_setup` is structured.

2) Use existing approach of executing `GEOSgcm.x` directly (this will require some modifications on my end and your end. I can address as best as I can in part2 then hand it to you). As I mentioned before, `subprocess` part didn't work for me on Milan nodes, but I can walk you through the execution part. 

3) Start from scratch, recreate your Skylab approach, which is somewhat similar to option 2 approach, but uses a shell script to execute `GEOSgcm.x`.

4) ?? Do you have any other suggestions?

This is not an end-all be-all PR, there is still ensemble forecast part of the puzzle.

@mathomp4 @sshakoor1 my proposed approach in this PR is to call `gcm_run.j` directly from Cylc. The relevant part for you is here:

https://github.com/GEOS-ESM/swell/blob/23bc8a5355d8a19c679322ce4a4ada76ecd759fb/src/swell/suites/3dfgat_coupled_cycle/flow.cylc#L163-L171

The reason I'm tagging you is that if `gcm_run.j` is executed in Cylc/SWELL some post-processing parts in `gcm_run.j` need to be deactivated. My csh and gcm templated suggestion is to make certain parts of `gcm_run.j` conditional on if SWELL is used or not. Could be a really dumb suggestion depending on the complexity of the gcm_run.j.tmpl structure. We can discuss this during an upcoming SWELL dev meeting.

mer-a-o:
@Dooruk, tough question! :)
For weather, I'm leaning toward option 1 with `gcm_setup`, since there are already many pieces in place for non-cycling experiments and also marine is using it, which will help me understand all the details and caveats before going further.

For CF, I'd start from scratch since the setup is simpler compared to weather. If that goes well, I'd look to implement something similar for weather later.

I did something similar in Skylab, started with running the forecast via gcm_run.j, and later moved all tasks under it into Skylab. So this feels like a natural path.

I'm open to suggestions if you or others see it differently.

Dooruk:
> @Dooruk, tough question! :) For weather, I'm leaning toward option 1 with `gcm_setup`, since there are already many pieces in place for non-cycling experiments and also marine is using it, which will help me understand all the details and caveats before going further.
> 
> For CF, I'd start from scratch since the setup is simpler compared to weather. If that goes well, I'd look to implement something similar for weather later.
> 

Sure, sounds like the best option! I will work on part2 accordingly.

Dooruk:
This is ready to go in and/or final testing. CI-workflows need to be modified to handle new suite names though.

src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml
```diff
@@ -19,11 +19,11 @@ existing_jedi_source_directory:
...
   default_value: /discover/nobackup/projects/gmao/advda/jedi_bundles_sles15/current_pinned_jedi_bundle/source/
 
-geos_experiment_directory:
-  default_value: 5deg_0701
+geos_homdir:
+  default_value: /discover/nobackup/projects/gmao/advda/SwellStaticFiles/geos/homdirs/coupled_5deg
 
-geos_restarts_directory:
-  default_value: restarts_20210701_210000_5deg
+initial_restarts_method:
```
mer-a-o, line 25: `initial_restarts_method` is not set in `nccs_discover_cascade`. Is Cascade used at all?

jeromebarre:
@Dooruk For the sake of clarity and simplicity, would it be possible to break down the PR in two for the Main change 1 and 2? They are two disctinct feature improvements here. That would also speed up and facilitate the reviews. Thanks a lot!

Dooruk:
> @Dooruk For the sake of clarity and simplicity, would it be possible to break down the PR in two for the Main change 1 and 2? They are two disctinct feature improvements here. That would also speed up and facilitate the reviews. Thanks a lot!

I already did this. I'm running out of time before I go on FMLA so I won't be able to do that again. Main change 1 is just renaming suites anyway. 

I created this PR after your similar comment, it was only part 1 for a couple of months or so. @mer-a-o reviewed it and decided she will implement the Skylab approach to execute GEOS for compo, and we will collaborate for running NWP afterwards. This is not end all be all in terms of executing GEOS but it is needed urgently for marine group to run some experiments while I'm gone.

Dooruk:
Once https://github.com/GEOS-ESM/CI-workflows/pull/32 goes in, this is good to go in but I need approvals

mer-a-o:
I was able to run the `3dvar_marine_cycle` experiment successfully. I reviewed parts of this PR and believe my comments on those sections have been addressed. It would have been preferable to separate the renaming and documentation updates into a different PR, but given @Dooruk’s availability, I think it’s fine to merge this one. We can always go back and fix any issues we encounter along the way.

mer-a-o (dismissed):

src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml
Dooruk: NCCS just sent out an email on Monday that they are decommissioning Cascade in 3-4 months and I don't know anyone who uses it for SWELL anyway. So I think we can slowly start avoiding changes to it.

mranst (changes requested):
Lots of changes, I appreciate the documentation upgrades. I don't fully know the intricacies of the change to gcm_run.j, but from what I can see, this looks great. Only one thing sticks out to me, which is the eva comparison task for observations uses the suite name, so if you could change the names of these files to reflect the new marine names, I can give an approval:

`src/swell/suites/compare/eva/comparison_observations-3dfgat_cycle-geos_marine.yaml'
'src/swell/suites/compare/eva/comparison_observations-3dfgat_cycle-geos_marine.yaml`

mranst (approved):
Looks good, thanks!

Dooruk:
Can't keep up with the pace of changes in develop:) thank you all for reviews

PR #684, opened 2026-01-23
Switch to pyproject.toml

mranst:
## Description

This PR replaces `setup.py` with the more modern `pyproject.toml`. Newer versions of pip (such as those in the new spack-stack) are starting to throw deprecation warnings for `setup.py`.

I've tested this works for:
- [x] Static pip install
- [x] Editable pip install (uv and non-uv)
- [x] On-prem tests


Dooruk:
Could this potentially impact/improve `requirements.txt` somehow or is that irrelevant to this?

And my favorite question, should we change any documentation?

pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+    'suites/*/*',
+    'suites/*/*/*',
+    'tasks/task_questions.yaml',
+    'test/suite_tests/*.yaml',
+    'configuration/*',
+    'configuration/*/*',
+    'configuration/*/*/*',
+    'configuration/*/*/*/*',
+    'configuration/*/*/*/*/*',
+    'configuration/*/*/*/*/*/*',
```
Dooruk: I know this was also the case for `setup.py`, but is there a clean one-liner approach for this by any chance?

mranst: Yep, recursive wildcards are a thing

```python
configuration/**
```

mranst:
> Could this potentially impact/improve `requirements.txt` somehow or is that irrelevant to this?
> 
> And my favorite question, should we change any documentation?

This shouldn't change anything with the requirements. The install method should be identical to the user, I did make one change in the docs related to the setuptools version

Dooruk (approved):
Thanks, might be placebo but uv pip install step for me is faster now 👍🏼

shiklomanov-an (approved):
Excellent!

Re: `requirements.txt` --- We can save it for a future PR, but `pyproject.toml` should also replace/subsume `requirements.txt`.

Everything that is currently in `requirements.txt` should go into `[project] -> dependencies = [...]`; e.g.,

```toml
[project]
dependencies = [
    "click>=8.0.0",
    "jinja2>=3.0.3",
    "pyyaml>=6.0",
    "pycodestyle>=2.11.0",
    "numpy<2",
    "pandas>=1.4.0",
    "isodate>=0.5.4",
    "f90nml>=1.4.3",
    "questionary>=1.10.0",
    "h5py>=3.7.0",
    "flake8==6.1.0",
    "netCDF4",
    "ruamel.yaml==0.17.16"
]
```

`pyproject.toml` also elegantly handles optional dependencies for different features. Here's an (AI-generated) example:

````toml {pyproject.toml}
[project]
dependencies = [
    "click>=8.0.0",
    "jinja2>=3.0.3",
    # ...other core deps...
]

[project.optional-dependencies]
dev = [
    "pytest",
    "black",
    "flake8==6.1.0"
]
github = [
    "some-github-only-package"
]
docs = [
    "sphinx",
    "myst-parser"
]
````

...that can then be installed with:

* `pip install .` -- core dependencies only
* `pip install .[dev]` -- core + `dev`
* `pip install .[dev,github]` -- core + `dev` + `github`

Note that when you install multiple things (e.g., `.[dev,github]`), pip will resolve everything into a single set of packages, so the versions have to be consistent. You can't have `dependencies = ["numpy<2"]` and do `optional-dependencies.dev = ["numpy = 2.1"]`. But you _can_ be more specific about your dependencies in one or the other place, as long as they're mutually compatible. E.g., If we wanted to test against several `numpy` versions, you can do things like `dependencies = ["numpy<2"]` and `optional-dependencies.np19 = ["numpy==1.9"]`, `optional-dependencies.np18 = ["numpy==1.8"]` and then have CI run tests with `pip install .[np18]`, `pip install .[np19]`, etc. (but `pip install .[np18,np19]` will fail because pip will try to satisfy both constraints, which is impossible).

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements

mranst:
> Re: `requirements.txt` --- We can save it for a future PR, but `pyproject.toml` should also replace/subsume `requirements.txt`.

Cool, I didn't know it could handle stuff like that. I'm going to go ahead and merge this one and move the requirements to a separate PR. It may take a little coordination to adopt this change. If we want to move to this syntax, we'll need to update CI-workflows, which may mess up people testing PR's not yet merged with develop. We could potentially leave requirements.txt in for a bit while people migrate

PR #686, opened 2026-01-26
Migrate requirements to pyproject.toml

mranst:
This PR moves the requirements to `pyproject.toml`. Github requirements are specified in a separate section of `pyproject.toml`. This will impact how users install swell, it basically just eliminates the need for separately installing requirements using the `--requirements` tag. Docs for all install methods have been updated. I've left `requirements.txt` present with a deprecation warning for the time being. This will give users a chance to move over to the new installation method without disruption before removing it completely.


docs/platforms/discover/installing_swell_uv_venv.md
```diff
@@ -36,9 +36,8 @@ git clone https://github.com/GEOS-ESM/swell.git swell-develop
 4) Load all the modules that SWELL needs: `mod_swell` (this is the `bash` function created in the preliminary steps)
 5) Create a Python virtual environment: `uv venv`
 6) Activate the virtual environment: `source .venv/bin/activate`
-7) Install SWELL dependencies `uv pip install -r requirements.txt`
-8) Install SWELL in editable mode: `python -m pip install -e .`. (note: make sure you run this while the `venv` is active)
-9) Now, work on SWELL. Any changes you make to the SWELL source code will be automatically applied to the install (because it's an editable install); no need to manually reinstall.
+7) Install SWELL in editable mode: `python -m pip install -e .`. (note: make sure you run this while the `venv` is active)
```
ghost: Why not use `uv pip install` here? That will work whether or not the `.venv` is active, and should be more performant because it will use `uv`'s much faster pip implementation.

shiklomanov-an (dismissed):
Looks good to me! Probably the sequence of PRs should be:

1. This one
2. Change CI-Workflows to use these build/install commands.
3. Here -- remove `requirements.txt`.

mranst:
> Looks good to me! Probably the sequence of PRs should be:
> 
>     1. This one
> 
>     2. Change CI-Workflows to use these build/install commands.
> 
>     3. Here -- remove `requirements.txt`.

Thanks, makes sense. The application tests should continue to run fine on develop with how it's done there, but if we switch to the new install commands it won't work for people's branches who haven't caught up yet, so I'm thinking I'll give it a little bit before changing it

docs/platforms/discover/installing_swell_uv_venv.md
Dooruk: Do you mean uv will first look for .venv even if it's not active?

Dooruk (approved):
Looks good! `uv pip install -e .` improved my life already.

If we are looking into making some changes to CI-workflows, is it worth revisiting:
https://github.com/GEOS-ESM/CI-workflows/pull/11?

docs/platforms/discover/installing_swell_uv_venv.md
ghost: Correct. uv is quite aggressive in requiring project-local environments. If a `.venv` exists, `uv pip install` will use it. If a `.venv` doesn't exist, `uv pip install` will fail with this clear error:

"error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment"

You have to force `uv` to not install into `.venv` via the `--system` flag.

PR #691, opened 2026-01-30
Define model types/lifetime for experiments and remove r2d2_local_path

### LINKED ISSUE #649: Define model types and lifetime for experiments.
This needs to be more straightforward.

### LINKED ISSUE #670: Remove r2d2_local_path
r2d2_local_path is no longer being used and should be removed.

ftgoktas:
## Description
- Makes model types and lifetime for experiments more straightforward.
- Removes unused `r2d2_local_path` and any references to it.

Closes #649, #670


src/swell/utilities/r2d2.py
```diff
@@ -10,54 +10,36 @@
...
 from swell.utilities.logger import Logger
 
 # --------------------------------------------------------------------------------------------------
+# R2D2 Model Name Mapping
+# --------------------------------------------------------------------------------------------------
+# To add a new model:
+#   1. Register the model in R2D2 (with r2d2.register_model())
+#   2. Add it to the dictionary below
 
+R2D2_MODEL_MAP = {
```
Dooruk, line 24: Two things here, 

1) `geos_marine` will have two models for DA for now: `mom6` and `cice6`.  So storing and getting will require special handling. So for this PR:

```py
R2D2_MODEL_MAP = {
    'geos_atmosphere': ['geos'],
    'geos_marine': ['mom6', 'cice6']
}
```
Might be a better approach for consistency between the dictionary items.

2) We need `geos_cf` as well now

mer-a-o: for geos_cf both should be `geos_cf`.

mer-a-o (commented):
Changes look good. Only one comment about `geos_cf`. 
I see experiment lifetime information is added as comments. I think lifetime can only be implemented once we sort out experiment id (https://github.com/GEOS-ESM/swell/issues/701).

src/swell/utilities/r2d2.py
Dooruk: To follow up on this, does this mean we need to define a new model type in `r2d2-client`, `cice6`?

I assume `geos_cf` is available already.

Dooruk: I added more context here:

https://github.com/GEOS-ESM/swell/issues/649#issuecomment-3961104741

Dooruk:
@ftgoktas, `cice6` is now added as a model. This can go forward with `list` values for `R2D2_MODEL_MAP` dictionaries.

ftgoktas:
I addressed the overwrite problem as well. It was happening because all `geos_marine` files were stored with a single R2D2 model name, I added  `r2d2_model` key to each file entry in the `geos_marine` JEDI config, so `MOM.res` entries carry `r2d2_model: 'mom6'` and `cice.res` entries carry `r2d2_model: 'cice6'`.

Dooruk:
@ftgoktas, did you try running other suites  like `3dvar_marine_cycle`?

I triggered the CI and cycling marine suites failed:

https://github.com/GEOS-ESM/swell/actions/runs/23017924302

ftgoktas:
Should be good now.

mer-a-o (dismissed):
Looks good. Thanks!

mer-a-o (approved):

mer-a-o:
@ftgoktas A minor comment: remove the imports instead of commenting them out?

mranst (approved):

ftgoktas:
> @ftgoktas A minor comment: remove the imports instead of commenting them out?

Right, since the whole script was already commented out, I figured the imports could stay commented out too. I agree that we should normally rely on version control to bring things back, but I think we plan to address `SaveRestart` soon. I'll leave them as-is for now so we don't have to restart the Tier 1 tests that have already started.

PR #692, opened 2026-01-30
GETKF calculation in two steps

metdyn:
### Two-step GETKF approach is implemented
suite=`localensembleda_2steps`

### One-step GETKF approach:  use
suite=`localensembleda`

This PR sets up a two-step calculations for GETKF, aiming to mitigate the computational-intensive/memory intensive/time-consuming HofX calculations that precedes the Kalman Filter solver run. Thus the first step is the observer calculation for HofX on ensemble members for background and save results to `obs_out.nc4`.  In the second step,  the solver for GETKF  reads in `obs_out.nc4` file, then compute the analysis and increment for ensemble mean. 

Until Aug-2026, this two-steps calculation is the preferred method to run full obs-system using GETKF at C90 resolution. With new development in OOPS, we are still to see the performance improvement due to two-step method compared with 1-step calculation. 

This  two-step GETKF implementation runs on 4 Milan nodes with 32 ensemble members at C90 resolution when roughly 30 obs files are each thinned by 75% using Monte Carlo importance sampling (Metropolis scheme). This obs thinning is neither a uniform rejection nor a random rejection, hence I am optimistic about its performance. 

### Run command

`swell create localensembleda_2steps -k`
`swell launch ...`

`swell create localensembleda -k`
`swell launch ...`

### Test
I also tested the letkf_marine suite. 
It is still running.
I have also run `localensembleda` suite. The output is as before. 

### Output from two steps calculations:

As output,  `geos.analysis.mean.20231010_000000z.nc4`  and `geos.mean-inc.20231010_000000z.nc4` will be the final products. 

The T500 mean increment data with file name  `geos.mean-inc.20231010_000000z.nc4` is shown below:
inc_t_500

obs=[aircraft_temperature aircraft_wind amsr2_gcom-w1 amsua_aqua amsua_metop-b amsua_metop-c amsua_n15 amsua_n19 atms_n20 atms_npp avhrr3_metop-b avhrr3_n19 gmi_gpm gps mhs_metop-b mhs_metop-c mhs_n19 mls55_aura omi_aura ompsnm_npp pibal satwind scatwind sfc sfcship sondes ssmis_f17]


metdyn:
I am still working on node assignment for executable runs from this PR.

metdyn:
Temporarily withhold this code change to add more improvements.

metdyn:
I have been inspecting the input and output from the ensemble runs when this PR was initially created and closed. Now I donot see problems with this infrastructure complement. Hence I will have a few other PRs related to local ensemble DA intended to merge into this base PR.

metdyn:
Hi all,

I want people to review this PR.  It is technical and affect local ensemble solver only. I want to get this PR and a few other GETKF related PR merged to develop, so that we can test the scientific part of the algorithm related to JEDI code. Currently @rtodling and me are having problems with the output from local ensemble solver, and I am not concerned with the swell implementation here. 

@mranst mentioned the T2 test failed and I think this is related to the ongoing UKMO work on UFO code
- https://github.com/JCSDA-internal/ufo/pull/4015
- https://github.com/JCSDA-internal/ufo/pull/3772
which I wrote in - https://github.com/GEOS-ESM/swell/issues/705

I modified the yaml input for GETKF to include only sondes, sfcship, atms_n20 and this should make the GETKF tier1 and tier2 test run without error.

mranst (dismissed):
I have a limited understanding of the technical aspects of this PR, but from my perspective it looks good. Tier 1 tests ran without issue.

mo-owenaljabar:
> @mranst mentioned the T2 test failed and I think this is related to the ongoing UKMO work on UFO code

@metdyn @mranst what's the configuration that's giving the failure? The amount of time for localisation to be implemented depends on which correlated R class you're trying to use, as we're treating the different classes one by one. Work on cross variable covariances is currently underway, but within group covariances and the new diffusion based R matrix are further off.

mranst:
> > @mranst mentioned the T2 test failed and I think this is related to the ongoing UKMO work on UFO code
> 
> @metdyn @mranst what's the configuration that's giving the failure? The amount of time for localisation to be implemented depends on which correlated R class you're trying to use, as we're treating the different classes one by one. Work on cross variable covariances is currently underway, but within group covariances and the new diffusion based R matrix are further off.

Tier tests are not currently failing due to `localensembleda`. I tested the suite against today's build of JEDI and it executed correctly, as far as I could tell.

metdyn:
@mo-owenaljabar, thank you for updating us about the work on the cross covariance of R and I have and I will continue to test and make adjustment to the obs being used in the GETKF work. (Only three conventional obs were left in the ymal for this PR).  It was intended to illustrate how the observer-solver two-step framework works.

src/swell/tasks/run_jedi_variational_executable.py
```diff
@@ -161,6 +161,11 @@ def execute(self) -> None:
             run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
                            jedi_config_file, output_log_file, perhost)
         else:
+            mpi_command = "mpirun"
+            if not (perhost is None or perhost == "None"):
+                mpi_command += f" -perhost {perhost}"
+            mpi_command += f" -np {np} {jedi_executable_path} {jedi_config_file} {output_log_file}"
+            print(f'intended mpi_command = {mpi_command}')
```
metdyn, line 168: This print statement is for copy and paste during interactive run when generate_yaml_and_exit is active.

Dooruk:
@metdyn is this ready for a review?

Dooruk:
Sorry, just saw "On Hold" 😄

metdyn:
Hi All, 

This PR is ready and you can launch the suite with

swell create localensembleda_2steps -k
swell launch ...

Dooruk:
Hi all, this PR is ready for review. Per @metdyn's request and considering this has been in the works for a while, could you take a look at this if this is of interest to you? We should merge this before other major PRs.

From my end, I can test the application and check the code but we won't use GETKF in marine so this is more interest for the other model interfaces @rtodling @mer-a-o @jeromebarre 

@metdyn does this work with both July16th and August 14th (gsibec 1.5.0) builds, or only with July 16th currently?

src/swell/configuration/jedi/oops/LocalEnsembleDA.py
```diff
@@ -15,14 +15,29 @@ class LocalEnsembleDA(OopsConfig):
...
+                    'northward_wind',
+                    'air_temperature',
+                    'air_pressure_at_surface',
+                    'air_pressure_levels',
+                    'water_vapor_mixing_ratio_wrt_moist_air',
+                    'cloud_liquid_ice',
+                    'cloud_liquid_water',
+                    'mole_fraction_of_ozone_in_air',
+            ]
+        elif self.jedi_interface == 'geos_marine':
```
mranst, line 30:
```suggestion
        else:
```
Suggest making this `else` so it doesn't cause errors with definition if more interfaces are added

src/swell/utilities/question_defaults.py
```diff
@@ -1765,7 +1819,35 @@ class vertical_localization_lengthscale(TaskQuestion):
...
+        models: List[str] = mutable_field([
+            "all_models"
+        ])
+        prompt: str = "What is the fraction of vertical retained variance for GETKF?"
+        widget_type: WType = WType.FLOAT
+
+    # --------------------------------------------------------------------------------------------------
+
+    @dataclass
+    class vertical_localization_unit (TaskQuestion):
```
mranst:
```suggestion
    class vertical_localization_unit(TaskQuestion):
```

mranst (dismissed):
Small comments, but from my perspective this looks good.

jeromebarre:
I can try to review it by the end of this week most likely next week if that's ok. Thanks for this PR as I am interested to test this for CF.

metdyn:
@Dooruk,

Jedi version is important, `2026-07-15` jedi build should work and `2026-08-14` will fail. 
(`06242026` soca build in the common JediBundles directory also works)

thank @mranst @jeromebarre @mer-a-o for doing and will be doing the review.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -13,6 +13,7 @@
 from ruamel.yaml import YAML
 
 from swell.tasks.base.task_base import taskBase
+from swell.utilities.yaml_utils import replace_key
```
jeromebarre, line 16: It is a duplicate import remove that line.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -100,6 +111,10 @@ def execute(self) -> None:
                 self.config.vertical_localization_ioda_vertical_coord_group())
             self.jedi_rendering.add_key('vertical_localization_function',
                                         self.config.vertical_localization_function())
+            self.jedi_rendering.add_key('vertical_localization_frac_retained_variance',
+                                        self.config.vertical_localization_frac_retained_variance())
+            self.jedi_rendering.add_key('vertical_localization_function',
```
jeromebarre: You don't n4ed to add this line, it is a duplicate of the above in L112

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
```diff
@@ -193,17 +208,18 @@ def execute(self) -> None:
 
         # bypass the writing of HofXs
         # ---------------------------
-        bypass_HofXs = False
+        bypass_HofXs = True
```
jeromebarre: This is hardcoded. Should it be passed as a configurable parameter in the config?

src/swell/utilities/question_defaults.py
jeromebarre: pleease apply the suggestion.

src/swell/configuration/jedi/oops/LocalEnsembleDA.py
jeromebarre: yes as we are interested to use this with CF.

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # seperate each obs and write to disk
+        # -------------------------------------------------------------------
+        driver = jedi_config_dict['driver']
+        driver['run as observer only'] = True
+        driver['read HX from disk'] = False
+        print(f'driver= {driver}')
+
+        observers = jedi_config_dict["observations"]["observers"]
+        npx = 1
```
jeromebarre: This is hardcoded please use configurable parameters for this.

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,228 @@
...
+        npy = 1
```
jeromebarre: hardcoded

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,228 @@
...
+        np = 6 * npx * npy
```
jeromebarre, line 197: hardcoded...

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,228 @@
...
+            )
+        cmd += f"wait \n"
+        print(f'nobs = {i+1}')
+        np_use = (i+1) * np
+        np_total = eval(str(model_component_meta['total_processors']))
+        error_msg = f'{i+1} obs: each {np} cores, np_use: {np_use} vs np_avail: {np_total}'
+        assert np_use <= np_total, error_msg
+
+        if not generate_yaml_and_exit:
+            subprocess.run(cmd, shell=True, stdout=subprocess.PIPE,
```
jeromebarre, line 228: With pipe stdout/stderr I think if any per-obs JEDI run fails, the only message will be "returned non-zero exit status 1" with no log output. But I am not entirely sure.  Maybe we don't need to pipe stdout/stderr?

src/swell/configuration/jedi/oops/LocalEnsembleDA.py
metdyn: Thanks, that makes sense! I have added `else:` statement; leave a placeholder for CF.

src/swell/tasks/run_jedi_variational_executable.py
jeromebarre: but that is out of the scope of the PR which is on the GETKF...

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,228 @@
...
+        #            'ioda vertical coordinate group':
+        #            self.config.vertical_localization_ioda_vertical_coord_group(),
+        #            'localization function': self.config.vertical_localization_function()}
+        # localizations = [horizLoc, vertLoc] if len(vertLoc) != 0 else [horizLoc]
+
+        # Include ensemble localizations and halo types with each observation
+        # -------------------------------------------------------------------
+
+        swell_path = get_swell_path()
+        localization_path = os.path.join(swell_path,
```
jeromebarre, line 160: this hardcoded, why not using configurable parameter?

jeromebarre:
Thanks for this PR. 

There is substantial overlap between  `EtkfSolver` and `LocalEnsembleDaExecutable` could it be merged into one task that takes care of the solver? Maybe somehting to think about for a next PR refactor. 

Is there a need to keep the pair `RunJediHofxEnsembleExecutable` `LocalEnsembleDaExecutable`  alive in the future and therefore the comment about refactoring doesn't even apply?

src/swell/tasks/run_jedi_etkf_observer.py
jeromebarre: we should really try to avoid this kind of shell injection in swell.

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,228 @@
...
+        driver['read HX from disk'] = False
+        print(f'driver= {driver}')
+
+        observers = jedi_config_dict["observations"]["observers"]
+        npx = 1
+        npy = 1
+        np = 6 * npx * npy
+        cmd = """
+        export SLURM_MPI_TYPE=pmi2
+        export I_MPI_PMI_LIBRARY=/usr/lib64/libpmi2.so
```
jeromebarre, line 200: hardcoded slurm environment in shell injection.

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: @jeromebarre, thanks,  the keywords are added to question defaults and they can adjust them in suite_config.py.

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: You are right: the failed observer run will not be easily distinguished. Users have to check if the omb files are complete, as well as the output print out from each run.  When error does occur, I would bisect the obs list to narrow down the root cause.  This script serves as a job submission tool to run a set of observer calculations and users need to make sure each of them can run individually. So far I got a list of obs that can run with GETKF calculations on the geos_atmosphere side.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: Thank you, @jeromebarre, good catch!   Some lines got duplicated when they were deleted from the marine side, and added back later on under an if statement model == atmos.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: You are right, thanks, I will add the yaml control for `bypass_writing_hofx_output`.  Another keyword `do_posterior_observer` also needs to be added to configuration.  I should mention that `run_jedi_local_ensemble_da_executable.py` is purely one step calculation for ETKF. 

In general, writing out hofx output while running ETKF is very costly. I will need to add standalone tasks in workflow  for this purpose.

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: solved

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: It is needed for innovation. I find it is the only way to submit one job that runs ~30 independent MPI processes on 2 nodes (each for a separate obs space using only 6 cores).

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: As swell config has more complexity than hard coding due to mutliple levels of override, I propose we solve this localization_path problem by simply using a standardized directory:  configuration/jedi/interfaces/$model/observations/localization. It is a minimal code change, works for all models, and straigt forward.

src/swell/tasks/run_jedi_local_ensemble_da_executable.py
metdyn: removed

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: In fact, this subprocess shell cmd is the most powerful HPC device used in this script. The script uses only 2 nodes to run 31 HofX tasks for atmosphere, saving compute time for each user, really exciting for discover!

src/swell/utilities/question_defaults.py
metdyn: Thanks, @mranst, good catch!

metdyn:
> Thanks for this PR.
> 
> There is substantial overlap between `EtkfSolver` and `LocalEnsembleDaExecutable` could it be merged into one task that takes care of the solver? Maybe somehting to think about for a next PR refactor.
> 
> Is there a need to keep the pair `RunJediHofxEnsembleExecutable` `LocalEnsembleDaExecutable` alive in the future and therefore the comment about refactoring doesn't even apply?

@jeromebarre,

Thanks for your review!  That helps. 

run_jedi_etkf_solver.py and run_jedi_local_ensemble_da_executable.py are similar except for a few tweaks in yaml. However,  having two file names "run_jedi_etkf_observer.py  run_jedi_etkf_solver.py" is a big win in concept, the rest of the implementation becomes secondary.

`RunJediHofxEnsembleExecutable` is a prior work from a colleague and the focus was using `ensemblePacket` approach to execute the MPI. There were subtle technical details in the code, but the two-step GETKF method have superseded that work, making is obsolete, except the concept, design and method.  It is not seriously used or maintained, hence can be ignored for now.

jeromebarre:
> > Thanks for this PR.
> > There is substantial overlap between `EtkfSolver` and `LocalEnsembleDaExecutable` could it be merged into one task that takes care of the solver? Maybe somehting to think about for a next PR refactor.
> > Is there a need to keep the pair `RunJediHofxEnsembleExecutable` `LocalEnsembleDaExecutable` alive in the future and therefore the comment about refactoring doesn't even apply?
> 
> @jeromebarre,
> 
> Thanks for your review! That helps.
> 
> run_jedi_etkf_solver.py and run_jedi_local_ensemble_da_executable.py are similar except for a few tweaks in yaml. However, having two file names "run_jedi_etkf_observer.py run_jedi_etkf_solver.py" is a big win in concept, the rest of the implementation becomes secondary.
> 
> `RunJediHofxEnsembleExecutable` is a prior work from a colleague and the focus was using `ensemblePacket` approach to execute the MPI. There were subtle technical details in the code, but the two-step GETKF method have superseded that work, making is obsolete, except the concept, design and method. It is not seriously used or maintained, hence can be ignored for now.

In an effort not to accumulate dead/irrelevant/obsolete code, should it be removed in this PR then?

metdyn:
> In an effort not to accumulate dead/irrelevant/obsolete code, should it be removed in this PR then?

No, not now; it is a standalone task and was many months of work. I still want to see the implemented techniques.

src/swell/tasks/run_jedi_etkf_observer.py
```diff
@@ -0,0 +1,231 @@
...
+        self.logger.info(f'{np_use} cores '
+                         f'on minimum {nnode_min} nodes is needed to run etkf_observer!')
+
+        if not generate_yaml_and_exit:
+            subprocess.run(cmd, shell=True, stdout=subprocess.PIPE,
+                           stderr=subprocess.PIPE, check=True)
+        else:
+            print(f'intended mpi_command = {cmd}')
+            self.logger.info('YAML generated, now exiting.')
+
```
mranst, line 233: @jeromebarre Has a good point about the shell injection. Having `shell=True` means that malicious code can potentially be injected into the command through user-specified variables. Check out this code snippet that gets around that, or you can try my branch `feature/mranst/getkf_2steps`. (side note, I'm going to open a pr fixing other instances of this in Swell)

```suggestion
        cmds = []

        cycle_dir = Path(self.cycle_dir())
        for file in cycle_dir.glob('log.*'):
            if file.isfile():
                file.unlink()

        for file in cycle_dir.glob('logfile*'):
            if file.isfile():
                file.unlink()

        for i, obs in enumerate(observers):
            x0 = copy.deepcopy(jedi_config_dict)
            x0["observations"]["observers"] = [obs]
            x0['geometry']['layout'] = [npx, npy]
            observation_name = obs['observation_name']
            tmp_file1 = os.path.join(self.cycle_dir(), f'diag_{observation_name}.yaml')
            tmp_file2 = os.path.join(self.cycle_dir(), f'log.diag_{observation_name}')
            with open(tmp_file1, "w") as f:
                yaml.dump(x0, f)

            cmds.append(['srun', '--exclusive', '--mpi=pmi2', '-n', str(np), jedi_executable_wi_path, tmp_file1, tmp_file2])

        print(f'nobs = {i+1}')
        np_use = (i+1) * np
        nnode_min = int(np_use/126) + 1
        self.logger.info(f'{np_use} cores '
                         f'on minimum {nnode_min} nodes is needed to run etkf_observer!')

        if not generate_yaml_and_exit:
            processes = [subprocess.Popen(cmd, cwd=self.cycle_dir(), stdout=subprocess.PIPE, stderr=subprocess.PIPE) for cmd in cmds]

            for process in processes:
                stdout, stderr = process.communicate()

        else:
            print('intended mpi_command: ')
            for cmd in cmds:
                print(" ".join(cmd))
            self.logger.info('YAML generated, now exiting.')
```

src/swell/tasks/run_jedi_etkf_solver.py
mranst, line 1: I think this task's time limit might need to be increased, it timed out at 20 minutes for me

mranst (changes requested):

src/swell/tasks/run_jedi_etkf_solver.py
metdyn: This solver task should take less than 10 min, if your modification kept my algorithm. You will easily see about 20 nc4 output within 2 min.

metdyn: In fact the observer takes less than 10 min with my setup, the solver only takes ~3 min to run.

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: Thank you, @mranst, for the very good modification!  I am trying a small twist to align your code with my algorithm and I will check if that works. (I know the difference between your code and mine)

src/swell/tasks/run_jedi_etkf_solver.py
mranst: This was on a default `localensembleda_2steps` experiment, without my modification (this branch). I just tried this again and got the same result

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: @mranst,

Does the variation: `cmds.append(['srun', '--exclusive', '--mpi=pmi2', '-n', str(np), jedi_executable_wi_path, tmp_file1, tmp_file2, '&\n'])`  when incorporated into python also run, or you have tested that it had fail?

mranst: The `&` doesn't work because it is a shell process, but it is unneeded. This method also launches all the processes in parallel at the same time, and I found a less than ten second difference between your method and this one

metdyn:
@jeromebarre and @mranst, 

I appreciate @mranst for writing a variation of the python sub process. 

I had spent a lot of time debugging and testing this code months ago, and in this PR is the most efficient and explicit source code to control and run the observer code. 

Some arguments like unsafe or malicious command does not make sense at all, because it is not running `os.bash()`, it is running subprocess.  And the subprocess dies out of the scope when this task is completed. `shell=True` is the way to communicate with `libmpi2` which allows for running multiple MPI processes within one job.  Without it, the code does not run as designed. 

```
        if not generate_yaml_and_exit:
            subprocess.run(cmd, shell=True, stdout=subprocess.PIPE,
                           stderr=subprocess.PIPE, check=True)
```

I checked @mranst's version of the code.  For some reasons, the logic from my PR is altered/lost and the effect is that the code runs in serial mode, hence 30 times slower.   When I try to twist it,  I find the software taste goes bad, because my code is a HPC technique and when run at C360 or C720, I might need to make small adjustment in the future.  I donot want to deal with a unclear python subprocess script that is out of my control for subtle details. 

Screenshot from 2026-09-02 09-31-32

See L213-214;  L225-226  in this picture, and notice the difference:
1. the  `&\n` is dropped from L215
2. the modification breaks the original single process into 30 processes as a sequential run.

src/swell/tasks/run_jedi_etkf_observer.py
metdyn: Thanks for letting me know, very nice that you have tested your modification runs in comparable CPU time as the bash script I had!  (It is understandable that python wrappers slow down )

One question:  in your branch `commit 747dd609`,  I donot see the equivalence of 
```
        export SLURM_MPI_TYPE=pmi2
        export I_MPI_PMI_LIBRARY=/usr/lib64/libpmi2.so
```
Do you find these are not needed either?
(I was in the middle of mering your code, then I see you have deleted one of your version:
```
env= [ **env, **os.environ ]
```

(This leaves me some suspicion, because I worry the overlap in processors when using your code).  Even though, the core mpi command is exclusive, meaning non-overlap. 
```
cmds.append(['srun', '--exclusive', '--mpi=pmi2', '-n', str(np), jedi_executable_wi_path, tmp_file1, tmp_file2])
```

I need to do some test.

src/swell/tasks/run_jedi_etkf_solver.py
metdyn: Thanks for letting know, I will check.

src/swell/tasks/run_jedi_etkf_observer.py
mranst: Yes, I found that adding those two environment variables actually broke mpi_init, for reasons unclear to me. I found that removing them allowed the task to run successfully, though you should definitely test this too

metdyn: @mranst,

I am changing mpirun wrapper to bash command in run_jedi_etkf_observer.py, these mpi runs completes in 13 min. 

In contrast, when I used subprocess.Popen command for 30 obs files in observer step, the computation is unstable. 
I find
    - sometimes the log files could not be generated
    - all the times, these runs cannot complete in 20 min

Explicitly,
      bash command `srun --exclusive --mpi=pmi2 -n np exe obs_i.yaml &`
      is a most effective run command on disocover.

This command line issue in bash versus python process is unrelated to JEDI. These comments are resolved.

src/swell/tasks/run_jedi_etkf_solver.py
metdyn: Resolved.  The posterior observer steps takes a long time and are now disabled.  The solver completes in `389.15 sec`.

metdyn:
Folks,
I have no more update on this branch. 
Please do a cylc test, then this framework is ready to go.

PR #694, opened 2026-02-03
breaking down PR/660 into smaller pieces -- part3, adding hofx_cf experiment

mer-a-o:
## Description

This PR adds the first GEOS-CF experiment, H(x) application, to SWELL.
See https://github.com/GEOS-ESM/swell/pull/660 for details.

## Dependencies
None

## Impact
None


mranst:
The obs plots aren't generated for me as `use_obs` is set to false with the following message:
`IODA tempo_no2_tropo output has 0 location dimension(s)`

Is this intended behavior? Otherwise it runs for me. Thanks for adapting this to the new configuration style

src/swell/configuration/jedi/interfaces/geos_cf/model/background.py
```diff
@@ -0,0 +1,42 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+from collections.abc import Mapping
+from swell.configuration.jedi.interfaces.geos_cf.model.shared import \
+        field_io_names
+
+# --------------------------------------------------------------------------------------------------
+
+state_variables = [
```
jeromebarre, line 15: Do we have a good way to make this more 'dynamic' or templated. I am asking because we will have likely several different suites with different state variables and therefore control variables for the different AC applications we will have. It's fine if we have to create a background file of the like per state definition but I wonder if we can have this done more elegantly in the future?

jeromebarre:
A few elements here:
- which r2d2 data store this suite is using? we need to move away form the jcsda datastore
- we need to populate input data (observation and background) into the swell dedicated data store. I think this links with https://github.com/GEOS-ESM/swell/pull/675 to have this capability
- we need to have a tier1 test for H(x) in CF

mer-a-o:
@mranst thanks for catching that. I initially forgot to add the composition observations to `observation_ioda_names.yaml`. I also noticed that the code doesn’t currently fail in this situation. I’ve updated `get_provider_for_observation` [in this commit](https://github.com/GEOS-ESM/swell/pull/694/commits/80caed4cf225b4bf74fc6f63145b67a16e708c34) so it now fails if the observation is not specified in the YAML file.

For cases where the observation file is empty, it’s fine if EVA simply doesn’t produce any plots.

mer-a-o:
@jeromebarre, I believe the files are stored on the swell datastore: https://github.com/GEOS-ESM/swell/blob/develop/src/swell/deployment/platforms/nccs_discover_sles15/r2d2_config.yaml#L9

@mranst, what is the process of adding this experiment to tier 1 testing? Thanks.

mranst:
> @mranst, what is the process of adding this experiment to tier 1 testing? Thanks.

To add it to the `t1test` command for swell, add a config called `hofx_cf_tier1` in `suite_config.py`. This can just be:
```
hofx_cf = ....

hofx_cf_tier1 = QuestionList(
    list_name = "hofx_cf_tier1"
    questions = [
         hofx_cf
    ]
)
```
To add it to the on-prem application tests, `CI-Workflows` needs to be updated through a PR, you can just copy and paste the blocks for hofx with the name changed to hofx_cf and add it to the requirements for clean up. I can open a PR for that if you want, you also may have to get permission to open a branch on that repo
https://github.com/GEOS-ESM/CI-workflows/blob/main/.github/workflows/swell-tier1_application_discover.yml

mer-a-o:
@mranst I create a branch with changes on the `CI-workflows` repository. Can you please give me the permission to open a PR there?

mranst:
@Dooruk can you grant permission?

src/swell/configuration/jedi/interfaces/geos_cf/model/background.py
Dooruk: hmm, are you thinking these would be more `oops` application dependent?

src/swell/configuration/jedi/interfaces/geos_cf/task_questions.yaml
```diff
@@ -0,0 +1,100 @@
+# What is the duration from the middle of the window when forecasts start?
+analysis_forecast_window_offset:
+  default_value: -PT3H
+  options:
+  - -PT3H
+
```
Dooruk:
```suggestion
```

this key is deprecated and became a method now

src/swell/configuration/jedi/interfaces/geos_cf/task_questions.yaml
```diff
@@ -0,0 +1,100 @@
...
+  default_value: '72'
+  options:
+  - '72'
+
+window_length:
+  default_value: PT6H
+
+window_offset:
+  default_value: PT3H
+
```
Dooruk:
```suggestion
```

also deprecated and became a method

Dooruk (changes requested):
Looks good, two minor changes.

Welcome `geos_cf` ❗

Dooruk:
Question for another PR, do you need an `ingest background` capability? Do you point to existing model outputs stored somewhere?

I tend to run experiments and register backgrounds as I cycle. Otherwise formatting outputs properly to be ingested by R2D2 would be tough.

src/swell/configuration/jedi/interfaces/geos_cf/model/background.py
jeromebarre: No I am thinking that we should template this. We should not have to create a new background.py for a new set of state_variables. We should have a set of yamls with different configurations of state variables. This is done in most of other def (for example see background just below) I am not sure why this isn't the case for state_varibles.

jeromebarre: I am not sure if by oops applcation dependent yon mean that state variables be defined as part of  `def render_oops(self)` in the many files in: https://github.com/GEOS-ESM/swell/blob/develop/src/swell/configuration/jedi/oops/
I would say then yes.

src/swell/configuration/jedi/interfaces/geos_cf/model/background.py
```diff
@@ -0,0 +1,42 @@
...
+def background(template_dict: Mapping) -> Mapping:
+
+    background = {
+        'datetime': template_dict['local_background_time_iso'],
+        'filetype': 'cube sphere history',
+        'provider': 'geos',
+        'max allowable geometry difference': 0.1,
+        'datapath': template_dict['cycle_dir'],
+        'filename': 'bkg.%yyyy%mm%ddT%hh%MM%ssZ.nc4',
+        'state variables': state_variables,
```
jeromebarre, line 36: should this be part of `template_dict`?

src/swell/configuration/jedi/interfaces/geos_cf/model/background.py
jeromebarre: But I think that's an overall swell design issue and this is beyond this PR.

jeromebarre (dismissed):
Ran the H(x) suite fine. 
The obs and bkg are in the swell datastore
I think the test has been added in swell tests.
All good for me.

mer-a-o:
> Question for another PR, do you need an `ingest background` capability? Do you point to existing model outputs stored somewhere?
> 
> I tend to run experiments and register backgrounds as I cycle. Otherwise formatting outputs properly to be ingested by R2D2 would be tough.

Ultimately yes, having an `ingest background` capability can be helpful for the CF work. Currently, we store backgrounds on R2d2 using a separate Python script. 

For cycling we would want to store the forecast files (backgrounds for the next cycle) on R2D2. @Dooruk , I'm curious what makes it tough to them R2D2?

src/swell/configuration/jedi/interfaces/geos_cf/model/background.py
mer-a-o: I think we can set `state_variables` the same way `analysis_variables` are set in `task_questions.yaml`, For [example](https://github.com/GEOS-ESM/swell/blob/develop/src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml#L2). `state_variables` can be longer than `analysis_variables` and we may want users to have less control over that? It can be addressed in a seperate PR as we want to keep it consistent between the 3 applications.

Dooruk:
> Ultimately yes, having an `ingest background` capability can be helpful for the CF work. Currently, we store backgrounds on R2d2 using a separate Python script.
> 
> For cycling we would want to store the forecast files (backgrounds for the next cycle) on R2D2. @Dooruk , I'm curious what makes it tough to them R2D2?

I see. I misdirected @ftgoktas then, I told him ingest background is not really a desirable feature. But we should implement features gradually anyhow, ingest obs first that one later.

The tough part, for me, is organizing all of the experiment outputs in the right folders with the right naming nomenclature, and going forward, handling multiple backgrounds/state files for multiple ensemble files. It is possible sure, but I tend to run my experiments in Swell environment and so I just directly ingest to R2D2.

Dooruk (dismissed):

jeromebarre:
> > Ultimately yes, having an `ingest background` capability can be helpful for the CF work. Currently, we store backgrounds on R2d2 using a separate Python script.
> > For cycling we would want to store the forecast files (backgrounds for the next cycle) on R2D2. @Dooruk , I'm curious what makes it tough to them R2D2?
> 
> I see. I misdirected @ftgoktas then, I told him ingest background is not really a desirable feature. But we should implement features gradually anyhow, ingest obs first that one later.
> 
> The tough part, for me, is organizing all of the experiment outputs in the right folders with the right naming nomenclature, and going forward, handling multiple backgrounds/state files for multiple ensemble files. It is possible sure, but I tend to run my experiments in Swell environment and so I just directly ingest to R2D2.

Yes that is why we want to have some definition and rigor on how we use the r2d2 keys. One worrisome thing currently is the experiment names. This will cause issues with overwriting results without notifications as long as we don't have a random and **unique** expeirment name generation (skylab for example generates a random 8 digits hexadecimal) or a warning system that prevents running a suite if the experiment name already exists in the datastore.  I would prefer to have the former. This deserves it's own issue that I will try to create soon.

src/swell/suites/hofx_cf/eva/observations-geos_cf.yaml
```diff
@@ -0,0 +1,160 @@
...
+  plotting_backend: Emcpy
+  figure_list:
+
+  # Correlation scatter plots
+  # -------------------------
+
+  # JEDI h(x) vs Observations
+  - batch figure:
+      variables: *variables
+      channels: *channels
```
jeromebarre: Noticing things here. We are likely not to use `channel` at all in composition stuff. Is it a required key in EVA?

Dooruk: not required and also not used in marine

mranst (dismissed):
Needs a merge with develop and there might be a few comments left to resolve, but from my perspective this looks good. Great work!

src/swell/suites/hofx_cf/eva/observations-geos_cf.yaml
mer-a-o: I removed channels.

mer-a-o:
Thanks @mranst. I think all the comments are addressed now.

Dooruk (approved):
Thanks! There is a PR on the CI-Workflows adding `hofx_cf` in tier1: https://github.com/GEOS-ESM/CI-workflows/pull/29

Dooruk:
This is good to go in. @mranst sorry it will add a bit more to https://github.com/GEOS-ESM/swell/pull/656 but hope that's ok

mranst:
> This is good to go in. @mranst sorry it will add a bit more to #656 but hope that's ok

No worries

PR #698, opened 2026-02-05
Search for model-specific tasks

mranst:
## Description

Addressing comments in #626, this allows for tasks that exist only in certain model-specific folders to be called. If these tasks are called from a different model and don't have counterpart for that model or a generic version, it will fail. In my opinion, this is the best way to handle this, and we should handle tasks only existing for certain models by baking that logic into the scheduling in flow.cylc, i.e:

```
{% if model_component == 'geos_atmosphere' %}
GenerateObservingSystemRecords-{{model_component}}
{% endif %}
{% endfor %}
```

Though this is somewhat complicated so I'm open to alternatives. This doesn't necessarily need #656 to work, since the scheduling is still templated in jinja there


Dooruk:
> ```
> {% for model_component in model_components %}
> {% if model_component == 'geos_atmosphere' %}
> GenerateObservingSystemRecords-{{model_component}}
> {% endif %}
> {% endfor %}
> ```
> 
> Though this is somewhat complicated so I'm open to alternatives. This doesn't necessarily need #656 to work, since the scheduling is still templated in jinja there

i guess at that point it would be 

 ```
 {% if model_component == 'geos_atmosphere' %}
 GenerateObservingSystemRecords-geos_atmosphere
 {% endif %}
 {% endfor %}
 ```
 
Did you start testing these model specific folders? I'm wondering how would it work in terms of listing available tasks; `swell task --help` if only generic, model-specific or both tasks exist?

I was able to use `geos_marine` specific task in my other `gcm_run.j` PR but I do like this solution you are proposing here rather than running and skipping tasks unnecessarily and crowding cylc workflow.

Also GPT suggested  `module names cannot contain hyphens (-).` in regards to `importlib`, so use `_` instead but this hasn't created a problem for me. Something to keep in mind I suppose.

mranst:
> Did you start testing these model specific folders? I'm wondering how would it work in terms of listing available tasks; `swell task --help` if only generic, model-specific or both tasks exist?
> 
> I was able to use `geos_marine` specific task in my other `gcm_run.j` PR but I do like this solution you are proposing here rather than running and skipping tasks unnecessarily and crowding cylc workflow.

It will list all tasks that are specified in an abstract sense. If you have both:
```
src/swell/tasks/task_example.py
src/swell/tasks/geos_marine/task_example-geos_marine.py
```
you will have `TaskExample` in the available tasks, same as if only one or the other exists. Yet another possible point of confusion, it'd be up to the user to make sure the proper versions for the tasks are being called in flow.cylc. 

> Also GPT suggested `module names cannot contain hyphens (-).` in regards to `importlib`, so use `_` instead but this hasn't created a problem for me. Something to keep in mind I suppose.

I've tested this and had no problem with hyphens, but that is against PEP guidelines so I've changed it

Dooruk:
I'm testing this with a new PR I'm working on. For this I moved a task that no other suite uses to `geos_marine`. So now I have a model specific task `GetEnsemble`. 

https://github.com/GEOS-ESM/swell/tree/feature/letkf_marine_2026

It doesn't show under available tasks (i.e., `swell task --help`) but I'm able to run since it is named `get_ensemble_geos_marine.py`. Is this the intended behavoir?

mranst:
> I'm testing this with a new PR I'm working on. For this I moved a task that no other suite uses to `geos_marine`. So now I have a model specific task `GetEnsemble`.
> 
> https://github.com/GEOS-ESM/swell/tree/feature/letkf_marine_2026
> 
> It doesn't show under available tasks (i.e., `swell task --help`) but I'm able to run since it is named `get_ensemble_geos_marine.py`. Is this the intended behavoir?

I can see it, are you asking why it's not called `GetEnsembleGeosMarine`?
image

Dooruk (approved):
Well, I just reinstalled and it shows at the end for me, so I missed it..

Screenshot 2026-07-13 at 4 46 17 PM

I like that it just shows the task Class

PR #702, opened 2026-02-12
Add task to publish comparison test results

mranst:
## Description

This PR adds a task that will copy relevant text files (namely `jedi_log_comparison.txt`) and plots generated by comparison suites to another location. This is intended to be used by the CI tests to publish these results to Dataportal. By default, no `publish_directory` is set, so this task will not do anything unless it is overridden (It also will not be run unless the comparison fails). A pending PR to `CI-workflows` will enable this feature for tier 2 tests.


src/swell/tasks/publish_comparisons.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # Output path base to copy files to
+        publish_directory = self.config.publish_directory(None)
+
+        # Skip this task if there is no publish directory
+        if publish_directory is None:
+            print('Skipping')
+            return
+
+        print('Not skipped')
```
ghost: Suggest `logger` statements here for consistency.

shiklomanov-an (dismissed):
Looks good!

Minor comment: Definitely not a required change for this PR, but take a look at Python's built-in `pathlib` module. It has a lot of quality-of-life improvements for working with system paths. For example, with this:

```python
from pathlib import Path
```

...you can write this line as something like:

```python
cycle_dir = Path(self.cycle_dir())
publish_location = Path(self.config.publish_directory())
if cycle_dir.is_dir():
    files = cycle_dir.glob("eva/**/*.png")
    out_path = publish_directory / self.__datetime__.string_directory() / self.get_model()
    out_path.mkdir(parents = True, exist_ok = True)
```

The other potential advantage of using `pathlib.Path` in more places is that it improves our ability to type-check places that need paths. Right now, we use `str` annotations for this, but `pathlib.Path` would be stricter (and therefore help catch more bugs earlier).

src/swell/tasks/publish_comparisons.py
```diff
@@ -0,0 +1,67 @@
...
+
+            # Copy eva png files
+            files = glob.glob(os.path.join(self.cycle_dir(), 'eva', '**', '*.png'), recursive=True)
+
+            out_path = publish_location / self.__datetime__.string_directory() / self.get_model()
+
+            out_path.mkdir(exist_ok=True, parents=True)
+
+            for file in files:
+                shutil.copy(file, out_path)
```
ghost: If we're going to use pathlib, let's use it consistently, unless there's a specific reason not to.

- `os.path.*` --> [pathlib equivalents](https://docs.python.org/3/library/pathlib.html#corresponding-tools)
- `glob.glob(...)` --> [`Path.glob`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob)
- `shutil.copy` --> [`Path.copy`](https://docs.python.org/3/library/pathlib.html#copying-moving-and-deleting) (note also the `Path.copy_into` --- seems like it could be handy elsewhere.

ghost (dismissed):
Minor suggestion on consistent usage of `pathlib`. Then, resolve merge conflicts, pass tests, and I think this is good to go.

ghost (approved):
I would also replace the `shutil.copy` calls with `pathlib.Path.copy`, unless there's a good reason not to. But otherwise, I think this looks good, assuming relevant tests pass.

src/swell/tasks/publish_comparisons.py
mranst: Removed os.path and glob.glob, Path.copy is >3.14 only, swell is still on 3.11

mranst:
Thanks! Path.copy is python >3.14, I'll keep that in mind for the future

Issue #700, opened 2026-02-09
rename variational and hofx suites to include `_marine` and `_atmos` label


jeromebarre:
We are starting to have both `_atmos` and `_cf` suite namespaces, but the variational suites without suffixes (3dvar, 3dvar_cycle, 3dfgat_cycle) are all marine-focused. To reduce ambiguity and align with current naming, we should rename these suites to include `_marine`.
Same goes with hofx that is now weather specific.

Proposal
Rename the following suite namespaces:

- [x] `3dvar` to `3dvar_marine`
- [x] `3dvar_cycle` to `3dvar_marine_cycle`
- [x] `3dfgat_cycle` to `3dfgat_marine_cycle`
- [ ] `hofx` to `hofx_atmos`
- [ ] `localensembleda` to `localensembleda_atmos`

Any other suite?

NOTE: `_atmos` is a confusing label as CF is also part of the atmosphere... but that's another issue for another day...

Dooruk:
I edited your description directly. Captain obvious here but this will break the CI-workflows.

jeromebarre:
I think `cf` might not be the best label as well. should we have `marine`, `weather`, `composition` as labels for our suites? @rtodling your take on this?

PR #703, opened 2026-02-12
Run gcm_run.j directly -- part2, rename suites

Dooruk:
More low-level changes to make this change mergeable with develop, still work in progress

See #677 for more details

Partially addresses #700 

Remaining things (addressed in part3):

- [x] Get rid of obsolete entries/tasks
- [x] Documentation update


Dooruk:
@mranst a question for you regarding `suite_config.py`. For this PR, I would like to test the `forecast_coupled_geos` (originally `forecast_geos`) suite, which is not in tier1 tests. I'm trying to change the `cycle_times` but is key only used if it's within model level? When I create with the following my `experiment.yaml` `cycle_times` is still `T00`:

```py

class SuiteConfig(QuestionContainer, Enum):

    forecast_coupled_geos = QuestionList(
        list_name="forecast_coupled_geos",
        questions=[
            sq.all_suites,
            qd.start_cycle_point("2021-07-02T12:00:00Z"),
            qd.final_cycle_point("2021-07-03T12:00:00Z"),
            qd.cycle_times([
                "T12",
            ]),
            qd.geos_build_method("use_existing"),
            qd.forecast_duration("P1D"),
        ],
    )

    forecast_coupled_geos_tier1 = QuestionList(
        list_name="forecast_coupled_geos_tier1",
        questions=[
            forecast_coupled_geos
        ]
    )

```

mranst:
> @mranst a question for you regarding `suite_config.py`. For this PR, I would like to test the `forecast_coupled_geos` (originally `forecast_geos`) suite, which is not in tier1 tests. I'm trying to change the `cycle_times` but is key only used if it's within model level? When I create with the following my `experiment.yaml` `cycle_times` is still `T00`:

Yes, `cycle_times` is usually a model-dependent key, it's also handled weirdly by swell. 

https://github.com/GEOS-ESM/swell/blob/1e92396b1c74b5d1d9578db9aafe515abd803ef1/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L228-L234

You can change this to ->

```python
        # If there are no models and the cycle_times is in the keys then remove the models key from
        # the cycle_times question dictionary
        if 'cycle_times' in self.question_dictionary_model_ind.keys():
            if not self.suite_needs_model_components:
                self.question_dictionary_model_ind['cycle_times'].pop('models')
                if self.question_dictionary_model_ind['cycle_times']['default_value'] == 'defer_to_model':
                    self.question_dictionary_model_ind['cycle_times']['default_value'] = 'T00'
```

Dooruk:
1) Thanks @mranst, I applied your suggestion regarding `cycle_times`. How should i go about `question_defaults` with `cycle_times` then?

2) I had an issue where `qd.geos_homdir` was not being recognized, you can check my `3dfgat_marine_cycle` `suite_config.py` for the key in question. Unlike model-dependent questions (which were already being checked for the defer_to_platform placeholder), model-independent questions were being overwritten by the platform's `task_questions.yaml` even if I had provided a specific value in the `suite_config.py`.

I did notice you have an issue addressing override mechanism. Would your new workflow refactoring PR or future override fix address 1 or 2?

mranst:
> 1. Thanks @mranst, I applied your suggestion regarding `cycle_times`. How should i go about `question_defaults` with `cycle_times` then?
> 
>     2. I had an issue where `qd.geos_homdir` was not being recognized, you can check my `3dfgat_marine_cycle` `suite_config.py` for the key in question. Unlike model-dependent questions (which were already being checked for the defer_to_platform placeholder), model-independent questions were being overwritten by the platform's `task_questions.yaml` even if I had provided a specific value in the `suite_config.py`.
> 
> 
> I did notice you have an issue addressing override mechanism. Would your new workflow refactoring PR or future override fix address 1 or 2?

What is the desired behavior with `cycle_times`? Does that fix not address the problem?

I see you have `geos_homdir` overridden in the `tier2` config for `3dfgat_marine_cycle`, but not the base suite. When I run `swell create 3dfgat_marine_cycle_tier2`, I see the key properly overridden in experiment.yaml

Dooruk:
@mranst, in this PR I did three minor additions following your comments:

1) `cycle_times`. This should be model independent for forecast only suites, e.g. `forecast_geos`. It should also be model-dependent in case different models being active in different cycles. To answer your question, I'm set for what I needed.
2) i did a minor change in `prepare_config_and_suite.py` after override didn't work. However, it works now.
3) I made `experiment.yaml` keep original order, instead of alphabetical. So I had to change `utitilies/dictionary.py`

Unfortunately it is really hard to follow these changes as with everyhing else in this PR. Changes are displayed for everything including Furkan and Maryam's recent PRs.

Rest of the PR is regarding GEOS execution and suite renaming, cylc reordering. It will be incomplete without the part3..

PR #707, opened 2026-02-13
Change localization parameters

metdyn:
## Description

A better estimate for localization parameters in observations/localization directory:
  - lengthscale: 1500.e3
  - max nobs: 10000

## Impact

Have an effect on Local Ensemble DA (still being tested)


metdyn:
The resulting increment graphics is not good from this set of parameters.

metdyn:
I think we can afford:
obs localizations:
  lengthscale: 1500.e3
  max nobs: 10000
  
Because I have thin obs data by 75%.

Dooruk (approved):

Dooruk:
@metdyn, would it be possible to create a tier 1 configuration with c12 an tier 2 with c90? You can see an example for marine here:
https://github.com/GEOS-ESM/swell/blob/develop/src/swell/suites/3dvar_marine_cycle/suite_config.py

I'm testing it with actions, then I can merge this.

Also does this close an open issue?

Dooruk:
Currently, localensembleda takes a very long time with c90 (~40 mins!) we really need to cut that down. Maybe you can create c12 states using JEDI changestate?

metdyn:
@Dooruk, 
I know your concern.  The code slowdown is due to `qd.local_ensemble_use_linear_observer(False)` in develop and C90 resolution is not the leading factor.  Using linear observer has a large speed up in code, but the code may crash until we use two steps in GETKF, i.e., the two steps GETKF PR (https://github.com/GEOS-ESM/swell/pull/692) will cure both run stability and runtime speed issue.  (We should neglect the code speed problem for this PR,  40min for C90.  We will get to 10 min for C90 from PR 692).

Issue #701, opened 2026-02-11
(feature request) Unique experiment ID for R2D2


Dooruk:
When creating an experiment with SWELL, R2D2 should check whether an `experiment_id` already exists in the datastore.

AFAIK in Skylab, each experiment has an 8-character long unique hex id, generated at the runtime.

We can keep the `experiment_id` and have our own hex function create a unique id during `swell create ...` stage and check whether that id is "taken" in R2D2 datastore.

That way scientists can still choose whatever names they please (no profanity ofcourse).

FYI @mer-a-o @jeromebarre

Dooruk:
Question for Skylab users, with a unique experiment ID is an experiment location set to be static?

Jerome is suggesting a `swell delete experiment-id` functionality but that requires experiment-id to be assigned a static location, I think?

mranst:
Just so I'm clear, you're envisioning having an experiment ID that R2D2 references that fits the pattern: `<swell-id>-<8 random hex characters>`? The random hash would be generated at the `swell create` stage?

Dooruk:
Something like that, the hash would be generated at the `swell create` stage, after making sure it is a unique ID in R2D2 database.

I guess it could be two ways:

1) `swell create` -> gives you `<8 random hex characters>`

2) `swell create --use-experiment-id experiment_id` gives you `<swell-id>-<8 random hex characters>`
I was thinking `swell create --override ` would look for `experiment_id` in `override.yaml` with this one. For this option to work, users would only use `experiment_id` to interact with SWELL.

IDK if this is an efficient way. I'm thinking about having unique IDs while giving a chance for people to name their own experiments. Open to suggestions.

mer-a-o:
@Dooruk No, it's not static. Each experiment lives in `exp_id` directory specific to that experiment. 
I think @jeromebarre was talking about an r2d2 delete experiment feature not a swell one.

jeromebarre:
fyi this is how it is done in skylab-ewok-r2d2:

https://github.com/JCSDA-internal/ewok/blob/b972719b6d1b43b863825219244a0d151a1f86d9/src/ewok/bin/create_experiment.py#L50

https://github.com/JCSDA-internal/ewok/blob/b972719b6d1b43b863825219244a0d151a1f86d9/src/ewok/r2d2_util.py#L53-L59

[...]

PR #713, opened 2026-02-25
Add unique experiment ID for R2D2

mranst:
I took an attempt at addressing #701. I'm not as familiar with R2D2 as others so I expect this may require some tweaking

This PR adds a separate experiment ID for R2D2. It will check to see if this experiment already exists within R2D2, and if it does, it will attach a random 8-character hex to the end. By default, the ID is set to the swell experiment ID, so if this is unique within R2D2, it will be accepted. If not, it will have the form <swell_id>-<hex>. This also adds a key called `r2d2_experiment_lifetime`, set to `debug` by default. When `swell create` is run, the `r2d2_experiment_id` is created, and the experiment is registered to `R2D2` with the specified lifetime. One thing I'm unsure about is whether it's ideal for users to be able to set long-term options like `lifetime` by themselves, we could potentially hard-code it to debug or limit the options, if not.

I'm not sure if this interferes with #649


src/swell/utilities/r2d2.py
```diff
@@ -161,5 +166,46 @@ def load_r2d2_credentials(
...
+
+
+def experiment_exists(r2d2_id: str):
+    import r2d2
+
+    try:
+        r2d2.get(item='experiment', name=r2d2_id)
+    except Exception as e:
+        if '400 Client Error' in str(e):
+            return False
```
mranst, line 232: I don't love the way this is handled. I assumed `search` would be a better option for this, but when I tried `search(item='experiment', name=r2d2_id)`, it gave me an `unexpected keyword, name` error. `get` also throws a generic `Exception` when it can't find one with that name, so this is what I came up with

ftgoktas: `r2d2-client` requires to be updated to give the status code for a proper check, and `search_experiment` doesn’t support lookup by name. Looks like parsing the error message with `get()` is the only option for now.

ftgoktas:
> This requires R2D2 to be loaded before the swell create stage. I'm trying to figure out the best way of facilitating this, if anyone has any ideas.

I’ve made some changes to address this on [top of this branch](https://github.com/GEOS-ESM/swell/tree/feature/fgoktas/r2d2_id). `r2d2-client` needs to be loaded before, so it’s now loaded during `swell create` instead of the main modules file. It works without error for now but I think we could eventually add `r2d2-client` as a pip dependency to simplify this.

mranst:
> I’ve made some changes to address this on [top of this branch](https://github.com/GEOS-ESM/swell/tree/feature/fgoktas/r2d2_id). `r2d2-client` needs to be loaded before, so it’s now loaded during `swell create` instead of the main modules file. It works without error for now but I think we could eventually add `r2d2-client` as a pip dependency to simplify this.

Cool, thanks! I have merged your branch. Is `r2d2-client` available in pip currently? I wasn't able to find it

ftgoktas:
> > I’ve made some changes to address this on [top of this branch](https://github.com/GEOS-ESM/swell/tree/feature/fgoktas/r2d2_id). `r2d2-client` needs to be loaded before, so it’s now loaded during `swell create` instead of the main modules file. It works without error for now but I think we could eventually add `r2d2-client` as a pip dependency to simplify this.
> 
> Cool, thanks! I have merged your branch. Is `r2d2-client` available in pip currently? I wasn't able to find it

It's installable via `python -m pip install git+https://github.com/JCSDA-internal/r2d2-client.git`. If we decide to move away from the module system, I think we should eventually have our own r2d2 and r2d2-client fork under the GEOS-ESM GitHub to maintain changes, especially since changes in JCSDA's repo might affect Swell.

src/swell/utilities/r2d2.py
Dooruk: Do we have a `r2d2-client` todo list of sorts?

src/swell/utilities/r2d2.py
```diff
@@ -9,13 +9,61 @@
...
+    'nccs_discover_cascade': {
+        'module_path': '/discover/nobackup/projects/gmao/advda/JediOpt/modulefiles/core',
+        'module_name': 'r2d2-client/112025',
+    },
+}
+
+# --------------------------------------------------------------------------------------------------
+
+
+def load_r2d2_module(logger: Logger, platform: str) -> None:
```
Dooruk, line 36: I agree with Furkan here, not loving this approach so having `r2d2-client` in requirements might be good idea.

src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
```diff
@@ -1,5 +1,4 @@
-# (C) Copyright 2021- United States Government as represented by the Administrator of the
-# National Aeronautics and Space Administration. All Rights Reserved.
+
```
Dooruk: Was this change done by mistake?

Dooruk (changes requested):
We should only allow `debug` for now, @ftgoktas is that something you can impose with our current database? We could do it in the SWELL code but bad faith actors(!) can easily change the source code.

With this change, R2D2 is becoming a necessary component of SWELL. However, I'm wondering if we should also provide an optional argument for `create` to avoid `r2d2`? Maybe discussion item for Monday.

I'm also working on changes to save forecast, analysis, and restart files, particularly forecast first. With my changes I'm keeping it versatile where some users could choose to circumvent R2D2. This might especially be important while trying to transition people to SWELL.

Otherwise, this is pretty much what I had in mind though, thanks @mranst!

src/swell/utilities/r2d2.py
mranst: I like it better in principle too, however `r2d2-client` is a locked repo in `JCSDA-internal`, so we'd have to give credentials to github runners for code tests, and any user installing swell would need credentials. Not insurmountable, but maybe a sticking point

mranst:
> We should only allow `debug` for now, @ftgoktas is that something you can impose with our current database? We could do it in the SWELL code but bad faith actors(!) can easily change the source code.
> 
> With this change, R2D2 is becoming a necessary component of SWELL. However, I'm wondering if we should also provide an optional argument for `create` to avoid `r2d2`? Maybe discussion item for Monday.
> 
> I'm also working on changes to save forecast, analysis, and restart files, particularly forecast first. With my changes I'm keeping it versatile where some users could choose to circumvent R2D2. This might especially be important while trying to transition people to SWELL.

I've changed it to be hard-coded to debug now. We can definitely add a key for skipping R2D2 (I assume just on the storing end, people will still need to get observations, etc.), that would also skip the corresponding tasks in flow.cylc

ftgoktas:
> We should only allow `debug` for now, @ftgoktas is that something you can impose with our current database? We could do it in the SWELL code but bad faith actors(!) can easily change the source code.
> 
> With this change, R2D2 is becoming a necessary component of SWELL. However, I'm wondering if we should also provide an optional argument for `create` to avoid `r2d2`? Maybe discussion item for Monday.
> 
> I'm also working on changes to save forecast, analysis, and restart files, particularly forecast first. With my changes I'm keeping it versatile where some users could choose to circumvent R2D2. This might especially be important while trying to transition people to SWELL.
> 
> Otherwise, this is pretty much what I had in mind though, thanks @mranst!

Database-level enforcement would need to be done in the R2D2 server codebase. I think the current approach (`debug` as default in Swell) should be fine for normal use. We can add stricter enforcement on the R2D2 fork on GMAO's AWS if we decide to.

mer-a-o:
Does using the same `user` for all gmao users impact anything? (my guess is not).
I'm not sure how r2d2 handles registration of experiments with same `name` but different `user`s.

mranst:
> Does using the same `user` for all gmao users impact anything? (my guess is not). I'm not sure how r2d2 handles registration of experiments with same `name` but different `user`s.

I believe that different users can have the same name for separate experiments. The only impact that everyone being under the same username is that the user is more likely to need the hex appended

mranst:
I've tried out the option to skip registering and storing to r2d2, it's under this branch: `feature/mranst/skip_r2d2`

ftgoktas:
> Does using the same `user` for all gmao users impact anything? (my guess is not). I'm not sure how r2d2 handles registration of experiments with same `name` but different `user`s.

The `user` is stored with the `experiment` when registering, R2D2 only allows only the experiment’s user (or an admin) to delete the specific experiment. If everyone uses the same R2D2 user/shares credentials, everyone can delete any experiment from that user (not sure how much we should worry about this).

ftgoktas:
> > Does using the same `user` for all gmao users impact anything? (my guess is not). I'm not sure how r2d2 handles registration of experiments with same `name` but different `user`s.
> 
> I believe that different users can have the same name for separate experiments. The only impact that everyone being under the same username is that the user is more likely to need the hex appended

Experiment names are globally unique, so different users cannot have separate experiments with the same name. There is only one experiment per name in the system.

mranst:
> Experiment names are globally unique, so different users cannot have separate experiments with the same name. There is only one experiment per name in the system.

Ah okay, thanks for clarifying

Dooruk:
Once we start assigning unique user IDs within our own GMAO server deleting experiment IDs should be fine. Admins can erase most but users are only can erase theirs. @ftgoktas globally unique, as in server-wide or database wide?

In terms of governance, I'm thinking it might be best we have multiple project groups/databases (NWP, compo, marine) and each to have a different data curator/admin under GMAO server. I should read Jerome's R2D2 document, maybe it's already there 🫤

ftgoktas:
> Once we start assigning unique user IDs within our own GMAO server deleting experiment IDs should be fine. Admins can erase most but users are only can erase theirs. @ftgoktas globally unique, as in server-wide or database wide?
> 
> In terms of governance, I'm thinking it might be best we have multiple project groups/databases (NWP, compo, marine) and each to have a different data curator/admin under GMAO server. I should read Jerome's R2D2 document, maybe it's already there 🫤

It's database-wide. Uniqueness is enforced within the database R2D2 connects to. If we run separate R2D2 databases per project, each database has its own experiment namespace so the same name could actually exist in different project databases.

ftgoktas:
The PR looks close to merge. One thought: we could add a `--lifetime` flag to `swell create` so users can set the R2D2 experiment lifetime instead of having it hardcoded (e.g. `swell create 3dvar --lifetime=science`).

src/swell/utilities/r2d2.py
```diff
@@ -161,5 +213,46 @@ def load_r2d2_credentials(
...
+    return True
+
+# ----------------------------------------------------------------------------------------------
+
+
+def unique_r2d2_id(swell_id: str, platform: str) -> str:
+    logger = get_logger('CreateR2D2ID')
+
+    # Load credentials to allow search
+    load_r2d2_credentials(logger, platform)
```
ftgoktas:
```suggestion

```

I realized `load_r2d2_credentials(logger, platform)` is being called twice. We can drop the one and just keep it under `prepare_config`

src/swell/deployment/create_experiment.py
```diff
@@ -163,6 +163,32 @@ def prepare_config(
...
         experiment_dict = {**experiment_dict, **slurm_dict}
 
+    # Register the experiment in R2D2
+    # -------------------------------
+    if 'r2d2_experiment_id' in experiment_dict:
+
+        from swell.utilities.r2d2 import load_r2d2_credentials, load_r2d2_module, unique_r2d2_id
+
+        load_r2d2_credentials(logger, platform)
+        load_r2d2_module(logger, platform)
```
ftgoktas:
```suggestion
        load_r2d2_module(logger, platform)
        load_r2d2_credentials(logger, platform)
```

We need to reorder the calls after removing `load_r2d2_credentials` from `unique_r2d2_id` so credentials are set after module load.

ftgoktas (dismissed):

Dooruk (changes requested):
I think this is good to go in, works great 👍🏼 

Just to confirm with @ftgoktas in particular, with this approach users can now use `r2d2_experiment_id` for `fetch` in a future experiment, correct? Except ofcourse the lifetime is `debug` so one has to be careful. This should be mentioned in the, you know it, `docs/`!

Sorry but I am going to ask `/docs` edits for https://github.com/GEOS-ESM/swell/blob/develop/docs/configs/r2d2_v3_credentials.md. Especially since `swell create` part is really intertwined now with R2D2. 

- I'm thinking in the welcome message, we should highlight the website, especially R2D2 related configuration instead of the source code. Lots of people are missing this and asking me about it:
https://github.com/GEOS-ESM/swell/blob/3aeafdf0ed778253220ab3a701fabaef2e2499af/src/swell/utilities/welcome_message.py#L22
- Currently R2D2 config docs is not in the _sidebar and not showing on the website. I am addressing that in my PR but just FYI

ftgoktas:
> Sorry but I am going to ask `/docs` edits for https://github.com/GEOS-ESM/swell/blob/develop/docs/configs/r2d2_v3_credentials.md. Especially since `swell create` part is really intertwined now with R2D2.
> 
> * I'm thinking in the welcome message, we should highlight the website, especially R2D2 related configuration instead of the source code. Lots of people are missing this and asking me about it:
>   https://github.com/GEOS-ESM/swell/blob/3aeafdf0ed778253220ab3a701fabaef2e2499af/src/swell/utilities/welcome_message.py#L22
> * Currently R2D2 config docs is not in the _sidebar and not showing on the website. I am addressing that in my PR but just FYI

Yes, that’s right. Users can use the `r2d2_experiment_id` from a newly created experiment as `obs_experiment` or `background_experiment` in a future experiment’s config to fetch from it.

I thought the R2D2 config was already in the sidebar. I’ll double-check and fix it and document new `swell create` part with the lifetime='debug' info. I agree the welcome message change would be helpful

mranst:
>     * I'm thinking in the welcome message, we should highlight the website, especially R2D2 related configuration instead of the source code. Lots of people are missing this and asking me about it:

How about this?

https://github.com/GEOS-ESM/swell/blob/f132fd281a8de9186fca08977c66897bde9d8d25/src/swell/utilities/welcome_message.py

Dooruk:
> 
> How about this?
> 
> https://github.com/GEOS-ESM/swell/blob/f132fd281a8de9186fca08977c66897bde9d8d25/src/swell/utilities/welcome_message.py

Perfection, let's hope users read it 🤞🏼

ftgoktas (approved):
Looks good to me, thanks!

Dooruk (approved):

PR #716, opened 2026-03-03
Add 3dvar_cf Suite for GEOS-CF

jeromebarre:
This PR adds support for 3D-Var data assimilation on the GEOS-CF 

As  a first step it uses Identity for background error. Subsequent PRs will add BUMP and diffusion.
More obs will be added over time too.


src/swell/tasks/run_jedi_variational_executable.py
```diff
@@ -93,8 +99,8 @@ def execute(self) -> None:
 
         # Atmosphere background error model
         # ---------------------------------
-        if npx_proc is not None and npy_proc is not None:
-            self.jedi_rendering.add_key('gsibec_configuration', self.config.gsibec_configuration())
+        if gsibec_configuration is not None:
```
jeromebarre, line 101: I changed this to allow for future background error covariance methods beyond gsibec. This conditional wasn't really appropriate. This now ensures flexibility for other FV3 configurations like GEOS-CF to adopt different B matrix flavors

jeromebarre:
@mranst I am not sure I have access to the tier1 test src. Let me know if you need me to add the test there.

src/swell/suites/3dvar_cf/flow.cylc
```diff
@@ -0,0 +1,163 @@
...
+
+            # Build JEDI source code by linking
+            CloneJedi => BuildJediByLinking?
+
+            # If not able to link to build create the build
+            BuildJediByLinking:fail? => BuildJedi
+
+            {% for model_component in model_components %}
+            # Stage JEDI static files
+            CloneJedi => StageJediCycle-{{model_component}}
```
mer-a-o: I think you need to run `StageJediCycle` for each cycle and can remove it from this section.

src/swell/suites/3dvar_cf/flow.cylc
```diff
@@ -0,0 +1,163 @@
...
+            RunJediVariationalExecutable-{{model_component}} => EvaObservations-{{model_component}}
+
+            # EvaJediLog
+            RunJediVariationalExecutable-{{model_component}} => EvaJediLog-{{model_component}}
+
+            # EvaIncrement
+            RunJediVariationalExecutable-{{model_component}} => EvaIncrement-{{model_component}}
+
+            # Save observations
+            RunJediVariationalExecutable-{{model_component}} => SaveObsDiags-{{model_component}}
```
mer-a-o, line 70: We also want to save analysis. There's no save_analysis task though. I'm not sure why.

src/swell/suites/3dvar_cf/flow.cylc
```diff
@@ -0,0 +1,163 @@
...
+        script = "swell task BuildJedi $config"
+        platform = {{platform}}
+        execution time limit = {{scheduling["BuildJedi"]["execution_time_limit"]}}
+        [[[directives]]]
+        {%- for key, value in scheduling["BuildJedi"]["directives"]["all"].items() %}
+            --{{key}} = {{value}}
+        {%- endfor %}
+
+    {% for model_component in model_components %}
+    [[StageJedi-{{model_component}}]]
```
mer-a-o: There's no `StageJedi` so you can remove it from here too.

src/swell/suites/3dvar_cf/suite_config.py
```diff
@@ -0,0 +1,64 @@
...
+            qd.window_type("3D"),
+            qd.horizontal_resolution("c90"),
+            qd.npx(91),
+            qd.npy(91),
+            qd.npx_proc(2),
+            qd.npy_proc(2),
+            qd.vertical_resolution(72),
+            qd.analysis_variables(["volume_mixing_ratio_of_no2"]),
+            qd.background_experiment("swell_test"),
+            qd.background_time_offset("PT9H"),
```
mer-a-o: I went down a rabbit hole following `background_time_offset` and trying to understand what it means. It is used to set background_time_dto = cycle_time - background_time_offset. Then background_time_dto is used only as "date" parameter for r2d2.fetch of bias correction files. So, I think you don't need it in this experiment. I removed it and ran one cycle successfully.

src/swell/suites/3dvar_cf/suite_config.py
```diff
@@ -0,0 +1,64 @@
...
+        questions=[
+            sq.common,
+            qd.start_cycle_point("2023-08-05T18:00:00Z"),
+            qd.final_cycle_point("2023-08-05T18:00:00Z"),
+            qd.jedi_build_method("use_existing"),
+            qd.model_components(['geos_cf']),
+            qd.check_for_obs(False)
+        ],
+        geos_cf=[
+            qd.cycle_times(['T18']),
```
mer-a-o: I would remove this. This is redundant with `start_cycle_point` and `final_cycle_point`. If you keep this and change to `final_cycle_point` to run a longer experiment, you will need to add other cycle_times there too: `T00` , `T06`, `T12`.

src/swell/suites/3dvar_cf/suite_config.py
```diff
@@ -0,0 +1,64 @@
...
+        list_name="3dvar_cf",
+        questions=[
+            sq.common,
+            qd.start_cycle_point("2023-08-05T18:00:00Z"),
+            qd.final_cycle_point("2023-08-05T18:00:00Z"),
+            qd.jedi_build_method("use_existing"),
+            qd.model_components(['geos_cf']),
+            qd.check_for_obs(False)
+        ],
+        geos_cf=[
```
mer-a-o, line 35: Why not put everything under `questions`?

src/swell/suites/3dvar_cf/suite_config.py
```diff
@@ -0,0 +1,64 @@
...
+                "tempo_no2_tropo",
+                "tropomi_s5p_no2_tropo",
+            ]),
+            qd.clean_patterns(['*.nc4', '*.txt', 'logfile.*.out']),
+        ]
+    )
+
+    # --------------------------------------------------------------------------------------------------
+
+    _3dvar_cf = QuestionList(
```
mer-a-o, line 56: Minor suggestion: I defined it the otherway for `hofx_cf`. First, the experiment and then `experiment_tier1 = experiment`. It makes more sense to me, logically.

src/swell/configuration/jedi/interfaces/geos_cf/model/varincrement1.py
```diff
@@ -0,0 +1,35 @@
...
+# --------------------------------------------------------------------------------------------------
+
+
+def varincrement1(template_dict: Mapping) -> Mapping:
+
+    experiment_id = template_dict['experiment_id']
+
+    varincrement1 = {
+        'write increment': True,
+        'increment': {
```
mer-a-o, line 22: This works fine. I'm just curious, what's the difference between writing the increment using these keys under `online diagnostics` vs using `increment to structured grid`?

src/swell/suites/3dvar_cf/suite_config.py
jeromebarre: My guess is that this will be most likely useful when cycling to fetch the backgrounds. But not sure at this point happy to remove but somehting will need to readded once we work on the cycling. Your call!

src/swell/suites/3dvar_cf/suite_config.py
jeromebarre: I followed the logic in place in the other suites like 3d_var marine. @Dooruk would like to tell why this structure between `questions` vs `geos_<application>` and `_3dvar_<app>_cycle_tier1` vs `_3dvar_<app>_cycle` is preferrable?

src/swell/suites/3dvar_cf/suite_config.py
jeromebarre: see comment above:
https://github.com/GEOS-ESM/swell/pull/716/changes/BASE..0a27a5381576d3d42123b2087ea1d5496041521b#r2885470752

src/swell/suites/3dvar_cf/flow.cylc
jeromebarre: good catch

src/swell/suites/3dvar_cf/suite_config.py
mer-a-o: For fetching background something else is used: https://github.com/GEOS-ESM/swell/blob/develop/src/swell/tasks/get_background.py#L115

At some point we need to clean some of the time variables up.

src/swell/suites/3dvar_cf/flow.cylc
jeromebarre: Maybe becasue of GEOS IAU. Nevertheless I think it's important to save the JEDI analysis and not just what is produce by IAU after that. I think working on a save analysis task should be a separate PR. I'll create an issue.

src/swell/suites/3dvar_cf/suite_config.py
jeromebarre: yes good catch I am not sure why this line is here...

src/swell/configuration/jedi/interfaces/geos_cf/model/varincrement1.py
jeromebarre: I wonder that too...
I followed what was done with `3dvar_atmos`

src/swell/suites/3dvar_cf/flow.cylc
jeromebarre: this is not super clean but was just placeholder for the actual `StageJedi` for for example the linking B matrix files... this will change in my next PR with B matrix.
I will remove this for now.

src/swell/suites/3dvar_cf/flow.cylc
jeromebarre: same reason as the related comment above. removing for now

src/swell/suites/3dvar_cf/flow.cylc
jeromebarre: I created https://github.com/GEOS-ESM/swell/issues/722

src/swell/suites/3dvar_cf/suite_config.py
jeromebarre: @mer-a-o removed for now. as we discussed, once you have a better picture of all the offset redundancies please open an issue.

mer-a-o (approved):

src/swell/suites/3dvar_cf/suite_config.py
Dooruk: That specificity is needed for coupled applications. You could have different `window type`, `background frequency` etc. for `geos_atmosphere` (or `geos_nwp` ) and `geos_marine`.

I think I addressed tiered question over Teams? I was going to ask if you guys have a toy model setup for geos_cf, say c12 for tier1 tests?

Dooruk (dismissed):
Looks good! Good idea to begin with ID matrix. Were you able to put your static files in `advda`?

I was using BUMP initially for marine stuff so some machinery might be there in the `GenerateBClimatology` task.

To repeat my comment below, and not to address it here but do you have a toy model setup for geos_cf, say c12 for tier1 tests?

mranst (dismissed):

jeromebarre:
> Looks good! Good idea to begin with ID matrix. Were you able to put your static files in `advda`?
> 
> I was using BUMP initially for marine stuff so some machinery might be there in the `GenerateBClimatology` task.
> 
> To repeat my comment below, and not to address it here but do you have a toy model setup for geos_cf, say c12 for tier1 tests?

I could run a convert state to have c12 bkgs but this would need to then be adjusted in the suite_config.py. I can issue a new PR for this.

jeromebarre:
And no I am still not in the `advda` group so I cannot copy the files there. I need to create an issue too so both Maryam and I are added.

mer-a-o (changes requested):
@jeromebarre In the past experments we wrote out increment in cubed sphere and used that in geos-cf anasetting configuration. Can you please add writing out these CS increment files in this PR?

mer-a-o:
I added a branch that adds writing increment in CS to the variational yaml.
We can either merge it in this PR or it can be a separate PR that can go after this one is merged:
https://github.com/GEOS-ESM/swell/tree/feature/mer-a-o/write-inc-cs

jeromebarre:
> I added a branch that adds writing increment in CS to the variational yaml. We can either merge it in this PR or it can be a separate PR that can go after this one is merged: https://github.com/GEOS-ESM/swell/tree/feature/mer-a-o/write-inc-cs

Thanks for that I will merge that branch into the current PR

mer-a-o (dismissed):

mranst (dismissed):
Looks good! Just needs a merge with develop. I'm curious if you are able to merge this by yourself. If not, just let me know

jeromebarre:
I am about to issue one last commit to address the `advda` restrictions. For CF we will use `/discover/nobackup/projects/gmao/geos_cf_dev/SwellStaticFiles` from the GEOS CF project space and groups. So there will be an override somewhere.

mranst:
> I am about to issue one last commit to address the `advda` restrictions. For CF we will use `/discover/nobackup/projects/gmao/geos_cf_dev/SwellStaticFiles` from the GEOS CF project space and groups. So there will be an override somewhere.

You can probably just override this in the suite config. Ideally we'd have this be dependent on platform *and* model in this case, but since I believe we are all using discover, it shouldn't be an issue for now.

mranst (approved):

PR #719, opened 2026-03-03
Bugfixes for OOPs and cylc dependencies

### LINKED ISSUE #714: ninner as a string?
I don't think there is a reason to have this as a string in the jedi yaml.
I would replace this from str to int if we want to enforce the type.

https://github.com/GEOS-ESM/swell/blob/1679a8fe411b389c7d24407fcadb324fdf473f1d/src/swell/configuration/jedi/oops/variational3D.py#L42
https://github.com/GEOS-ESM/swell/blob/1679a8fe411b389c7d24407fcadb324fdf473f1d/src/swell/configuration/jedi/oops/variational4D.py#L45

### LINKED ISSUE #715: CleanCycle missing dependencies in certain suites
The CleanCycle task in the var suites has incomplete task dependencies, causing it to run prematurely before all downstream EVA tasks have completed. This results in critical files being deleted while they are still needed by subsequent tasks, causing job failures...

e.g.:
https://github.com/GEOS-ESM/swell/blob/1679a8fe411b389c7d24407fcadb324fdf473f1d/src/swell/suites/3dfgat_atmos/flow.cylc#L96

https://github.com/GEOS-ESM/swell/blob/1679a8fe411b389c7d24407fcadb324fdf473f1d/src/swell/suites/3dvar/flow.cylc#L79-L80

but this is correctly defined in other suites e.g.:
https://github.com/GEOS-ESM/swell/blob/1679a8fe411b389c7d24407fcadb324fdf473f1d/src/swell/suites/3dfgat_cycle/flow.cylc#L119

mranst:
Closes #714, closes #715


mranst:
Sorry, it should be good to merge now if I can get another approval

Dooruk (approved):

PR #720, opened 2026-03-03
Remove set_obs_as_local

mranst:
This was added to potentially address point 1d of #568, but it currently serves no purpose beyond being confusing. We can add it back when it serves a function


Dooruk (approved):
Thanks, this shouldn't break anything(?) If so can be merged.

Issue #718, opened 2026-03-03
GetObservations (fetch) with the new R2D2v3 is slow


Dooruk:
I'm noticing that with the new R2D2v3, `fetch` is slower. This is especially the case if there are more missing obs. There could be number of reasons, one being the location of the current R2D2v3 datastore located in `gmaoswell` project space.

Unless we have a potential fix on the `r2d2-client` side we can look into running GetObs in parallel, similar to how it is done in Eva tasks. This would be running python in parallel on a single node, so shouldn't be too complicated. Curious though if others are experiencing this?

I wasn't able to test `store` yet because of the experiment register stuff.

Below is comparison with similar observations:

### With R2D2v3
```
CYLC_JOB_INIT_TIME=2026-03-03T15:52:06Z
CYLC_JOB_EXIT=SUCCEEDED
CYLC_JOB_EXIT_TIME=2026-03-03T15:59:03Z
```

### With old R2D2:
```
CYLC_JOB_INIT_TIME=2025-11-23T03:38:24Z
CYLC_JOB_EXIT=SUCCEEDED
CYLC_JOB_EXIT_TIME=2025-11-23T03:38:56Z
```

Dooruk:
So I'm doing multiple simultaneous calls to R2D2 and it is quite slow now. I think I'm testing the scalability by myself 😃 

In the same time I'm deleting some observations from our datahub, ingesting new observations and fetching observations for experiments I'm running. It is hard to quantify the number of simultaneous API calls but I would say more than 50.

Even `swell create` take a while because of this heavy loading, I think this is concerning.

Dooruk:
How slow you may ask 🐌 

```
GetObservations: -----------------------------
GetObservations:         Task Complete
GetObservations: -----------------------------
GetObservations: Timing statistics:
GetObservations: Constructed in 0.6781 seconds
GetObservations: Executed in 1443.4773 seconds
```

mranst:
Closed by #756

PR #721, opened 2026-03-04
GetObservations in parallel

mranst:
Addressing #718. Restructures `GetObservations` to fetch all files in parallel, similar to how eva observations works. It also changes the way fetching empty obs works. It only fetches an empty obs file from r2d2 once, then copies that file for other empty observations in the experiment, to avoid having to fetch the same file repeatedly. (I'm assuming this works as long as the filename matches the target file, unless r2d2 is doing something additional in the fetch stage that this wouldn't work for)

With the number of processes set to 4, for `3dvar` default experiment goes from 150 seconds to 100 seconds, and `3dfgat_atmos`, execution time goes from 255 seconds to 220. I believe this is still not nearly as quick as it used to be, but maybe `r2d2-client` changes can improve that further? I'm sure more processes can increase this, but I don't have a good sense of what we should use for this. GetObs can't run on a compute node, and I imagine I/O load might start to be a concern.

Tier1 tests pass

### DIFF
--- a/src/swell/tasks/get_observations.py
+++ b/src/swell/tasks/get_observations.py
@@ -32,6 +32,7 @@ r2d2_model_dict = {
 
 # ----------------------------------------------------------------------------------------------
 
+
 def run_r2d2_fetch(r2d2_dict: dict) -> None:
 
     """Runs fetch command for all types of obs files
@@ -39,8 +40,9 @@ def run_r2d2_fetch(r2d2_dict: dict) -> None:
     Arguments:
     r2d2_dict: Dictionary of r2d2 fetch parameters, ALSO including additional information including:
     **r2d2_dict['fetch_empty']: bool whether fetching empty obs file is appropriate
-    **r2d2_dict['cycle_dir']: Experiment cycle directory (has to be specified this way for multiprocessing)
-    **r2d2_dict['logger']: Swell logger (has to be specified this way for multiprocessing)
+    **r2d2_dict['cycle_dir']: Experiment cycle directory
+    **r2d2_dict['logger']: Swell logger
+    (specified this way for multiprocessing)
 
     These values will be popped from the dictionary before running the fetch command
     """
@@ -70,13 +72,13 @@ def run_r2d2_fetch(r2d2_dict: dict) -> None:
                     window_length='PT6H',
                     target_file=empty_obs_file,
                 )
-            
+
             # Copy the empty file to the target file directory
             shutil.copy(empty_obs_file, target_file)
-        
+
         else:
             raise Exception(e)
-    
+
     # Change the permissions
     os.chmod(target_file, 0o644)
 
@@ -201,7 +203,7 @@ class GetObservations(taskBase):
         # Create a dictionary of all fetch criteria
         # -----------------------------------------
         r2d2_fetch_dicts = []
-        
+
         # Dictionary tracking all observation files
         # -----------------------------------------
         observation_dicts = {}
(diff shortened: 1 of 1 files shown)


src/swell/tasks/get_observations.py
```diff
@@ -307,20 +326,62 @@ def execute(self) -> None:
...
+                })
+
+        for fetch_dict in r2d2_fetch_dicts:
+            fetch_dict['logger'] = self.logger
+            fetch_dict['cycle_dir'] = self.cycle_dir()
+
+        # Run through all files to fetch
+        # ------------------------------
+        number_of_workers = 4
+        self.logger.info(f'Running parallel plot generation with {number_of_workers} workers')
```
mer-a-o: "Running parallel fetch observations"

mer-a-o (commented):
Changes look good to me. I only had one minor comment regarding this PR. Thanks @mranst 

I noticed that in Cylc get observation task is one task and if fetching one of the observations fails it's difficult to see it in the tui. I think it's also not possible to rerun only for failed observation. Having separate logs can also make things cleaner and easier for experiment with lots of observaitons. Would having a Cylc family of get observation with each observtion as sub-task under this family make more sense?

mranst:
> I noticed that in Cylc get observation task is one task and if fetching one of the observations fails it's difficult to see it in the tui. I think it's also not possible to rerun only for failed observation. Having separate logs can also make things cleaner and easier for experiment with lots of observaitons. Would having a Cylc family of get observation with each observtion as sub-task under this family make more sense?

That's an interesting concept - I can try this out. The only thing I'm unsure about is the impact on execution time/IO

jeromebarre:
> > I noticed that in Cylc get observation task is one task and if fetching one of the observations fails it's difficult to see it in the tui. I think it's also not possible to rerun only for failed observation. Having separate logs can also make things cleaner and easier for experiment with lots of observaitons. Would having a Cylc family of get observation with each observtion as sub-task under this family make more sense?
> 
> That's an interesting concept - I can try this out. The only thing I'm unsure about is the impact on execution time/IO

To make the parallel with Skylab/Ewok, this is done like so over there to create separate tasks under the getObservation family:
https://github.com/JCSDA-internal/ewok/blob/develop/src/ewok/tasks/getObs.py
and the number of obs fetch tasks running at the same time are controled by:
`maxlogintasks` which can be very debatable... 
https://github.com/JCSDA-internal/ewok/blob/04c875303e8e58d4cc14480bd46c80e15f5ec962/src/ewok/workflows/ecflow/ecflow.py#L66
or
https://github.com/JCSDA-internal/ewok/blob/04c875303e8e58d4cc14480bd46c80e15f5ec962/src/ewok/workflows/cylc/cylc.py#L220

ftgoktas:
I tested GetObservations on `3dfgat_atmos` with three modes (timed from `CYLC_JOB_INIT_TIME` to `CYLC_JOB_EXIT_TIME`):

`develop` (sequential): 275s
`multiprocessing.Pool` (4 procs): 137s
`ThreadPoolExecutor` (10 threads): 135s - tested in [feature/getobs-threading](https://github.com/GEOS-ESM/swell/tree/feature/getobs-threading)

So both get around 2x speedup, and threading vs processes is essentially takes same amount of time.

Although I’d recommend using threads instead of processes here. `r2d2.fetch()` is mostly I/O bound since it's mainly HTTP requests and file transfer. Threads handle this well and avoid extra process overhead. We'd also avoid passing objects between processes, and we currently pass `self.logger` to worker arguments.

mranst:
Obsolete via #756

PR #724, opened 2026-03-05
Separate GetObs into individual tasks per observation

mranst:
This is an alternative to #721, we can look at merging either one or the other depending on which makes the most sense. Looping through observations has been removed in favor of fetching each observation as its own task

Pros:
Much simpler to see which observations are failing to fetch and why
Can re-trigger failed tasks easily

Cons:
May complicate flow.cylc somewhat
Implementation may be a bit hacky

In order to limit the amount of concurrent tasks, I introduced a key for this cylc parameter that is limited to 10 by default


src/swell/suites/hofx/flow.cylc
```diff
@@ -131,8 +143,19 @@
     [[GetBackgroundGeosExperiment-{{model_component}} ]]
         script = "swell task GetBackgroundGeosExperiment $config -d $datetime -m {{model_component}}"
 
-    [[GetObservations-{{model_component}}]]
-        script = "swell task GetObservations $config -d $datetime -m {{model_component}}"
+    # ---------------------------------------------------------------------------------------------
+    # This has to be here so swell will successfully register the task
+    # [[GetObservations-{{model_component}}]]
+    #    script = "swell task GetObservations $config -d $datetime -m {{model_component}}"
+    # ---------------------------------------------------------------------------------------------
```
mer-a-o, line 151:
```suggestion
    # ---------------------------------------------------------------------------------------------
    
    [[GetObservations-{{model_component}}]]
```

src/swell/suites/hofx/flow.cylc
```diff
@@ -131,8 +143,19 @@
...
+
+    {% for observation in models[model_component]['observations'] %}
+    [[GetObservations-{{observation}}-{{model_component}}]]
```
mer-a-o, line 154:
```suggestion
    [[GetObservations-{{observation}}-{{model_component}}]]
        inherit = GetObservations-{{model_component}}
```

mer-a-o (commented):
I added two lines to `hofx_cf` suite to make `GetObservations` nested like the image below:

Image

It'll make the tui a little cleaner. Is this something that we can use for other suites too?

mranst:
> It'll make the tui a little cleaner. Is this something that we can use for other suites too?

Cool, I didn't know inherit would group tasks like that in the tui, thanks!

src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/iasi_metop-c.yaml
```diff
@@ -218,7 +218,7 @@ obs post filters:
     options:
       channels: *iasi_metop-c_channels
       error parameter vector: *iasi_metop-c_obserr
-      use_flag: &useflag_iasi_metop-c {{iasi_metopb_active_channels}}
+      use_flag: &useflag_iasi_metop-c {{iasi_metopc_active_channels}}
```
Dooruk, line 221: wow did you just catch this bug? This might be big if they are different between metops

mranst: Yep, this is an advantage to having these be separate tasks. Since previously the `JediRendering` object gets filled with keys from every single observation, it's easier for stuff like this to slip through. Under this method, it just fails

Dooruk: They are identical apparently..

src/swell/suites/3dvar/flow.cylc
```diff
@@ -125,8 +130,17 @@
     [[ GetBackground-{{model_component}} ]]
         script = "swell task GetBackground $config -d $datetime -m {{model_component}}"
 
-    [[GetObservations-{{model_component}}]]
-        script = "swell task GetObservations $config -d $datetime -m {{model_component}}"
+    # ---------------------------------------------------------------------------------------------
+    # This has to be here so swell will successfully register the task
+    # [[GetObservations-{{model_component}}]]
+    #    script = "swell task GetObservations $config -d $datetime -m {{model_component}}"
+    # ---------------------------------------------------------------------------------------------
```
Dooruk: Is this commented out section on purpose?

src/swell/suites/3dvar_atmos/flow.cylc
```diff
@@ -149,9 +161,20 @@
     [[GetBackgroundGeosExperiment-{{model_component}} ]]
         script = "swell task GetBackgroundGeosExperiment $config -d $datetime -m {{model_component}}"
 
+    # ---------------------------------------------------------------------------------------------
+
```
Dooruk, line 165: ditto

Dooruk (commented):
I really like the ability to see individual observations, it would tremendously help in a monitoring setting. It's better than #721 in that regard.

One downside is now we would have 30 log files instead of 1 for each GetObs task, which is an IOnode issue over multiple timesteps. But we could look into some other solution for that. I agree it complicates flow.cylc but not too bad IMO.

Does it help with the execution time? I can test it when I'm not abusing R2D2 API calls but maybe you already made comparisons?

src/swell/suites/3dvar/flow.cylc
mranst: Nope, thanks 👍

mranst:
> I really like the ability to see individual observations, it would tremendously help in a monitoring setting. It's better than #721 in that regard.
> 
> One downside is now we would have 30 log files instead of 1 for each GetObs task, which is an IOnode issue over multiple timesteps. But we could look into some other solution for that. I agree it complicates flow.cylc but not too bad IMO.
> 
> Does it help with the execution time? I can test it when I'm not abusing R2D2 API calls but maybe you already made comparisons?

IOnodes, always a concern... I'm not sure there's a good solution to this, since having each have their separate log file is part of the appeal of doing it like this

It's a little difficult to gauge because there's not the clean time readout at the end, but from what I can tell from the scheduling log it takes over twice as long as #721 🙃

mranst:
Obsolete via #756

PR #725, opened 2026-03-06
Use IASI metop-c active channels

mranst:
Discovered a bug in the metop-c IASI obs file while working on #724


Dooruk (approved):
Not sure if this will change reference results for Tier2.

mranst:
I ran a tier2 config for 3dfgat_atmos and saw no residual norm diff

gmao-wgu:
It should have no difference.

PR #734, opened 2026-03-11
Enable Swell to use GMAO AWS R2D2 server and S3 data store

### LINKED ISSUE #733: R2D2 GMAO Server
Swell currently uses JCSDA's R2D2 server and Discover as a local storage. We need to have the option to direct Swell to use GMAO's AWS R2D2 server and store observations in S3.

ftgoktas:
## Description

AWS R2D2 server is ready and configured for GMAO users. Users and API keys can be created, and Swell can use this server to ingest and fetch observations via S3 instead of the JCSDA server and Discover local storage.

See the [configuration guide](https://github.com/GEOS-ESM/swell/blob/feature/aws-datastore/docs/configuring_aws_server.md) for setup steps, example configs and IngestObs test commands.

Closes #733


docs/configuring_aws_server.md
```diff
@@ -0,0 +1,173 @@
...
+```bash
+# Create the experiment
+swell create ingest_obs_marine
+
+# Edit the generated experiment.yaml:
+#   - dry_run: false
+#   - obs_to_ingest: ['adt_cryosat2n']
+
+# Run the task
+swell task IngestObs /discover/nobackup/fgoktas/SwellExperiments/swell-ingest_obs/swell-ingest_obs-suite/experiment.yaml \
```
mer-a-o: Just a suggestion to show a generic path like:
`<experiment_root>/swell-ingest_obs/swell-ingest_obs-suite/experiment.yaml`

docs/configuring_aws_server.md
```diff
@@ -0,0 +1,173 @@
...
+  -d 2021-07-02T06:00:00Z \
+  -m geos_marine
+```
+
+## 2. Verify it stored
```
mer-a-o:
```suggestion
## 2. Verify it is stored
```

docs/configuring_aws_server.md
```diff
@@ -0,0 +1,173 @@
...
+
+```python
+python3 -c "
+import r2d2
+results = r2d2.search(
+    item='observation',
+    observation_type='adt_cryosat2n'
```
mer-a-o: This will find all `adt_cryosat2n` observations. You can add date information to narrow it down. 


Do you know what will happen if an observation is saved twice by the same (or different) user(s)? Will that be overwritten or will there be another instance of this observation with a different create date or modification date?

docs/configuring_aws_server.md
```diff
@@ -0,0 +1,173 @@
...
+#### a. Set up your environment
+
+Copy or create the required scripts into your working directory:
+- [load_r2d2.sh](../src/swell/utilities/scripts/load_r2d2.sh)
+- [prod_setup_env.sh](../src/swell/utilities/scripts/prod_setup_env.sh)
+
+Then source them to load R2D2 and set environment variables:
+
+```bash
+source load_r2d2.sh
```
mer-a-o: what is this lind in `load_r2d2.sh`?

```
# load spack-stack modules
mod_swell
```

src/swell/utilities/scripts/prod_setup_env.sh
```diff
@@ -3,15 +3,16 @@ unset R2D2_SERVER_PORT
...
+export R2D2_HOST="discover"
+export R2D2_COMPILER="intel"
+export R2D2_SERVER_HOST="http://13.217.72.149"
+export R2D2_SERVER_PORT="8080"
 
 source venv_client/bin/activate
 
 echo “  R2D2 Production environment:”
 echo “  R2D2_API_KEY:  [set]”
 echo “  R2D2_SERVER_HOST: $R2D2_SERVER_HOST '(should be empty)'”
```
mer-a-o: You are setting r2d2_Server host and port earlier and this wouldn't be empy?

src/swell/utilities/scripts/prod_setup_env.sh
```diff
@@ -3,15 +3,16 @@ unset R2D2_SERVER_PORT
...
 export R2D2_USER=username
 export R2D2_API_KEY=api_key
-export R2D2_HOST=discover-gmao
-export R2D2_COMPILER=intel
+export R2D2_HOST="discover"
+export R2D2_COMPILER="intel"
+export R2D2_SERVER_HOST="http://13.217.72.149"
+export R2D2_SERVER_PORT="8080"
 
 source venv_client/bin/activate
```
mer-a-o: What does this line do?

ghost: This activates the corresponding Python virtual environment.

src/swell/utilities/r2d2.py
```diff
@@ -190,21 +135,37 @@ def load_r2d2_credentials(
     if 'api_key' in credentials and 'R2D2_API_KEY' not in os.environ:
         os.environ['R2D2_API_KEY'] = credentials['api_key']
 
+    if 'r2d2_server_host' in credentials and 'R2D2_HOST' not in os.environ:
+        os.environ['R2D2_SERVER_HOST'] = credentials['r2d2_server_host']
```
ghost: This is the only block where the conditional -- `R2D2_HOST` -- doesn't match the `os.environ` block below --- `R2D2_SERVER_HOST`.

If that's intentional, it may be worth a quick comment explaining why.

src/swell/utilities/scripts/prod_setup_env.sh
mer-a-o: right, but not all users have this environment name in this location.

docs/configuring_aws_server.md
ftgoktas: It initializes the `spack-stack` on Discover so that the subsequent `module load r2d2-client/...` command works. Is this alias (`mod_swell`) defined for you? I assume it's not for most people, so I'll need to change it.

src/swell/utilities/scripts/prod_setup_env.sh
ftgoktas: True, we use ` R2D2_SERVER_HOST` and `R2D2_SERVER_PORT` if it's not the JCSDA's server. Fixing it.

src/swell/utilities/r2d2.py
ftgoktas: Good catch! Just fixed it. `R2D2_HOST` is the compute host identifier (e.g. discover) while `R2D2_SERVER_HOST` is the server URL. It should be checking `R2D2_SERVER_HOST`.

docs/configuring_aws_server.md
ftgoktas: Storing the same observation twice reuses the existing database record (no duplicate created). So, the new file overwrites the old one. If multiple users store the same observation type for the same window, the last write stays.

src/swell/utilities/scripts/prod_setup_env.sh
ftgoktas: Good catch -- that line is a leftover from my local dev setup. I'll remove it since `load_r2d2.sh` already handles loading the `r2d2-client` via the shared module.

ftgoktas:
The AWS R2D2 server is up and running from 8:30am EST to 5:30pm EST on weekdays. Lifetime values and model names have been registered in the database with the following configuration:

Lifetimes:
- debug: 14 days
- science: 180 days
- publication: 1825 days (5 years)
- release: indefinite

Models:
- geos
- mom6
- geos_cf
- mom6_cice6_UFS

Please follow the instructions in the [configuration guide](https://github.com/GEOS-ESM/swell/blob/feature/aws-datastore/docs/configuring_aws_server.md) to test against the server. Let us know if you run into any issues or have questions.

docs/configuring_aws_server.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Verify it is stored
+
+```python
+python3 -c "
+import r2d2
+results = r2d2.search(
+    item='observation',
+    observation_type='adt_cryosat2n',
+    window_start='20210702T060000Z',
```
mer-a-o: This should be `20230702T030000Z`

docs/configuring_aws_server.md
mer-a-o: Right, `mod_swell` is not defined for me.

src/swell/utilities/r2d2.py
```diff
@@ -180,21 +177,37 @@ def load_r2d2_credentials(
...
+    if 'aws_session_token' in credentials and 'AWS_SESSION_TOKEN' not in os.environ:
+        os.environ['AWS_SESSION_TOKEN'] = credentials['aws_session_token']
+
     # Set host and compiler (YAML config takes precedence over platform detection)
-    if 'host' in credentials and 'R2D2_HOST' not in os.environ:
-        os.environ['R2D2_HOST'] = credentials['host']
-        logger.info(f"Using platform host '{r2d2_host}' (overriding YAML '{credentials['host']}')")
+    if 'r2d2_host' in credentials and 'R2D2_HOST' not in os.environ:
+        os.environ['R2D2_HOST'] = credentials['r2d2_host']
+        logger.info(f"Using platform host '{r2d2_host}' \
```
jeromebarre: maybe this log info is a bit misleading.
should this say 'YAML is overriding platform defaults', to me it reads as the reverse

src/swell/utilities/scripts/prod_setup_env.sh
```diff
@@ -3,15 +3,14 @@ unset R2D2_SERVER_PORT
 
 export R2D2_USER=username
 export R2D2_API_KEY=api_key
-export R2D2_HOST=discover-gmao
-export R2D2_COMPILER=intel
-
-source venv_client/bin/activate
+export R2D2_HOST="discover"
+export R2D2_COMPILER="intel"
+export R2D2_SERVER_HOST="http://13.217.72.149"
```
jeromebarre, line 8: Should we replace such hardcoded server IP with a hostname or config variable. Sould we avoid plaintext AWS keys unless that is truly expected for our operations? @shiklomanov-an ?

jeromebarre (commented):
Looks good to me but see my minor comments

src/swell/utilities/scripts/prod_setup_env.sh
ftgoktas: The IP is an Elastic IP so it won't change, but agreed a DNS hostname would be cleaner long-term. 

The server now supports two data stores: the local one on Discover (`geos_cf` directory with priority 1) and the S3 bucket (priority 2). For Discover, users only need `R2D2_USER` and `R2D2_API_KEY` and no AWS credentials. AWS credentials are only required when accessing the S3 data store.

docs/examples/r2d2/ingest_obs.md
mranst, line 1: Minor question, why is this changed to 2023? I thought 2021 is the default time for geos_marine

mranst (commented):
Sorry it took me a while to review this. I was able to ingest and fetch obs from AWS, but I found that having to swap out the credentials file made it difficult to switch between working on other things. I wonder if we could make use of the dictionary structure in the yaml file to allow for datastores to be selected using an experiment key. Maybe something like:

`r2d2_credentials.yaml`
```yaml
jcsda:
  user: ...
  api_key: ...
  ...

gmao_aws:
  user: ...
  api_key: ...
  ...
```

Swell could then select which of those credentials to use based on a key called `r2d2_datastore`, or something like that

docs/examples/r2d2/ingest_obs.md
ftgoktas: That makes more sense, changing it to this way. I updated the date to `2023` since looks like the marine obs files we wanted to ingest no longer exist for `2021`, and were causing errors when we tried to point to that directory and ingest them.

ftgoktas:
Tested on Discover with both JCSDA's and GMAO's R2D2 servers. Users can now select which R2D2 server and datastore to use directly from `experiment.yaml` via `r2d2_server` and `r2d2_datastore` fields, without needing to modify credentials or code between runs. It defaults to `gmao_server` and the `r2d2-experiments-prod-us-east-1` S3 datastore. Updated the docs with more details for testing: [Configuring different R2D2 servers and datastores with Swell on Discover](https://github.com/GEOS-ESM/swell/blob/feature/aws-datastore/docs/configuring_aws_server.md)

src/swell/utilities/question_defaults.py
```diff
@@ -161,6 +161,35 @@ class r2d2_experiment_id(SuiteQuestion):
 
     # --------------------------------------------------------------------------------------------------
 
+    @dataclass
+    class r2d2_server(SuiteQuestion):
+        default_value: str = "gmao_server"
```
mranst: Defaulting to None is better here, to maintain previous behavior and not force the user to create configurations to match exact names

```suggestion
        default_value: str | None = None
```

src/swell/utilities/r2d2.py
```diff
@@ -141,37 +138,74 @@ def _get_platform_r2d2_config(logger: Logger, platform: str = None) -> tuple:
...
         try:
             yaml = YAML(typ='safe')
             with open(yaml_path, 'r') as yaml_file:
-                credentials = yaml.load(yaml_file)
+                credentials_yaml = yaml.load(yaml_file) or {}
         except Exception as e:
             logger.error(f"Error loading R2D2 credentials from {yaml_path}: {e}")
             logger.info("Continuing with existing environment variables...")
-            credentials = {}
+            credentials_yaml = {}
```
mranst, line 179: Is there a common type of exception you're expecting here?

src/swell/utilities/r2d2.py
```diff
@@ -141,37 +138,74 @@ def _get_platform_r2d2_config(logger: Logger, platform: str = None) -> tuple:
...
     else:
         logger.info(f"R2D2 credentials file not found at {yaml_path}")
-        logger.info("R2D2 v3 will use existing environment variables if set")
+        logger.info("R2D2 will use existing environment variables if set")
+
+    if not isinstance(credentials_yaml, dict):
+        logger.error("R2D2 credentials file must contain a YAML mapping at the root.")
```
mranst: I would just raise a TypeError here. If the user has something set up incorrectly, it's best to let them know rather than move on

src/swell/utilities/r2d2.py
```diff
@@ -141,37 +138,74 @@ def _get_platform_r2d2_config(logger: Logger, platform: str = None) -> tuple:
...
+        credentials = {}
+    elif server_name:
```
mranst:
```suggestion
if server_name is not None:
```
Changing this to an if statement since this block of logic also depends on the object being a mapping. The `is not None` is more specific and readable, in my opinion

src/swell/utilities/r2d2.py
```diff
@@ -141,37 +138,74 @@ def _get_platform_r2d2_config(logger: Logger, platform: str = None) -> tuple:
...
+        server_credentials = credentials_yaml.get(server_name)
+        if isinstance(server_credentials, dict):
+            credentials = server_credentials
+            logger.info(f"Using R2D2 credentials for server: {server_name!r}")
+        else:
+            logger.error(
```
mranst: I would raise an exception here

src/swell/utilities/r2d2.py
```diff
@@ -180,22 +214,40 @@ def load_r2d2_credentials(
...
-        os.environ['R2D2_COMPILER'] = credentials['compiler']
-        logger.info(f"Using platform compiler '{r2d2_compiler}' \
-                    (overriding YAML '{credentials['compiler']}')")
-        logger.warning("Using compiler from YAML file")
+    if 'r2d2_compiler' in credentials and 'R2D2_COMPILER' not in os.environ:
+        os.environ['R2D2_COMPILER'] = credentials['r2d2_compiler']
+        logger.info(
+            f"YAML r2d2_compiler ({credentials['r2d2_compiler']!r}) overrides "
+            f"platform default ({r2d2_compiler!r})"
+        )
```
mranst, line 252: I can understand changing `host` to the more-specific `r2d2_host` now that we also have `r2d2_server_host` in the mix. My only concern is that this will require users to change this file once this PR goes in. Maybe we can consider accommodating both keys, keeping `host` and `compiler` as functional but deprecated? We can even throw a deprecation warning  if the old names are detected.

mranst (commented):
Thanks for accommodating my request, this is looking great. I have a few last points - the largest of which is maintaining continuity once this PR goes in. I'd like to keep the amount of changes that users need to make to their environment to run swell as low as possible. Configuring R2D2 is probably the biggest stumbling point users encounter when trying to run swell, so if possible I think it's best to try to maintain compatibility with the current format of the credentials file

src/swell/utilities/r2d2.py
ftgoktas: Done

src/swell/utilities/r2d2.py
ftgoktas: Done

src/swell/utilities/r2d2.py
ftgoktas: Changed it

src/swell/utilities/question_defaults.py
ftgoktas: Agreed, done

r2d2_credentials.yaml
```diff
@@ -1,13 +1,33 @@
...
 
-# R2D2 v3 API credentials
-user: your_username
-api_key: your_api_key_here
+Two formats are supported:
 
-# Platform configuration  
-host: discover-gmao
-compiler: intel
+1. Single server — put credentials at the root level:
```
mer-a-o: Some lines in this yaml file should be comments?

src/swell/utilities/question_defaults.py
```diff
@@ -161,6 +161,36 @@ class r2d2_experiment_id(SuiteQuestion):
...
+        default_value: str | None = None
```
mer-a-o, line 176: `default_value` is defined twice here

src/swell/utilities/r2d2.py
ftgoktas: Good point. `host` / `compiler` are now accepted as deprecated aliases, they still work but log a warning telling users to rename to `r2d2_host` / `r2d2_compiler`. The new key takes precedence if both are present.

src/swell/utilities/r2d2.py
ftgoktas: Narrowed this to `OSError` and `YAMLError`, those should be the two expected failure modes here.

src/swell/utilities/question_defaults.py
ftgoktas: Removed

src/swell/utilities/question_defaults.py
```diff
@@ -171,6 +171,35 @@ class r2d2_experiment_id(SuiteQuestion):
...
+            "Server/profile name in ~/.swell/r2d2_credentials.yaml "
+            "(e.g. 'gmao_server'). Leave empty if credentials are at the root level."
+        )
+        widget_type: WType = WType.STRING
+
+    # --------------------------------------------------------------------------------------------------
+
+    @dataclass
+    class r2d2_datastore(SuiteQuestion):
+        default_value: str = "r2d2-experiments-prod-us-east-1"
```
mranst: Should this be set to None as well? This was causing me issues when trying to fetch from jcsda

ftgoktas: Changed to `None`, it should be working now

mranst (approved):
Looks great! 👍

Dooruk (approved):
Works for me 👍🏼 

One star for backwards R2D2 config compatibility + two stars for the detailed documentation
⭐ ⭐ ⭐

jeromebarre (approved):

Issue #738, opened 2026-03-16
Make CODEOWNERS more specific?


mathomp4:
The current `CODEOWNERS` file is very sparse. Just a single line for the `.github/workflows` directory.

But maybe in some cases, swell should have more code owners now? Like say `src/swell/suites/localensembleda` should always notify @metdyn ? 

Just a thought.

cc: @mranst

PR #739, opened 2026-03-16
Update CODEOWNERS for workflows

mathomp4:
## Description

In asking #738 I noticed that `CODEOWNERS` had the old id for @shiklomanov-an. So, well, we update it. (Unless @shiklomanov-an thinks he should be removed?)

## Dependencies

None

## Impact

None


shiklomanov-an (approved):
Approved!

PR #743, opened 2026-03-18
Fix variational3D oops configuration

mranst:
Sorry, this is on me for not running the tier1 tests before merging. The addition to the `variational3D` oops configuration in #716 breaks the suites for other models. This is a band-aid fix that avoids having to revert the PR adding `3dvar_cf` and restoring functionality of all suites. This may not be an ideal solution but we can amend it later

I've tested this works for tier1 suites + `3dvar_cf`


mer-a-o (approved):
Thanks @mranst. Sorry, I should have checked the other models before changing the yaml template.

PR #746, opened 2026-03-20
Adding background error options in 3dvar_cf

jeromebarre:
## Description

This PR adds configurable SABER background-error support for geos_cf by plumbing  `saber_central_block`  and `saber_outer_block` through Swell question defaults, task question registration, JEDI template rendering, and GEOS-CF interface generation. It updates the background error configuration to support both `identity` and `bump_nicas` (and later other flavors such as `diffusion`) central blocks, with optional outer blocks such as `stddev_bkg_scaled` or `stddev_fixed_values` (for testing purposes), and updates cycle staging so NICAS files are only linked when `bump_nicas` is actually selected. 

## Dependencies

None

## Impact

3dvar_cf suite


src/swell/configuration/jedi/oops/variational3D.py
```diff
@@ -52,7 +52,12 @@ def render_oops(self):
                 },
                 'prints': {
                     'frequency': 'PT3H'
-                }
+                },
+                'increment': {
+                    'geometry': self.interface_model('geometry'),
+                    'output': self.interface_model('increment_cs'),
+                },
+
```
mranst: Can you make sure the fix from #743 is incorporated?

mranst:
Looks good, I was able to run the suite with these changes. I wanted to point out there's a shift in the residual norms from the develop version to this PR, not sure if that's important

This branch -> develop
image

src/swell/configuration/jedi/oops/variational3D.py
mer-a-o: I'm testing this PR right now, and it looks like it's already incorporated in line 65-68. So there's no need to make any changes to this file in this PR.

mer-a-o (commented):
I tested this PR with different options for background error and it's working. Thanks @jeromebarre. My only comment is in `src/swell/configuration/jedi/oops/variational3D.py`.

jeromebarre:
> Looks good, I was able to run the suite with these changes. I wanted to point out there's a shift in the residual norms from the develop version to this PR, not sure if that's important
> 
> This branch -> develop image

Yes this is expected with those changes.

mranst (approved):
👍

mer-a-o (approved):

Issue #748, opened 2026-03-21
An indentation error in jedi_ensmeanvariance_config.yaml


metdyn:
In develop branch, the  jedi_ensmeanvariance_config.yaml file contains an indentation error for keyword  `pattern:. `  etc. 

```
ensemble:
  members from template:
    template:
      datetime: '2023-10-10T00:00:00Z'
      filetype: cube sphere history
...

  pattern: '%mem%'
  nmembers: 2
  zero padding: 4
```

The correct lineup should have 4 spaces.

mranst:
Closed by #749

PR #749, opened 2026-03-21
minor yaml corrections related to computing ensemble prior mean and variance 

metdyn:
## Description

solves the issue: 
- https://github.com/GEOS-ESM/swell/issues/748

## Impact

fv3jedi_ensmeanvariance.x output:
geos.prior.mean.20231010_000000z.nc4
geos.prior.variance.20231010_000000z.nc4

cov Ps

fv3jedi_letkf.x output:
geos.analysis.mean.20231010_000000z.nc4
geos.mean-inc.20231010_000000z.nc4


mer-a-o (commented):
Changes look good to me. Thanks @metdyn. I just have a question and a minor suggestion. 

Which suite uses `fv3jedi_ensmeanvariance.x`?

Since calculation of ensemble mean and spread is not limited to prior and can be done for posterior too, would it make sense to change `ensemble_latlon_prior_mean_output` to `ensemble_latlon_mean_output` (same for variance)?

mranst (approved):
Looks good 👍

PR #750, opened 2026-03-24
Add comparisons for 3dvar_cf

mranst:
## Description

This PR enables comparison tests for 3dvar_cf, same as the other variational suites in swell. 

To create a comparison suite, create an override yaml containing at least the following entry:
```yaml
comparison_experiment_paths:
 - /path/to/3dvar_cf_exp1/experiment.yaml
 - /path/to/3dvar_cf_exp2/experiment.yaml
```
(these can be the same paths for testing purposes)

then launch `swell create compare_variational_cf -o <override.yaml>`

The comparison suite checks the jedi log to ensure the `Residual norms` are identical between the two experiments. If they are not, the suite will run eva tasks generating plots of comparisons/differences. If you don't have a test case with differing residual norms, you can check out the plots by removing the fail condition for those tasks from the `flow.cylc` graph


jeromebarre (approved):
looks good to me

Issue #736, opened 2026-03-16
New suite of on-prem code tests


mranst:
As discussed in the previous Swell development meeting, it would be beneficial to have a set of code tests that are more involved that the current set that are run on commit to pr, but lighter than the tier1 tests. These "Tier 0.5" tests could be run as another requirement for merge to swell, without needing the 40+ minutes that it currently takes to run the full tier1 tests.

We need a set of concepts that would reasonably assess swell's continuing performance without needing to run a full suite to completion.
My current ideas are:

- Suite creation test: This would simply run `swell create` for all suites, to make sure that they are able to be constructed. I know this would be helpful for me, at least, as I test changing configuration methods.
- JEDI config test. This would run `RenderJediObservations` and generate the jedi config file, without checking for the existence of observation files. This file can then be checked against a reference to make sure the files are able to be generated correctly. This should be a reasonably lightweight test, as it is just generating the yaml file without fetching obs or interfacing directly with jedi

Alexey and I have discussed adding code tests for every individual task within swell, to make sure they continue to function correctly. It's difficult to come up with tests for all, so let me know if there's any suggestions!

shiklomanov-an:
One conceptual suggestion: At a high level, Swell tasks break down into:
1. Write configuration files for other tools
2. Prepare simulations (e.g., copy/move files) for other tools
3. Execute other tools using their configuration files
4. Post-process the outputs of other tool results

In theory, as long as tasks that do 1 and 2 do not change, 3 and 4 should be the same (or at least, changes there are not Swell's fault; e.g., results of changes outside Swell). These steps --- creating config files and copying/moving files around --- should also be relatively fast.

So my suggestion would be to target "Tier 0.5" at precisely these kinds of tasks --- writing config files and preparing simulations. A good structure for a typical Tier 0.5 test might be something like:

1. Run Tier 1+ tests to see exactly what config files are produced. Save a copy of the relevant config files _as test fixtures in Swell_.
2. Tier 0.5 tests would write configs _but would not execute the big tool suites_ (`--dry-run` or `--write-configs-only` or whatever flag we use). Instead, the tests would only check that the _config_ files are identical, on the presumption that, if the inputs haven't changed, the outputs shouldn't change either.

PR #751, opened 2026-03-26
Add suite creation and jedi config render tests

mranst:
This PR adds two code tests discussed in #736. The first simply creates all suites to ensure they are configured correctly, and do not contain `defer_to_platform` or `defer_to_model`.

The second test renders the jedi config yaml for all JEDI suites and evaluates differences to a set of stored files. These are constructed in a `dry-run` mode, where observations are not fetched and all filepath prefixes are replaced by placeholders. The idea behind this test is to ensure all configs are able to be successfully constructed, and also can be used to evaluate changes to jedi configs, as any change will be reflected in the file diff as part of a PR. Any change to a jedi yaml is expected to trigger a failure in this test. To make it easy to account for changes, I have introduced a script that can be run using `swell utility CreateMockConfigs` that will automatically regenerate all of the configs used for comparison in the source code. 

I have implemented these tests as part of the regular code testing, as they add only about 20 seconds to the runtime of the code tests, but we can potentially look at implementing them in a different way, such as part of a new workflow


src/swell/deployment/create_experiment.py
```diff
@@ -249,7 +249,12 @@ def create_experiment_directory(
 
     # Read override file
     # ------------------
-    override_dict = read_override_file(override)
+    if isinstance(override, str):
+        override_dict = read_override_file(override)
+    elif override is None:
+        override_dict = {}
+    else:
+        override_dict = override
```
ghost: Minor thing, but my suggestion might be tighten this interface by moving the reading of the override file all the way out to the top-level in `swell.py` and make this always take a `dict` as an argument.

I.e., Here, it would just become: `def create_experiment_directory(..., override: dict, ...)`

And all of the `if isinstance ... elif override is None ... ` + `read_override_file` stuff would live in `swell.py` in the `create` function.

src/swell/tasks/run_jedi_variational_executable.py
```diff
@@ -96,6 +96,14 @@ def execute(self) -> None:
         self.jedi_rendering.add_key('crtm_coeff_dir', self.config.crtm_coeff_dir(None))
         self.jedi_rendering.add_key('window_begin', window_begin)
 
+        # Add placeholder names if mock experiment
+        # ----------------------------------------
+        if self.config.mock_experiment(False):
+            print('MOCK EXPERIMENT')
```
ghost: Stray print statement.

docs/code_tests/code_tests.md
```diff
@@ -17,3 +17,11 @@ By default, swell will create several directories in the working directory that
 ```yaml
 test_cache_location: /discover/nobackup/<user>/swell-test-cache
 ```
+
+## Code tests
+
+### Suite creation test
+The suite creation test attempts to construct experiments for all suites within swell in a temporary directory. If one fails, try creating the suite on its own to make sure it is configured properly. Ensure all values are valid and are not filled by the templates `defer_to_model` or `defer_to_platform`.
```
mer-a-o, line 24: I noticed that the temporary directory is in a strange location. Can it be set in `~/.swell/swell-test.yaml` similar to how `test_root` or `test_cache_location` are set?

src/swell/test/code_tests/jedi_config_test.py
```diff
@@ -0,0 +1,65 @@
...
+                                             f'jedi_{suite}_config.yaml')
+            with open(comparison_config, 'r') as f:
+                comparison_dict = yaml.load(f)
+
+            if config_dict != comparison_dict:
+                raise AssertionError(f'Rendered JEDI config for suite {suite}, {config_file} '
+                                     f'did not match comparison version {comparison_config}. '
+                                     'Please check the file diffs. If these differences '
+                                     'are intentional, new mock test files can be generated using '
+                                     '`swell utility MockJediConfigs`')
```
mer-a-o: This should be `swell utility CreateMockConfigs`

docs/code_tests/code_tests.md
```diff
@@ -17,3 +17,11 @@ By default, swell will create several directories in the working directory that
...
+
+### JEDI Config test
+The JEDI config test generates mock configs for jedi executables in a dry-run mode, where obs will not be checked and placeholders will be used for experiment filepaths. These configs are compared against reference files located in `src/swell/test/jedi_configs/`, and named `jedi_<suite>_config.yaml`. Any difference in values in these yamls will cause this test to fail, so ensure any differences created are intentional, then run `swell utility CreateMockConfigs` to automatically generate new reference files for all suites. These new files are placed in the `jedi_config` location in the source code.
```
mer-a-o: I noticed that chaning something like `start_cycle_point` or `final_cycle_point` in `suite_config.py` doesn't result in failure. So perhaps it's worth mentioning it here.

src/swell/test/code_tests/code_tests.py
```diff
@@ -52,6 +54,12 @@ def code_tests() -> None:
     # Load Pinned Versions Test
     test_suite.addTests(unittest.TestLoader().loadTestsFromTestCase(PinnedVersionsTest))
 
+    # Load Suite Creation Test
+    test_suite.addTests(unittest.TestLoader().loadTestsFromTestCase(SuiteCreationTest))
+
+    # Load Suite Creation Test
```
mer-a-o: Load JEDI config test

src/swell/utilities/scripts/create_mock_configs.py
```diff
@@ -0,0 +1,71 @@
...
+from swell.swell_path import get_swell_path
+from swell.utilities.mock_jedi_config import mock_jedi_config
+
+# --------------------------------------------------------------------------------------------------
+
+defaults_dict = {}
+
+marine_default_datetime = '20210701T120000Z'
+atmosphere_default_datetime = '20231010T000000Z'
+compo_default_datetime = '20230805T1800Z'
```
mer-a-o, line 22: Why are these dates hard coded?

src/swell/utilities/scripts/create_mock_configs.py
```diff
@@ -0,0 +1,71 @@
...
+
+defaults_dict['3dvar_marine'] = {'datetime': marine_default_datetime,
+                                 'model': 'geos_marine',
+                                 'executable_type': 'variational'}
+
+defaults_dict['3dvar_marine_cycle'] = defaults_dict['3dvar_marine'].copy()
+
+defaults_dict['3dfgat_marine_cycle'] = {'datetime': marine_default_datetime,
+                                        'model': 'geos_marine',
+                                        'executable_type': 'fgat'}
```
mer-a-o, line 32: Just a small detail, why `executable_type` for `3dfgat_marine_cycle` and `3dfgat_atmos` are different?

pyproject.toml
```diff
@@ -45,7 +45,8 @@ dependencies = [
     "h5py>=3.7.0",
     "flake8==6.1.0",
     "netCDF4",
-    "ruamel.yaml==0.17.16"
+    "ruamel.yaml==0.17.16",
+    "xarray==2024.7.0"
```
mer-a-o, line 49: Is this intentional?

docs/code_tests/code_tests.md
mranst: Thanks, added a note for this. I didn't realize this as I was writing this test, but changing the cycle times does not have an effect on the configs, since the cycle directory is replaced with a placeholder and swell is not checking for obs

pyproject.toml
mranst: It's necessary for the code tests to run on Github

src/swell/utilities/scripts/create_mock_configs.py
mranst: This is correct to Swell, but I'm not sure why they run different executables

src/swell/utilities/scripts/create_mock_configs.py
mranst: Normally the individual cycling points are calculated by the Cylc scheduler, which I am bypassing here. These aren't too important here because these mock configs don't change with the cycling point as long as the two cases match, but I could potentially look into re-creating this calculation in swell

mer-a-o:
@mranst I'm testing this branch, I'm getting this:
`SwellCreateExperiment: Creating experiment: 'swell-ufo_testing' in '/gpfsm/dnb33/tdirs/login/discover36.174920.mabdiosk/tmpeqil1jlx'` in the output. I also see directories and files generated in `test_cache_location: /discover/nobackup/mabdiosk/SwellExperiments/swell-test-cache` which is iset in my `~/.swell/swell-test.yaml`

mranst:
> @mranst I'm testing this branch, I'm getting this: `SwellCreateExperiment: Creating experiment: 'swell-ufo_testing' in '/gpfsm/dnb33/tdirs/login/discover36.174920.mabdiosk/tmpeqil1jlx'` in the output. I also see directories and files generated in `test_cache_location: /discover/nobackup/mabdiosk/SwellExperiments/swell-test-cache` which is iset in my `~/.swell/swell-test.yaml`

Thanks, they should both now go to the test cache

mer-a-o (dismissed):
It's working correctly now. Thanks @mranst !

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
```diff
@@ -330,6 +330,12 @@ produce_geovals:
   - true
   - false
 
+saber_central_block:
```
mer-a-o, line 333: Is the addition of `saber_central_block` is related to this PR?

src/swell/test/jedi_configs/jedi_3dvar_cf_config.yaml
```diff
@@ -45,7 +45,29 @@ cost function:
   background error:
     covariance model: SABER
     saber central block:
-      saber block name: ID
+      saber block name: BUMP_NICAS
```
mer-a-o, line 48: Probably these diffs will go away when you update the branch with develop?

mer-a-o (approved):

src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
mranst: Yes, should've made a note about this. This key is added to other models because before it would've rendered as 'defer_to_model', which would cause this test to fail

src/swell/test/jedi_configs/jedi_3dvar_cf_config.yaml
mranst: It's odd that this rendered as a diff, since this file is new to develop, but this file did have to be updated when #746 went in, as it changed the config file

PR #752, opened 2026-03-26
Adding cycling to 3dvar_cf suite

mer-a-o:
## Description

This PR adds a new cycling 3DVar suite for GEOS-CF. The experiment is designed to run a 12h forecast starting from the beginning of the assimilation window and apply the increment from the middle of the window using IAU. The forecast length and frequency can be set using `forecast_length` and `forecast_output_frequency` but I didn't test changing the forecast length. 

Summary of changes:

- RC files that get templated with SWELL are in `src/swell/configuration/jedi/interfaces/geos_cf/namelists`. The remaining RC files (static) are copied into the scratch directory from geos_cf_run_dir.

- `get_background` task fetches backgrounds from R2D2 using `background_experiment` in the config as the experiment ID. In cycling experiments (those with "cycle" in their name), the experiment ID is set to `background_experiment` for the first cycle. For subsequent cycles, it is set to the `r2d2_experiment_id `of the current experiment, so that backgrounds from the previous cycle are fetched for the variational task in the current one.

-  `save_forecast` task stores forecasts in R2D2 using r`2d2_experiment_id` as the experiment ID. These files are then fetched from R2D2 during the `get_background` task.

- Restarts are saved (`save_restart`) and fetched (`get_restart`) from R2D2. To save space, restart files are not stored on R2D2 at every cycle. The `rst_store_interval` key controls how many cycles pass before restart files are stored as real files rather than symlinks. In intermediate cycles, restarts are saved as symlinks.

- prep_forecast prepares the scratch directory within the current cycle's run directory:
  - Copies static RC files from `geos_cf_run_dir` and copies/edits RC files from `src/swell/configuration/jedi/interfaces/geos_cf/namelists`.
  - Copies GEOS-FP files for replay into the scratch directory.
  - Changes the format of the JEDI increment file using `inc_template`.

- `run_forecast` submits gcm_run_geoscf.j (a lighter version of `gcm_run.j `used for running GEOS-CF) to the queue and waits until the job finishes. One limitation: the Cylc interface shows this job as "running" whether it is waiting in the queue or actually executing. There may be better approaches for handling `gcm_run.j `execution.

- `clean_cycle.py` now cleans the scratch directory of the previous cycle after the current cycle completes. This way, the restart files from the previous cycle that are saved as symlinks are kept until the current cycle uses them.


src/swell/configuration/jedi/interfaces/geos_cf/namelists/cap_restart
rtodling, line 1: this file should not exist in the repo - it should only exist in the experiment. A way to avoid this file being in the repo is to have swell create this file based on the initial date/time of the experiment. From there on, the experiment (GCM) will create this file on its own; the file can then be recycled from one cycle to the next.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/CAP_c360.rc
```diff
@@ -0,0 +1,34 @@
...
+ROOT_CF: AGCM.rc
+HIST_CF: HISTORY.rc
+
+# doesn't matter, cap_restart
+BEG_DATE:     20000101 000000
+# can cap your run
+END_DATE:     29990302 210000 
+
+# length of a sigle segment/job 1day
+JOB_SGMT:     >>SWELL_FC_LENGTH<<
```
rtodling: For consistency with existing conventions in GEOS can we make:
'>>SWELL_FC_LENGTH<< '
into:
'>>>SWELL_FC_LENGTH<<<'
?

I imagine there are a number of other places with similar templating notation. Can the same comment apply to all?

src/swell/configuration/jedi/interfaces/geos_cf/namelists/AGCM_c90.rc
```diff
@@ -0,0 +1,775 @@
...
+# Resolution of state passed to atmos analysis
+# --------------------------------------------
+ANA_GRIDNAME: PC90x540-DC
+ANA_LM: 72
+
+# Restart Record Parameters
+# -------------------------
+RECORD_FINAL:  NO
+RECORD_FREQUENCY: 000000       000000
+RECORD_REF_DATE: >>SWELL_FC6H_DATE<< 21220215
```
rtodling: here too - please convert doube gt  lt into triple gt / lt.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/HISTORY.rc
```diff
@@ -0,0 +1,141 @@
...
+
+PC360x181-DC.GRID_TYPE: LatLon
+PC360x181-DC.IM_WORLD: 360
+PC360x181-DC.JM_WORLD: 181
+PC360x181-DC.POLE: PC
+PC360x181-DC.DATELINE: DC
+PC360x181-DC.LM: 72
+
+
+  geoscf_jedi.template:  '%y4%m2%d2T%h2%n200Z.nc4',
```
rtodling: This looks odd to me. If I understand this correctly, the cycle you do for CF is no different than what we do for ATMOS - which is a 12 hour cycle in which the first 6 hours model is forced by the analysis (and updated with Compo-Ana), followed by a 6-hour free running model. The prognostic files (backgrounds to the next Compo-Ana) should come out of the last 6-hour part of the integration, but it looks to me what you have here is the backgrounds being written at some frequency throughout the 12-hour integration. I don't think that's what you want.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/HISTORY.rc
```diff
@@ -0,0 +1,141 @@
...
+  geoscf_jedi.archive:   '%c/Y%y4',
+  geoscf_jedi.format:    'CFIO',
+  geoscf_jedi.mode:      'instantaneous' ,
+  #geoscf_jedi.frequency:  010000,
+  geoscf_jedi.frequency:  >>SWELL_FC_OUTPUT_FREQ<<,
```
rtodling: We should talk about the naming conventions for these templated frequencies and such.

src/swell/tasks/get_background.py
```diff
@@ -38,12 +38,22 @@ def execute(self) -> None:
         bkg_steps = []
 
         # Parse config
-        background_experiment = self.config.background_experiment()
```
rtodling, line 47: I am trying to understand the reason for the changes here. Is this because the backgrounds for various cycles for the swell experiment can be located outside of r2d2?

src/swell/tasks/get_restart.py
```diff
@@ -0,0 +1,76 @@
...
+from swell.utilities.r2d2 import load_r2d2_credentials
+
+import isodate
+import os
+import r2d2
+
+# --------------------------------------------------------------------------------------------------
+
+r2d2_model_dict = {
+    'geos_cf': 'geos_cf',
```
rtodling: Given the name and location of this task, I assume this is a general task to hand restarts - but this seems wired to geos_cf ... am I misreading? 

It seems that restarts are being handled by r2d2 - is there a way to extend this so that restarts can also be in tar-ball ? Not asking for it to be implemented now, but I'm wondering. The idea of lose restarts is something the ADAS moved away quite sometime ago given the sheer number of restarts and what those do to inode count.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_AnaSettings_CO.rc
```diff
@@ -0,0 +1,19 @@
+SpeciesName: CO
```
rtodling, line 1: I puzzled as to why this RC and other RCs and GEOS yaml need to be here. These files should be in the tag of GEOS that supports CF, no? What am I missing?

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_AnaSettings_NO2.rc
```diff
@@ -0,0 +1,20 @@
+SpeciesName: NO2
```
rtodling, line 1: Same comment as above for RC files ...

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_ExtData.yaml
```diff
@@ -0,0 +1,2769 @@
+Samplings:
```
rtodling, line 1: Same comment as for the RC files - this should be in the GEOS tag and in the settings of your experiment.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/HISTORY.rc
jeromebarre: if we want to do FGAT with `model: PSEUDO` we might want to output at a certain frequency. Likely hourly.

src/swell/tasks/prep_forecast.py
```diff
@@ -0,0 +1,331 @@
...
+        # -------------------
+        self.parse_andate = self.cycle_time_dto()
+        self.parse_wbegin = self.da_window_params.window_begin(self.window_length, dto=True)
+        self.parse_wend = self.da_window_params.window_end_iso(self.window_length, dto=True)
+        self.parse_wlen = isodate.parse_duration(self.window_length)
+        self.parse_fclen = isodate.parse_duration(self.forecast_length)
+
+        # Determine analysis variables for GEOS-CF (NO2 and CO)
+        # -------------------------------------------------------
+        an_vars_long_tg = ['volume_mixing_ratio_of_no2', 'volume_mixing_ratio_of_co']
```
jeromebarre: I would add o3 as well.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_AnaSettings_O3.rc
```diff
@@ -0,0 +1,31 @@
+SpeciesName: O3
+Active: 1
+AnalysisFreq: 3
+AnalysisHour: 1
+AnalysisMinute: 30
+ForwardLooking: 1
+FileTemplate: /discover/nobackup/projects/gmao/yotc/pub/fp/das/Y%y4/M%m2/D%d2/GEOS.fp.asm.tavg3_3d_asm_Nv.%y4%m2%d2_%h2%n2.V01.nc4
```
jeromebarre: I would try to make this AnaSettings consistent with NO2 and CO for the file template so we don't overlap with FP

src/swell/configuration/jedi/interfaces/geos_cf/namelists/CAP_c360.rc
mer-a-o: Sure. I replaced double > to triple > throughout the code.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/cap_restart
mer-a-o: removing `cap_restart`. Note that `prep_forecast.py` handles writing the correct initial date/time for the experiment.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_AnaSettings_CO.rc
mer-a-o: For each cycle, `FileTemplate` is replaced to point to the JEDI increment file. I'm keeping all the RC files that are modified for each cycle in SWELL. The rest of RC files are copied from a tagged version of GEOS specified in `suite_config.py`: `geos_cf_run_dir('/discover/nobackup/mabdiosk/rundir/GCv14.0_GCMv1.17_c90_Skylab')`
I think this way we can easily keep track of what goes into the swell experiment.

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_AnaSettings_O3.rc
mer-a-o: done

src/swell/configuration/jedi/interfaces/geos_cf/namelists/HISTORY.rc
mer-a-o: @rtodling let's talk about it. Same thing you described is happening here too.

src/swell/tasks/get_background.py
mer-a-o: Please see the comment in the code. For the first cycle the backgrounds are fetched from R2D2 using the experiment_id specified by the user. For the cycles after that swell ('get_background.py') fetches the backgrounds files from the 12h forecast done in the previous cycles. Those forecast files from the previous cycle get updloaded to R2D2 with the current experiment id. In the following cycle this experiment id is used to fetch those forecast files and will be used as backgrounds for the current cycle.

src/swell/tasks/get_restart.py
mer-a-o: This is a good idea. We can upload the tar balls to R2D2 instead of individual rst files to be mindfull of the ionode counts. I will open an issue to implement it in the next PR but if you think it's more appropiate to be done in this PR I can make the changes here.

src/swell/tasks/prep_forecast.py
mer-a-o: done

jeromebarre:
Should this be moved out of the draft status at this point? Ideally we want this ready for merge ASAP even if this is isn't perfect this is IMO good enough and replicates what has been done in slylab. 
There are fixes to be done on the coding norms and the swell CI. Which is something to clean in saveforecast.

Over time a few improvements on top of my head:
- getting met forcings for replay from the original locations and not from my dir in geos
- considering handling tarballs for restarts and not individual files
- GEOS-CF forecast logs written in the correct locations. But maybe this was fixed since last time I tested

src/swell/tasks/save_forecast.py
jeromebarre: the swell CI says it is unused 

```
  File "/home/runner/.local/lib/python3.11/site-packages/swell/test/code_tests/unused_variables_test.py", line 41, in test_unused_variables
    self.assertEqual(flake8_output, '', f"Unused variables found in {file_path}")
AssertionError: "/home/runner/.local/lib/python3.11/site-[249 chars]used" != ''
- /home/runner/.local/lib/python3.11/site-packages/swell/tasks/save_forecast.py:11:1: F401 'datetime.datetime as dt' imported but unused
- /home/runner/.local/lib/python3.11/site-packages/swell/tasks/save_forecast.py:18:1: F401 'swell.utilities.datetime_util.datetime_formats' imported but unused : Unused variables found in /home/runner/.local/lib/python3.11/site-packages/swell/tasks/save_forecast.py
```

src/swell/tasks/get_restart.py
mer-a-o: I opened an issue about this: https://github.com/GEOS-ESM/swell/issues/774

src/swell/tasks/geoscf-run.sh
mranst, line 1: Since this script and the RunForecast task are relatively simple, would it be possible to run the executable directly from the task runtime like how Doruk uses gcm_run.j? 

```
    [[RunForecast-{{model_component}}]]
        script = "{{experiment_path}}/run/$datetime/{{model_component}}/gcm_run_geoscf.j"
        platform = {{platform}}
        execution time limit = {{scheduling["RunForecast"]["execution_time_limit"]}}
        [[[directives]]]
            --output="{{experiment_path}}/run/$datetime/{{model_component}}/gcm_run.log"
        {%- for key, value in scheduling["RunForecast"]["directives"][model_component].items() %}
            --{{key}} = {{value}}
        {%- endfor %}
```

src/swell/tasks/geoscf-run.sh
```diff
@@ -0,0 +1,37 @@
...
+
+echo "SCRATCH_DIR = ${SCRATCH_DIR}"
+
+GCM_SCRIPT=${SCRATCH_DIR}/gcm_run_geoscf.j
+LOG_FILE=${SCRATCH_DIR}/gcm_run.log
+
+echo "Submitting GCM job: ${GCM_SCRIPT}"
+
+set +e
+sbatch --wait --output=${LOG_FILE} ${GCM_SCRIPT}
```
mranst: If I'm not mistaken the RunForecast task will be queued into slurm, then this will allocate its own job on top of that. This command also failed for me when I tried it

src/swell/tasks/get_background.py
```diff
@@ -38,12 +38,22 @@ def execute(self) -> None:
...
+        # for the first cycle, use background_experiment in config
+        # as the experiment id for fetching from r2d2 for cycles after
+        # the first, use the current experiment id for fetching from r2d2
+        if self.cycle_time_dto() != self.start_cycle_point_dto() and 'cycle' in self.suite_name():
+            background_experiment = self.config.r2d2_experiment_id()
+        else:
+            background_experiment = self.config.background_experiment()
+
+        self.logger.info(f'Fetching background from experiment {background_experiment}')
+
```
mranst, line 62: Will this affect `3dvar_marine_cycle` and `3dfgat_marine_cycle`?

jeromebarre (dismissed):
I just reran this one last time. It's all good with me. All the files are correctly there on r2d2.

One other missing piece is the save analysis files into r2d2. I created an issue months ago about this and never did it, so this one is one me.  

Another issue that need to be solved with R2D2 work with JCSDA will be to deregister the rst files that are actually not copied to the DS so this will keep the DB cleaner. But such functionality on deregistering single files are not there yet. The methods in R2D2 are missing so far. Tagging @jessemarks92 here from JCSDA for awareness.

An immediate cleaner option is to work with the tar files on r2d2. But wondering how timely this is to compress/decompress each cycle.


Thanks for this PR!

src/swell/tasks/geoscf-run.sh
mer-a-o: @mranst I tried what you suggested too. The issue is that gcm_run.j needs to be run with the command sbatch gcm_run.j. There are Slurm directives set in gcm_run.j: [see here](https://github.com/GEOS-ESM/swell/blob/cd3cf8f027daf8afbeef90a2f48c5126b827f377/src/swell/configuration/jedi/interfaces/geos_cf/namelists/gcm_run_geoscf_c90.j#L6-L11 )

So the main reason for the current design is to have gcm_run.j take care of the Slurm directives. One issue is that the job status will be set to "running" in cylc tui, even if the job is still waiting in the queue. However, the logs and errors are captured correctly. I'm open to suggestions for a better implementation.

In the case of marine, I wonder what would happen if we increased the number of requested nodes? In the marine gcm_run.j, the number of nodes is set to 1:

```
#SBATCH --time=00:18:00
#SBATCH --nodes=1 --ntasks-per-node=120
#SBATCH --job-name=coupled_5deg_l91_RUN
#SBATCH --constraint=mil
#SBATCH --qos=debug
#@BATCH_NAME -o gcm_run.o@RSTDATE
```

src/swell/tasks/geoscf-run.sh
mer-a-o: please see my previous comment. What failed exactly?

src/swell/tasks/get_background.py
mer-a-o: I was able to run `3dvar_marine_cycle` with this branch successfully. I will try `3dfgat_marine_cycle` too.

src/swell/tasks/geoscf-run.sh
mranst: Swell can set slurm directives depending on the task, and also gives the opportunity to set overrides. In my case, I don't have access to the compute group that's hard-coded into that file, so I'm not able to run the task - though it could be added dynamically to the sbatch command in run_forecast.py, I suppose.  I created a branch where I was able to execute `gcm_run_geoscf.j` directly with the correct slurm parameters if you're interested - `feature/mranst/3dvar_cf_cycle`

src/swell/tasks/geoscf-run.sh
mranst: Since this file is just running an sbatch command, there's no need for the `[[[directives]]]` section in flow.cylc. This is causing it to request a queue only for the task to request another queue for a different set of nodes. Mentioned above but it fails because I don't have access to the compute group

src/swell/configuration/jedi/interfaces/geos_cf/namelists/GEOSCHEMchem_AnaSettings_CO.rc
Dooruk: I have to repeat Ricardo's sentiment here regarding all these namelist files. I see `prep_forecast` edits these templated files but I suggest we keep GEOS relevant config files outside of Swell. There are many ways to do this:

1) Like Ricardo suggested, a particular GEOS-CF tag with SWELL/JEDI resource + namelist files. Especially since the GEOS versions of GEOS-CF versions get updated so infrequently. This could be resolution specific with c90 or c360 (or higher res going forward) and `gcm_setup` could create those directories with a SWELL/JEDI specific RC folder.
2) Creating a static/template GEOS-CF experiment directory on Discover to store these RC files. Could be platform specific going forward. They would live under `SwellStaticFiles`
3) (more involved, more GEOS version specific) Modify `prep_forecast.py` to handle these templates without having to store them in Swell. This is what I did before but I wouldn't recommend it.

src/swell/suites/3dvar_cf_cycle/flow.cylc
Dooruk, line 1: This is interesting. With my coupled cycling runs I run the forecast first then create the first analysis. i see this is the opposite. I'm not quite sure which one makes more sense in a DA cycling sense.

This approach requires backgrounds from the get go, my approach requires model restarts only.

src/swell/tasks/geoscf-run.sh
Dooruk: Looking at this `run_forecast` + `geoscf-run.sh` approach I also suggest my way of calling `gcm_run.j` might be easier (totally objective opinion 😄 ). 

When I run `marine_cycle_tier2`, I request 11 nodes fur RunGeos task using SLURM overrides. Currently, there is no mechanism in SWELL that forces the proper number of nodes are requested by `RunGeos` depending on the target `gcm_run.j` `#SBATCH` setup in the experiment directory.

src/swell/tasks/get_background.py
Dooruk: There is an obscure hidden call `self.start_cycle_point_dto` in base methods. For cycling VarBC I had to create an exception for the first cycle so that is what I introduced.

https://github.com/GEOS-ESM/swell/blob/dc8f6e95decc0e8bb1f50ec66caf5b3ef2927982/src/swell/tasks/get_observations.py#L262-L275

src/swell/tasks/get_restart_cf.py
Dooruk, line 1: I see Ricardo suggested using tarballs instead?

src/swell/tasks/prep_forecast_cf.py
Dooruk, line 1: if this is very CF specific, can we call it prep_cf_forecast?

The alternative is https://github.com/GEOS-ESM/swell/pull/698 organizing tasks under `geos_cf` `geos_marine` etc. but we never got around merging that..

src/swell/tasks/save_forecast.py
```diff
@@ -38,106 +36,48 @@ def execute(self) -> None:
...
+        forecast_length_dur = isodate.parse_duration(forecast_length)
+        forecast_frequency_dur = isodate.parse_duration(forecast_output_frequency)
+
+        step_dur = forecast_frequency_dur
+
+        while step_dur <= forecast_length_dur:
+            step = isodate.duration_isoformat(step_dur)
+
+            file_time = forecast_start + step_dur
+            fname = f"CF2.geoscf_jedi.{file_time.strftime('%Y%m%dT%H%M%SZ')}.nc4"
```
Dooruk: If this is CF specific let's rename the task, `save_cf_forecast`. I try to do this for my tasks as well and add `coupled` to their names. Or keep it under `geos_cf` after Michael's PR merges in. 

To respond to @jeromebarre's comments on https://github.com/GEOS-ESM/swell/pull/790 my goal was to make this task more generic between atmos, cf, and marine. It doesn't have to be that way ofcourse.

src/swell/tasks/save_restart.py
Dooruk, line 1: Please feel free to erase marine bits from this, I have to modify or add a new task for that. I do like the saving restart file intervals switch.

Dooruk (dismissed):
I'm very late to the party and I don't want to hold this back since this was introduced two months ago. I do have some comments that I would like to be addressed eventually though. Happy to discuss next Monday.

I tried running it but the queue is too long at the moment, got stuck at RunVar...

src/swell/tasks/get_background.py
mer-a-o: I ran `3dfgat_marine_cycle` successfully too.

src/swell/tasks/geoscf-run.sh
mer-a-o: I removed directives from flow.cylc for runForecast task.

src/swell/tasks/geoscf-run.sh
mer-a-o: My plan is to eventually remove `geoscf-run.sh`, so `run_forecast` + `geoscf_run.sh` is an intermediate step. :)

src/swell/tasks/get_background.py
mer-a-o: @Dooruk, so is it ok that I used `self.start_cycle_point_dto` here?

src/swell/suites/3dvar_cf_cycle/flow.cylc
mer-a-o: Since we save backgrounds more frequently than restarts, I think it makes more sense to start from backgrounds. Perhaps having the option to do either would be best.

src/swell/tasks/get_restart_cf.py
mer-a-o: Yes, I have an issue open about it: https://github.com/GEOS-ESM/swell/issues/774

src/swell/tasks/geoscf-run.sh
Dooruk: Sounds good. I don't want to force you into my way, objectively the **best way** 😄

src/swell/suites/3dvar_cf_cycle/flow.cylc
Dooruk: Yeah, different flow.cylc provides that flexibility. All good!

src/swell/tasks/get_background.py
Dooruk: Yeah, I don't see why not.

mer-a-o:
I made some changes to this branch to address some of the comments. @Dooruk and @mranst, could you review the changes? If everything looks good, we can merge this PR.

Here’s a summary of the updates:

* I changed the execution of the runForecast task. It is now called from flow.cylc, very similar to @Dooruk 's approach, except that gcm_run.j is in cycle_dir/scratch. @mranst , the Slurm account is now picked up automatically, so you should be able to run this task without any issues.

* I renamed the tasks that are more specific to CF:
```
PrepForecast     -> PrepForecastCf
GetRestart       -> GetRestartCf
SaveForecast     -> SaveForecastCf
SaveRestart      -> SaveRestartCf
```

* The forecast output name is now more centralized and set in geos-cf/r2d2.py. Both `HISTORY.rc` and `SaveForecastCf` now pick up the name from `r2d2.py`.

* Some small cleanups.

src/swell/suites/3dvar_cf_cycle/suite_config.py
```diff
@@ -0,0 +1,109 @@
...
+            # mom6_iau is available. I don't know if we want run 3dvar without iau
+            qd.iau(True),
+            qd.forecast_length('PT12H'),
+            qd.forecast_output_frequency('PT3H'),
+            qd.clean_patterns(['*.nc4', '*.txt', 'logfile.*.out']),
+            qd.inc_template(
+                '/discover/nobackup/mabdiosk/rundir/handle_inc/'
+                'GCC_c90_FPens.geoscf_jedi.20210805_0600z.nc4'
+            ),
+            qd.geos_cf_install_dir('/discover/nobackup/mabdiosk/GEOS-mil/GEOSgcm/install'),
```
Dooruk, line 95: Just a comment, I'm guessing these will be in generic folders eventually?

Dooruk (dismissed):
Thanks for addressing my comments. I will check running it but otherwise it is good to go in from my end 👍🏼

src/swell/suites/3dvar_cf_cycle/suite_config.py
mer-a-o: correct.

src/swell/suites/3dvar_cf_cycle/suite_config.py
mranst, line 1: I think you might need to add saber_central_block and saber_outer_block to the suite config now that they're no longer task questions

```File "/gpfsm/dhome/manstett/develop/src/swell/configuration/jedi/oops/variational3D.py", line 29, in render_oops
    'background error': self.interface_model('background_error'),
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dhome/manstett/develop/src/swell/utilities/oops_config.py", line 82, in interface_model
    config_value = self.jedi_rendering.render_interface_model(config_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dhome/manstett/develop/src/swell/utilities/render_jedi_interface_files.py", line 237, in render_interface_model
    config_value = config_func(self.__template_dict__)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfsm/dhome/manstett/develop/src/swell/configuration/jedi/interfaces/geos_cf/model/background_error.py", line 123, in background_error
    raise ValueError(
ValueError: Unknown background_error_model 'None'. Choose from: ['identity', 'bump_nicas']
2026-06-02T14:08:07Z CRITICAL - failed/ERR```

mranst (commented):
One small comment, otherwise looks good 👍

src/swell/suites/3dvar_cf_cycle/suite_config.py
mer-a-o: Thanks for catching it! I added the saber blocks.

src/swell/tasks/save_forecast.py
mer-a-o: done

mranst (approved):

Dooruk (approved):
one last approval before merging!

PR #756, opened 2026-04-02
Use threading for parallel GetObservations fetches and optional skip when files exist

ftgoktas:
Parallelizes R2D2 observation, bias, and tlapse fetches inside `GetObservations` using ThreadPoolExecutor. Fetch calls are mainly HTTP and file I/O, so threads allow several fetches run at the same time while each one waits on the network or disk.

Adds a `cache_fetch` task question (default on) that skips `r2d2.fetch` when the target path already exists and is non-empty, so rerunning a failed cycle does not repeat successful fetches.


mer-a-o (approved):
Changes look good. I tested with with `3dvar_cf` suite and it works correctly. Thanks @ftgoktas !

mer-a-o (dismissed):

mranst (dismissed):
Very nice! Quite a bit faster from what I can tell

mranst (approved):

PR #763, opened 2026-04-06
Fix pandas enforcing float dtype in GenerateObservingSystemRecords

mranst:
Pandas version 3 apparently started enforcing type based on the first entry in the data row. This trips up on a single line in `observing_system_records.py` that initializes as a float but is subsequently entered as a string. This isn't an issue in versions < 3, but pip has started preferring versions > 3 by default. I've tested that this is the only issue moving to pandas>3, so we shouldn't have to alter the requirements after this fix.

### DIFF
--- a/src/swell/deployment/platforms/nccs_discover_cascade/modules
+++ b/src/swell/deployment/platforms/nccs_discover_cascade/modules
@@ -41,7 +41,7 @@ module load r2d2-client/112025
 
 # Load eva and jedi_bundle
 # ------------------------
-module load eva/sles15_skylab9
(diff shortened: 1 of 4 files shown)


shiklomanov-an (dismissed):
Note that you may also have to fix this in `gsi_record_parser.py : update_return_df`.

mranst:
> Note that you may also have to fix this in `gsi_record_parser.py : update_return_df`.

Good catch, thanks!

rtodling (approved):
Works for me - as it was not working before these fixes.

mer-a-o (approved):

PR #765, opened 2026-04-09
Derive obs file extension from obsfile YAML

### LINKED ISSUE #723: GetObservations causing fetch failures for observations other than `.nc4`
`GetObservations` always uses `file_extension='nc4'` when fetching observations from R2D2. If observations were ingested with a different extension (like `.nc`), the fetch fails because the stored `file_extension` does not match.

We need to change hardcoded `nc4` in `get_observations.py` and instead determine `file_extension` before building the target path and calling `r2d2.fetch`.

ftgoktas:
Previously the observation file extension was hardcoded to `nc4` in both the fetch and combine phases. It is now derived from the obsfile template in each observation's JEDI YAML. 

- Built on top of `feature/r2d2-caching`

### DIFF
--- a/src/swell/tasks/get_observations.py
+++ b/src/swell/tasks/get_observations.py
@@ -356,7 +356,7 @@ class GetObservations(taskBase):
         # Run through all files to fetch
         # ------------------------------
         number_of_workers = 4
-        self.logger.info(f'Running parallel plot generation with {number_of_workers} workers')
+        self.logger.info(f'Fetching observations in parallel with {number_of_workers} workers')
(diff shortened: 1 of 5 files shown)


mer-a-o (approved):
Changes look good. Thanks @ftgoktas !

mranst (approved):
Looks good to me. Do you want to merge this into #756 first or wait?

ftgoktas:
Yes, let's merge this into #756 first and then merge #756 into develop. Thanks @mranst!

PR #770, opened 2026-04-16
Ingest suite to download raw obs and convert it to ioda format, and ingest it to R2D2 

### LINKED ISSUE #769: Extend `ingest_obs` to download, convert to ioda, and ingest the obs to r2d2
Extend the ingest_obs suite to download, convert, and ingest **OMPS Nadir Mapper O3** HDF5 observations into R2D2, using NASA GES DISC as the data source.

Depends on https://github.com/JCSDA-internal/ioda-converters/pull/1768

mer-a-o:
## Description

This PR adds an end-to-end "download-convert-ingest" pipeline with initial support for OMPS O3 Nadir Mapper (OMPS-O3-NM) data in the GEOS-CF interface.

**Summary of updates**

* `DownloadObs`: downloads raw observation files from remote servers (e.g. via wget) for a given cycle window. Authenticates using ~/.netrc instead of a hardcoded Earthdata token. 

* `ConvertObsToIoda`: runs an ioda-converters script for the downloaded raw files to produce a single IODA-formatted NetCDF per cycle. The converter binary directory is configurable via `converter_path`, falling back to jedi_bundle/build/bin.

* `flow.cylc` in obs ingest suite now supports optional DownloadObs -> ConvertObsToIoda -> IngestObs task chain, gated by a new `download_convert_pipeline` boolean flag. When the flag is false, the suite runs IngestObs alone (preserving existing behavior).
    * Suggestions for a better approach are welcome.

* In `IngestObs`, a `local` acquisition method is added so IngestObs can pick up files produced by `ConvertObsToIoda` within the same cycle, in addition to the existing `cp` (and s3) path methods.

* New obs configuration files (geos_cf)
  * `download_observations/omps_o3_nm.yaml`: wget pipeline config with filename pattern and date placeholders.
  * `convert_observations/omps_o3_nm.yaml`: ioda-converter script name, output filename template, and extra converter flags (-q, -e).
  * `ingest_observations/omps_o3_nm.yaml`: R2D2 ingest config using the new local retrieval method.
  

## Dependencies
- [x] https://github.com/JCSDA-internal/ioda-converters/pull/1768

closes: https://github.com/GEOS-ESM/swell/issues/769


src/swell/deployment/platforms/nccs_discover_sles15/modules
mer-a-o, line 26: I believe `pyioda` should get added to `PYTHONPATH` here but when I tested it didn't work and had to update this file. I will dig a little more.

mranst: You need to add the step to clone/build jedi in flow.cylc, otherwise this location never gets linked

```
            # Triggers for non cycle time dependent tasks
            # -------------------------------------------
            # Clone JEDI source code
            CloneJedi

            # Build JEDI source code by linking
            CloneJedi => BuildJediByLinking?

            # If not able to link to build create the build
            BuildJediByLinking:fail? => BuildJedi
```

src/swell/suites/ingest_obs/suite_config.py
mranst, line 1: There needs to be an entry for download_convert_pipeline in all suite configs, otherwise it breaks when trying to create flow.cylc

mranst (commented):
I was able to run the default case 👍

src/swell/configuration/jedi/interfaces/geos_cf/ingest_observations/omps_o3_nm.yaml
```diff
@@ -0,0 +1,5 @@
+# OMPS Nadir Mapper Total Ozone L2 — R2D2 ingestion configuration
+# Source: output of ConvertObsToIoda, relative to the cycle work directory.
+
+retrieval_method: local
```
jeromebarre: I would call this differently as we plan to call the L2 `obs_type` `retrieval`.
`aquisition_method` is preferrable here. Please change through the code.

src/swell/configuration/jedi/observation_ioda_names.yaml
jeromebarre, line 345: I am really not sure what is the purpose of this yaml... but in terms of consistency with what has been done for the observation above I would rename all L2 products as `retrieval`. so replace all trace gas`inst type` which are `no2, co, o3 `with `retrieval`

jeromebarre, line 348: this should be `nasa` remove `_VCD`

jeromebarre, line 352: this should be `nasa` remove `_VCD`

src/swell/configuration/jedi/observation_ioda_names.yaml
```diff
@@ -358,3 +358,7 @@ ioda instrument names:
     full name: TROPOMI tropospheric column CO
     inst type: co
     provider : esa
+  - ioda name: omps_o3_nm
+    full name: OMPS total column O3 (Nadir Mapping)
```
jeromebarre: Mapper

src/swell/configuration/jedi/observation_ioda_names.yaml
```diff
@@ -358,3 +358,7 @@ ioda instrument names:
     full name: TROPOMI tropospheric column CO
     inst type: co
     provider : esa
+  - ioda name: omps_o3_nm
```
jeromebarre: the name for FP is `ompsnm_npp` but to be consistent with the other CF names we could use `omps_npp_o3_total`

src/swell/suites/ingest_obs/suite_config.py
```diff
@@ -38,3 +38,22 @@ class SuiteConfig(QuestionContainer, Enum):
...
+            qd.final_cycle_point("2023-08-11T00:00:00Z"),
+            qd.model_components(['geos_cf']),
+            qd.runahead_limit("P5"),
+            qd.download_convert_pipeline(True),
+        ],
+        geos_cf=[
+            qd.window_length("PT6H"),
+            qd.obs_to_download(['omps_o3_nm']),
+            qd.obs_to_ingest(['omps_o3_nm']),
+            qd.converter_path('/discover/nobackup/projects/jcsda/s2127/maryamao/jedi-bundle/build-intel-1.9/bin/'),
```
ftgoktas: Will everyone have access to this directory since it's under a personal username? Could be better if we put it under a shared team path.

src/swell/configuration/jedi/interfaces/geos_cf/download_observations/omps_o3_nm.yaml
```diff
@@ -0,0 +1,18 @@
...
+# Product:    OMPS_NPP_NMTO3_L2 v2.1
+# Source:     NASA GES DISC (Earthdata authentication required in ~/.netrc)
+#
+# Files are named: OMPS-NPP_NMTO3-L2_v2.1_YYYYmMMDDtHH*.h5
+# where HH is the UTC hour at the start of the orbit granule.
+
+remote_host: https://snpp-omps.gesdisc.eosdis.nasa.gov
+remote_path_template: /data/SNPP_OMPS_Level2/OMPS_NPP_NMTO3_L2.2/YYYY/JJJ/
+filename_pattern: OMPS-NPP_NMTO3-L2_v2.1_YYYYmMMDDtHH*.h5
+auth_type: earthdata_token
```
ftgoktas: Looks like `auth_type` isn't being read anywhere in `DownloadObs`.

src/swell/tasks/download_obs.py
```diff
@@ -0,0 +1,230 @@
...
+        """Return filenames found in an HTML directory listing at ``url``."""
+        response = session.get(url)
+        response.raise_for_status()
+        return re.findall(r'href="([^"/][^"]*)"', response.text)
+
+    def _download_file(
+        self, session: requests.Session, url: str, dest_path: str
+    ) -> None:
+        """Stream a remote file to ``dest_path`` in 1 MB chunks."""
+        with session.get(url, stream=True) as response:
```
ftgoktas: We need to have a `timeout` argument in `_download_file` and `_list_remote_dir`.  If the server doesn't respond, the task will wait for 30 minutes before cylc kills it with no useful error message. 10s timeout for connect and a 300s read timeout should be enough.

src/swell/suites/ingest_obs/suite_config.py
mer-a-o: Thanks, adding this solved the issue with finding pyioda

src/swell/configuration/jedi/interfaces/geos_cf/ingest_observations/omps_o3_nm.yaml
mer-a-o: done

src/swell/configuration/jedi/observation_ioda_names.yaml
mer-a-o: We can change this once we switch to the new data_store. In the current version, it is saved under `nasa_VCD`.

src/swell/configuration/jedi/observation_ioda_names.yaml
mer-a-o: done

src/swell/configuration/jedi/observation_ioda_names.yaml
mer-a-o: We can change this once we switch to the new data_store. In the current version, it is saved under `nasa_VCD`.

src/swell/configuration/jedi/observation_ioda_names.yaml
mer-a-o: I changed it to `omps_o3_nm_total`. Note that `omps_o3_nm_total` will be the obs name in R2D2.

src/swell/suites/ingest_obs/suite_config.py
mer-a-o: Yes, this should be accessible. Once the a newer version of ioda-converters with updated `omps_o3_nm_h52ioda.py` is built we can remove this line. Then the code use the converter in `existing_jedi_build_directory`
https://github.com/GEOS-ESM/swell/blob/develop/src/swell/deployment/platforms/nccs_discover_sles15/task_questions.yaml#L10-L11

src/swell/configuration/jedi/interfaces/geos_cf/download_observations/omps_o3_nm.yaml
mer-a-o: I removed it.

src/swell/tasks/download_obs.py
mer-a-o: done

src/swell/configuration/jedi/observation_ioda_names.yaml
mer-a-o: done

jeromebarre (dismissed):
Looks good to me now!

docs/examples/r2d2/ingest_obs.md
mranst, line 1: I initially had a string of comments suggesting not to advise users to configure options in `suite_config.py`, but to use the override yaml. The reasons for this were that some people use static installs that would make this cumbersome to use, and forcing common user configuration options to be made in source code generally isn't ideal (in this case, users can always still the override, but might get confused). I've deleted these as I realized these docs are written more with devs in mind who will be creating new configurations for obs, but I did want to mention this for consideration

src/swell/suites/ingest_obs/suite_config.py
mranst: Add `qd.download_convert_pipeline(False)` to the base `ingest_obs` config. `swell create ingest_obs_marine` is broken because the config has no entry for the `download_convert_pipeline` key in `flow.cylc`

mranst (changes requested):
One fix needs to be made to the suite_config but otherwise looks good

docs/examples/r2d2/ingest_obs.md
mer-a-o: I added this to the document, does it capture your concern?

If you are developing a new built-in observation configuration, update
`suite_config.py` and the source configuration files below. If you are running
from an installed/static Swell environment, prefer an override YAML or edit the
copied configuration in your experiment directory after `swell create`.

mer-a-o:
@mranst This PR is ready for merge (if all your comments are answered).

docs/examples/r2d2/ingest_obs.md
mranst: Very much, thanks!

mranst (approved):

PR #772, opened 2026-04-21
Set stall timeout from swell launch

mranst:
@rtodling This pr adds the flag '-t` to the swell launch command to dynamically set the cylc stall timeout.

`swell launch <experiment> -b -t PT30S`

The launch timeout is usually set to an hour by default, or to the value under `~/.cylc/flow/global.cylc`. This allows the user to set it manually, intended mostly for the purpose of throwing an error immediately after the suite fails. The implementation here is not ideal, but this will be made simpler in #656.


src/swell/deployment/create_experiment.py
```diff
@@ -589,9 +589,18 @@ def prepare_cylc_suite_jinja2(
     render_dictionary['scheduling']['BuildJedi']['execution_time_limit'] = 'PT3H'
     render_dictionary['scheduling']['EvaObservations']['execution_time_limit'] = 'PT30M'
 
+    # Set jinja templated string to use upon runtime
+    # ----------------------------------------------
+    render_dictionary['scheduling']['stall_timeout'] = (
+            "    {% if environ['SWELL_CYLC_TIMEOUT'] != 'unset' %}"
+            "\n    [[events]]"
+            "\n        stall timeout = {{environ['SWELL_CYLC_TIMEOUT']}}"
+            "\n    {% endif %}")
```
ghost: Multi-line string is cleaner.

```suggestion
    render_dictionary['scheduling']['stall_timeout'] = """\
    {% if environ['SWELL_CYLC_TIMEOUT'] != 'unset' %}"
    [[events]]"
    stall timeout = {{environ['SWELL_CYLC_TIMEOUT']}}"
    {% endif %}"""
            
```

ghost: Setting an unset variable to `"unset"` or other such sentinel value is an anti-pattern. Just do this (combined with the change to `launch_experiment` below:

```suggestion
    render_dictionary['scheduling']['stall_timeout'] = """\
    {% if environ.get('SWELL_CYLC_TIMEOUT') %}"
    [[events]]"
    stall timeout = {{environ['SWELL_CYLC_TIMEOUT']}}"
    {% endif %}"""
```

Or, you can do `if 'SWELL_CYCL_TIMEOUT' in environ`.

src/swell/deployment/launch_experiment.py
```diff
@@ -50,6 +52,12 @@ def cylc_run_experiment(self) -> None:  # NB: Could be a factory based on workfl
         if os.path.exists(self.suite_path + 'global.cylc'):
             os.environ['CYLC_CONF_PATH'] = self.suite_path
 
+        # Tell cylc to exit a stall quickly
+        if self.cylc_timeout is not None:
+            os.environ['SWELL_CYLC_TIMEOUT'] = self.cylc_timeout
+        else:
+            os.environ['SWELL_CYLC_TIMEOUT'] = 'unset'
```
ghost: Two things:

1. Per my comment above, setting the variable to `"unset"` is an anti-pattern. I would just leave the variable literally unset.

```python
        if self.cylc_timeout is not None:
            os.environ['SWELL_CYLC_TIMEOUT'] = self.cylc_timeout
        # no else
```

2. Do you want to give users the ability to control this directly via environment variable? Might come in handy for certain edge cases. If so, consider leaving any user-specified values of `SELF_CYLC_TIMEOUT` unchanged.

```python
        if self.cylc_timeout and 'SELF_CYLC_TIMEOUT' not in os.environ:
```

src/swell/suites/3dfgat_atmos/flow.cylc
```diff
@@ -1,3 +1,4 @@
+#!jinja2
```
ghost, line 1: This is new. Are these a requirement and/or best practice of jinja templates? (We can leave them in --- they're useful to have; just asking for my own awareness).

ghost (changes requested):
The `variable = "unset"` thing is an anti-pattern, and cuts off the ability to set this directly via environment variable (which seems useful to preserve). But otherwise, this looks good — happy to approve once that's resolved.

src/swell/deployment/launch_experiment.py
mranst: Fixed, thanks. As for 2, the way it is currently a user would have to go out of their way to specify `-t`, so I think it would be best to allow that to override the environment variable, so the hierarchy of control would be (in ascending order)
1. Cylc default
2. Value in global.cylc
3. User-set environment variable
4. User-specified value on command line

src/swell/suites/3dfgat_atmos/flow.cylc
mranst: The cylc docs make it sound as if it's a requirement, though I have actually found it's not necessary to function. Figured it would be best to include it, though.

rtodling:
Ok, I am trying to test this out, but SWELL is not backward compatible. So, an exp.yaml I had not longer works w/ the version of SWELL used in this branch.

This means for me to test this I need to recreate my suite, readapt the exp.yaml and only then will I be able to test this.

rtodling:
> Ok, I am trying to test this out, but SWELL is not backward compatible. So, an exp.yaml I had not longer works w/ the version of SWELL used in this branch.
> 
> This means for me to test this I need to recreate my suite, readapt the exp.yaml and only then will I be able to test this.

rtodling (dismissed):
Works for me. I created a case that I know things would crash ... ran twice, once w/o the time opt at launch and it hang for a long time; another time w/ the 15 sec tolerance of existing time. Code exist w/in 15sec and returns a non-zero status. 

All good from my perspective.

mranst:
@shiklomanov-an When you get a chance, could you take another look at this one? I can't merge since there are requested changes

ghost (approved):
Sorry for the delay --- looks good to me, assuming tests pass!

rtodling (approved):
Works for me in the context of context of convert_ncdiags

PR #776, opened 2026-04-29
Add changes to satwind yaml based on the satwind code changes in x0054.

gmao-yzhu:
Satwind usages in x0054 partially turned on qc_satwind option. This PR is to add necessary corresponding changes in satwind.yaml for JEDI UFO.


gmao-yzhu:
These new changes to satwind data usage exclude additional data, so data counts passing QC are reduced as expected.  

A geoval variable is used in the satwind yaml file with different name in different version. Here are the names used for this variable: height_above_mean_sea_level_at_surface, geopotential_height_at_surface, surface_geometric_height.

Although height_above_mean_sea_level_at_surface is used in this PR, I would prefer geopotential_height_at_surface.
Need to make sure the same name is used in Geovals. 

hofx and OmF passing QC for each kx (240-260) matches those from GSI. 
gsi_hofx_vs_jedi_hofx_QC_satwind_windEastward_252
gsi_hofx_vs_jedi_hofx_QC_satwind_windEastward_250

JEDI effective observation error also matches GSI final observation errors for each kx from 240-260 except for 250 and 252. For 250, with 3985 observations passing QC, 276 observations have different observation errors from GSI observation errors and they are smaller than GSI observation errors. The corresponding code changes in this PR only remove observations, as shown above, data are removed as expected, but observation errors are not supposed to be affected. Comparing GSI and JEDI codes, it is seen that observation error inflation due to duplicated observations are implmented slightly different, dup is accumulated before QC in GSI while QC is considered in JEDI. I think JEDI implementation is more reasonable in this case, so I don't make code changes to UFO. kx 250 and 252 also show observation error difference between JEDI and GSI before changes are added, but with only a couple  observations. With more data are removed after changes are added in, more observation locations are affected, so we see more data with smaller observation errors than GSI final observation errors. 
obserror_250_gsipropressing_one

rtodling:
I just want to add a little more on the tests here. The figs here compare stats from a 3dvar test before and after the changes proposed here. This is comparing x0053RPY v x0054 (first analysis cycle - ie, same background fields) - and comparing JEDI before and after the changes proposed here.

This is what I see for VIS GMS winds (kx=242): **x0053RPY** (before satwind changes)



and this is what I see in **x0054** (after sat-wind changes introduced in GSI) for the same kx=242:



in order words, data above approx. 700 mb has been excluded from the analysis for this data source.


Now this is what I see for JEDI for the same kx=242 (VIS GMS):

**DEFAULT** (before this PR):



**with the changes in this PR**




It should be noted that in my comps above, JEDI uses a different date/time and backgrounds than that in the x-exp comps. JEDI uses the present 20231010 default date.

I mean to say that I believe the changes introduced here seem to be consistent with the changes introduced the GSI when going from v5.43 to v5.44 (x0053RPY to x0054).

rtodling:
I checked other observing systems - following the comps I showed above - and things seem consistent to my eyes.

rtodling (approved):
These changes seems good to me.

Dooruk:
I will merge this now as @gmao-yzhu is away for a while

rtodling:
Hum ... oddly enouugh I am puttig these changes into a cycling experiment and I am getting the winds to go wild - to the point of crashing the model. To be fair, I don't know if this is the change that causes the problem, but ...

rtodling:
I don't know - it looks like the issue I am having w/ my cycling run is unrelated ... this looks good as is.

rtodling:
Now folded into:  https://github.com/GEOS-ESM/swell/pull/835

Issue #697, opened 2026-02-05
Use persistent SLURM workers


shiklomanov-an:
Right now, our general Slurm pattern is that cylc submits new jobs every time it needs to do something in slurm via `sbatch`. That makes sense as long as the number of slurm jobs you need to run is relatively small, the jobs are relatively large, and the jobs each have relatively unique resource requirements.

However, for the case where the jobs have the same resource requirements (or, at least, can accommodate the same resources), a shortcut is to request the node(s) asynchronously at the beginning of the workflow and then submit tasks directly to those nodes _without waiting around in the queue_.

I just tested this on Discover and can confirm that this works:

```sh
# Request a node, but don't connect to it (--no-shell)
salloc -A <account> --time=<HH:MM:SS> --job-name=workers --no-shell

# Wait until the node gets allocated...

# Get the JobID (filter to current user and job name "workers")
# --format=%i returns only the job ID
# head -n 1 ensures only the most recent job number (adapt if you submit multiple such jobs; if only one, can drop this)
JOB_ID=$(squeue -u $(whoami) -n workers --noheader --format=%i | head -n 1)

# Submit commands directly to the job *without waiting in the queue!*
# These commands will run synchronously
srun --jobid=$JOB_ID echo "hello, world"
srun --jobid=$JOB_ID hostname
# ^^ useful test; should return the compute node name

# Cancel the allocation when done
scancel $JOB_ID
```

We should look into whether we can hack cylc's job submission system into using this pattern somehow (e.g., via a custom job runner) when it's appropriate.

But even if we can't get `cylc` to do this natively, there may be situations when we can just do this sort of thing by hand (e.g., for testing individual Swell tasks).

Dooruk:
I just saw this, does the following help with what you had in mind:

https://www.nccs.nasa.gov/nccs-users/instructional/using-slurm/multiple-jobs

Downside is `--packable` option is only available for certain project groups, unless they activated it by default.

ftgoktas:
I implemented this approach for the `3dvar_atmos` suite and ran several tests on Discover (See PR #777). Instead of a custom job runner, I added a `persistent_workers` flag to `slurm.yaml`. When enabled, an `Allocate` task runs `salloc --no-shell` at the start of the workflow and all heavy tasks (`RunJediVariationalExecutable`, `BuildJedi`, `EvaObservations`) run as `srun --jobid` steps inside that allocation. A `Deallocate` task calls `scancel` once all cycles finish.  
[...]

PR #777, opened 2026-05-05
Adding persistent workers for 3dvar_atmos suite

ftgoktas:
When running a workflow, every heavy task (`RunJediVariationalExecutable`, `BuildJedi`, `EvaObservations`) normally submits its own `sbatch` job to SLURM. Each job waits in the queue before it can run, even if the previous task just finished 2 seconds ago.

The changes adds a `persistent_workers` flag to `slurm.yaml`. When `persistent_workers` is set to `true`, Swell reserves compute nodes at the start of the workflow using `salloc --no-shell` and holds them until all cycles finish. Every heavy task then runs as an `srun --jobid` step inside that allocation  without individual queue waits. The nodes are released via `scancel` once all cycles finish.

How to test:                                                                                                                                                           
                                                                                                                                                                        
Before creating an experiment, make your `src/swell/deployment/platforms/nccs_discover_cascade/slurm.yaml` has:           
                                                                                                                                                                        
- `persistent_workers: true`

Some things I've tried and encountered errors during testing (fixed now):                                                                                                                                           
                                                                                                         
  1. **MPI processes couldn't talk across nodes**: Calling `srun --jobid` from a login node (outside a batch job) requires `--mpi=pmi2` to set up MPI across nodes. Without it, each node starts its own independent group of MPI processes instead of one single communicator, and the run fails.
  2. **Wrong `--ntasks-per-node`**: The global `ntasks-per-node: 64` from `slurm.yaml` was being passed to the `srun` step. The correct value is `np // nodes (96 / 3 = 32)`. Passing 64 would request 192 total tasks and cause the run to fail.                                                                          
  3. **Single-task steps killed by CPU cgroup**: `srun --ntasks=1` allocates exactly 1 CPU. `EvaObservations` creates 40 Python workers and `BuildJedi` runs parallel compile threads. SLURM's `cgroup` killed both. Fixed by adding `--cpus-per-task` from each task's `srun_directives`.
  4. **EvaObservations blocked by RunJedi's `--exclusive lock`**: RunJediVariationalExecutable uses `--exclusive` which locks all allocated nodes. `EvaObservations` with `--exclusive` would queue behind it indefinitely. Fixed by running `EvaObservations` as `srun --ntasks=1` without `--exclusive`  it only needs CPUs on one node and doesn't need to lock all three.

#697


mer-a-o:
@ftgoktas, here's the branch I mentioned today: https://github.com/GEOS-ESM/swell/pull/800
Each background tar ball includes 6 background files. I need to run ConvertState task for each of them to change their resolutions. These tasks are independent of each other. Right now each takes one node which is not efficient.

image


image

ftgoktas:
Ran 100 trials of each approach on Discover. Both sets submitted simultaneously so they were under same cluster conditions.

  | | Mean queue wait | Max queue wait |
  |---|---|---|
  | Non-persistent (RunJediVariationalExecutable, 3 nodes, 1hr limit) | ~25 min (1508s) | ~40 min (2411s) |
  | Persistent (`salloc`, 3 nodes, 40min limit) | ~4.6 min (278s) | ~10 min (604s) |

So it's 5.4x improvement in average queue wait. The max on non-persistent (~40 min) matches what was being observed on real `3dvar_atmos` runs before this approach.

Issue #699, opened 2026-02-05
adding jedi TIER 2 testing to SWELL


rtodling:
Is there a way we could add JEDI TIER 2 testing in SWELL - at least for fv3-jedi? 

What I understand is that JEDI (JCSDA) does not do CI for TIER 2 - which nearly makes TIER 2 pointless. 

My point is this: if JCSDA is not running those tests, can we run them before we decide when to move the pin?

Dooruk:
Do we need to build with a certain switch turned on for Tier2 tests? @jeromebarre @mer-a-o knows this better

jeromebarre:
before build set: `export FV3JEDI_TEST_TIER=2`

mranst:
It seems like that might already be done, this is from the nightly tier2 test jedi build:
```
-- Use test data from fv3-jedi-data repository
-- The env variable FV3JEDI_TEST_TIER is defined, the test tier will be set from this env variable
-- Running fv3jedi tests for tiers up to and including tier 2
-- Creating static B yaml files for cycle: 2020-12-14:21
-- Creating static B yaml files for cycle: 2020-12-15:00
-- Creating static B yaml files for cycle: 2020-12-15:03
-- Creating static B yaml files for cycle: 2020-12-14:21
-- Creating static B yaml files for cycle: 2020-12-15:00
-- Creating static B yaml files for cycle: 2020-12-15:03
-- Creating static B yaml files for cycle: 2018-04-14:21
-- Creating static B yaml files for cycle: 2018-04-15:00
-- Creating static B yaml files for cycle: 2018-04-15:03
-- Creating ensemble B yaml files for cycle: 2020-12-14:21
-- Creating ensemble B yaml files for cycle: 2020-12-15:00
-- Creating ensemble B yaml files for cycle: 2020-12-15:03
```
(If these are indeed the tier tests you are looking for) Is there something we should be looking for in these files? Or is the fact that they run successfully good enough?

PR #778, opened 2026-05-06
Add processing and comparison of ctests in JEDI builds

mranst:
## Description

Addressing #699 and #442, this PR adds tasks to run ctests for selected bundles within JEDI, and compare the number passed between two builds. I'm envisioning this being a part of tier2 tests by running ctests for fv3-jedi whevever JEDI is built nightly, then confirming the same tests pass as the previous pinned version (For fv3-jedi, at least, there's always some that fail). It's currently geared to run for fv3-jedi, but can theoretically work for any JEDI bundle

Output of CompareJediCtests looks like this:
```
CTL: /discover/nobackup/manstett/SwellExperiments/swell-build_jedi/swell-build_jedi-suite/experiment.yaml
EXP: /discover/nobackup/manstett/SwellExperiments/swell-build_jedi/swell-build_jedi-suite/experiment.yaml

fv3-jedi                               CTL   EXP
fv3jedi_staticb_nicas_gfs              Fail  Fail
fv3jedi_hofx_nomodel_abi_radii         Fail  Fail
fv3jedi_staticb_split_nicas_gfs        Fail  Fail
fv3jedi_hyb                            Fail  Fail
fv3jedi_staticb_dirac_local_gfs_12pe   Fail  Fail
fv3jedi_staticb_cor_geos               Fail  Fail
fv3jedi_staticb_dirac_local_gfs_6pe    Fail  Fail
fv3jedi_staticb_nicas_geos             Fail  Fail
fv3jedi_staticb_dirac_global_gfs_6pe   Fail  Fail
fv3jedi_staticb_dirac_global_gfs_12pe  Fail  Fail
```


mer-a-o:
@mranst, Can you please add documentation for this new feature?

src/swell/tasks/compare_jedi_ctests.py
```diff
@@ -0,0 +1,140 @@
...
+        # Format the string for readable output
+        results_str = 'JEDI CTest Results Comparison\n'
+        results_str += f'{experiment_tag_1}: {experiment_path_1}\n'
+        results_str += f'{experiment_tag_2}: {experiment_path_2}\n\n'
+
+        width_col_1 = max(len('Fail'), len(experiment_tag_1)) + 2
+        width_col_2 = max(len('Fail'), len(experiment_tag_2)) + 2
+
+        for bundle in bundles:
+            max_width = max(len(bundle), max([results_dict[bundle][test]['width']
```
mer-a-o, line 109: What will happen here if all ctests pass for both experiments?

src/swell/tasks/run_jedi_ctests.py
```diff
@@ -0,0 +1,55 @@
...
+        # Identify the bundles to run ctests on
+        bundles = self.config.bundles_to_run_ctests()
+
+        for bundle in bundles:
+            bundle_dir = os.path.join(build_dir, bundle)
+
+            # Run the ctests
+            cwd = os.getcwd()
+            os.chdir(bundle_dir)
+            command = ['ctest', '-V', '-I']
```
mer-a-o: What does `-I` do, I'm not familiar with it?

src/swell/swell.py
```diff
@@ -303,3 +303,6 @@ def main() -> None:
 
 
 # --------------------------------------------------------------------------------------------------
+
+if __name__ == '__main__':
+    test('code_tests')
```
mer-a-o: is this debugging leftover?

src/swell/tasks/compare_jedi_ctests.py
mranst: There wouldn't be any tests listed, I've added a line specifying all tests pass. If all tests fail for both, this task would pass, I rely on the user to ensure that the control build is acceptable

src/swell/tasks/run_jedi_ctests.py
mranst: `-I` lets you run a certain range of tests, for example -I 4,5 would run tests 4 and 5. I removed it here since it's not doing anything

src/swell/swell.py
mranst: Yes, sorry about that

mer-a-o (approved):

src/swell/utilities/question_defaults.py
```diff
@@ -367,6 +367,27 @@ class bundles(TaskQuestion):
 
     # --------------------------------------------------------------------------------------------------
 
+    @dataclass
+    class bundles_to_run_ctests(TaskQuestion):
+        default_value: list[str] = mutable_field([
+            "fv3-jedi"
```
Dooruk, line 373: Don't want to keep you from merging but did you test it with two bundles here? We could add `soca` at some point?

Dooruk (approved):
Haven't tested it but looks good 👍🏼

PR #785, opened 2026-05-14
Use bufr_query from jedi_bundle

mranst:
## Description

Uses build of JEDI with bufr-query installed (January 15 build), so we can control the installation through `jedi_bundle`. Zero diff from previous January 15 version. Tested with tier1 and convert_bufr suite

### DIFF
--- a/src/swell/deployment/platforms/nccs_discover_cascade/modules
+++ b/src/swell/deployment/platforms/nccs_discover_cascade/modules
@@ -20,6 +20,7 @@ module load soca-env
 module load gmao-swell-env/1.0.0
 module unload gsibec crtm fms
 module load fms/2024.02
+module unload bufr-query
 
 # JEDI Python Path
 # ----------------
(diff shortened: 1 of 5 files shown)


Dooruk:
A quick comment. I see we use `Path` in certain parts of SWELL, e.g.: https://github.com/GEOS-ESM/swell/pull/702. 

Should we be consistent throughout, especially for the new PRs?

mranst:
> A quick comment. I see we use `Path` in certain parts of SWELL, e.g.: #702.
> 
> Should we be consistent throughout, especially for the new PRs?

Used os.path here since it's used elsewhere in the task. Opened an issue for this: #786

Dooruk (approved):
This worked for me. 

Does it take ~10 minutes to run default `convert_bufr`? Is that the expected timeframe @cohen-seth ? It could use some of the parallelization applied in other tasks perhaps.

Issue #722, opened 2026-03-04
Create save analysis task


jeromebarre:
Create save analysis task as this is missing form the current tasks in swell

Dooruk:
I have a branch on this, I will create a draft PR soon. For this to work https://github.com/GEOS-ESM/swell/pull/713 needs to go in though so experiments are registered when they get created.

PR #790, opened 2026-05-18
Compress and SaveForecast with R2D2v3

### LINKED ISSUE #215: Coupled DA runs minor issues
I will document here some of the minor issues I encountered during my coupled runs to address them over time. 

### General 

- [ ] Make `forecast_dir` take list argument only.
- [ ] Google doc-string type documentation https://github.com/GEOS-ESM/swell/pull/677#discussion_r2622837351

### Cylc

- [x] Node definitions, how many nodes are required for jobs. This requires better planning within Swell.

### MOM6

- [x]  `diag_table` is used to generate history files requires a beginning date. This could be dynamic, changing at each forecast time or static. 
- [x] Parallel MOM restart is hard coded so depending on the resolution, random number of multiple restarts will be created. This requires finding proper analysis fields in these restart files for high resolution runs. As a solution, we use IAU and we don't need to update the restart files directly.
- [x] SOCA and GEOS MOM6 model settings, in particular land masks are different. New SOCA MOM6 inputs are required. This causes the NaN/e+20 observations to sneak into SOCA.

### CICE6

- [ ] Saving background happens after soca2cice, which should be before

### GEOS

- [x]  [Critical] Currently Swell copies whatever is in the GEOS run folder to setup hot start. `gcm_setup`, on default, creates `$HOMDIR` and `$EXPDIR` at two separate locations but if the user combines these two this will cause issues in terms of copying everything from the `$HOMDIR`.
  - [x] why not use `~/.HOMDIRroot`??
- [x] HISTORY.rc files will mess up obtaining time from a _rst file. Overall, time information obtaining from _rst files should be abandoned. 
- [x] Why can't we point to external directories for restart and experiment folders?
- [x] `mom6_increment.nc` from a previous run might be located in the `restarts` folder but they won't get copied. That needs to be fixed. 
- [x] In `AGCM.rc`, define `RECORD_REF_DATE` as the beginning of the DA window. That way less checkout files are written which saves on computational time.
- [x] Check to see if `RC` folder is different between months? `gcm_setup` does not have time input so perhaps is is static? Also `RC` folder needs to be copied from `$EXPDIR` -> `$HOMDIR` (I think this is just GEOSgcm version dependent)
- [x] Starting from GEOS v11.6.0, `linkbcs` is a separate file (used to be inside `gcm_run.j`). Hence, like the `RC` folder, that needs to be located inside the `expdir`.
- [x] #499
- [x]  Default `HISTORY.rc` creates geosgcm*_checkpoint files at the beginning of the month. These files don't provide time information like the other _checkpoint/_rst outputs which Swell relies on to set forecast start date currently. This would be a quick fix however.  (Simple solution for now is to turn off `HISTORY.rc` outputs)
- [x] `GeosRunExecutable` fails to run again if there is a previous run outputs  inside the run folder with existing history files. This could lead to some issues as we expect `cylc` to repeat tasks occasionally. Hence, just erase the .nc files created by outputs.
- [ ] `geos_forecast` uses an input `_checkpoint` file to define the start date. There needs to be a mechanism to check if the cycle time is way off from that  start date/time. There is a `MOM_input` setting for this, `FATAL_INCONSISTENT_RESTART_TIME` this is `False` in the GEOS setup.
- [x] `RECORD_FREQUENCY` does not work when the values is equal to or greater than 1 day (see https://github.com/GEOS-ESM/GEOSgcm_App/issues/697#issuecomment-2755623149)
- [x] `LD_PRELOAD` was not defined but now it is. There might be default lib files between different GEOSgcm versions. Following works, similar to `gcm_run.j` command. It's just a matter of picking the correct models depending on the models (MOM6, CICE6 etc.).
```
        command = f'source {geos_modules} \n' + \
            f'cd {cycle_dir} \n' + \
            f'env LD_PRELOAD={geos_lib_path}/libmom6.so:{geos_lib_path}/libCICE4.so ' + \
            f'mpirun -np {np} {geos_executable} ' + \
            f'--logging_config logging.yaml'

```
### R2D2 (fetch in `get_observations`)
- [x] There should be a mechanism to handle missing observations. Either fail immediately or have a certain number for max missing observations.

### JEDI / SOCA
- [x] We need to update static bump files for `IAU` to work with horizontal ocean currents (U,V). MOM6 will expect U, V increments. The U,V assimilation is not active for now.
- [ ] Save backgrounderr option
- [x] Handling of negative durations via isodate is not possible (see the Limitations in their official [page](https://pypi.org/project/isodate/)). Currently `get_background.py` and `write_background.py` tasks likely rely on this. Cylc somehow handles this so we could copy their approach.
- [x] `EXPLICIT_DIFFUSION` requires `MLD` information to generate vertical correlation parameters. With high resolution MOM6 runs, there are multiple restart files (hard coded and numbers are not preset). Hence, this would require a method within `GenerateBClimatology` searching for `MLD` within these multiple restart files. I will for now stick with history outputs which are guaranteed to contain `MLD` (with `diag_table` options that is).

### LINKED ISSUE #547: Running `geos_marine` without cice6 analysis skip saving cice background

### LINKED ISSUE #823: Compressing files before storing them into R2D2
I'm modifying the`SaveForecast` (or background) task where it needs to compress multiple model output files and store them under R2D2. Obvious benefit is saving disk space and no of IOnodes.

For MOM6 & CICE6 history and restarts are NetCDF-3 (64-bit offset and classic) and for certain workflows (such as S2S), there could be 20 total files that needs to be combined, 2.2GB each for ocean there are multiple 4D prognostic fields to save.

So I'm exploring potential (simple and efficient) options that can be used here.

AFAIK in GEOS this is done in a straightforward manner:
https://github.com/GEOS-ESM/GEOSgcm_App/blob/f6f7d231f0e42de0f250f20d0d9ea2ca4dc3ea63/gcm_run.j#L652-L653

GPT suggested using `pigz` or `zstd` via python, both of which would require using compute nodes. I used `pigz` before but I'm not familiar with `zstd`. 

This could be a SWELL dev discussion as well but I'm curious to hear thoughts before that. Any other potential/better approaches?

FWIW, with my initial crude tar.gz test on a login node (sorry Discover) I got %90 compression for sea-ice files (9 x 110MB -> ~150MB) and %50 compression for ocean files (9 x 2.2GB -> ~10GB) which took about 13 minutes.

https://github.com/GEOS-ESM/swell/pull/790

Dooruk:
## New R2D2v3 approach to store and get forecast/background files.

I had to create  `geos_marine` specific tasks as the requirements between marine model interface and the others are quite different. As I eluded to it on multiple occasions, varying DA window lengths and marine history file sizes made it a bit challenging to store files individually. In this PR, all state files are compressed under a single archive first and then gets stored afterwards. Because of R2D2's API calls, compression and store had to be separated into two different tasks. Hence, this PR introduces two major steps:

1)  Listing and compressing all marine states, using `pigz` for parallel gunzip compression using compute nodes. Otherwise doing this on login node was quite inefficient and wasting other people's login source. This uses `compress.py` utility from @ftgoktas's PR: https://github.com/GEOS-ESM/swell/pull/853

For milan nodes, using 8 threads proved to be most efficient so that is hardcoded. pigz compression level varies between 1-9, I picked 6 but that can be modified in the future. It takes ~80 seconds to compress 20GB worth netcdf files to ~9.7GB

2) Saving the same archives using store. Step is hardcoded as `PT00`, which is relative to the background (middle of the window for 3D case, beginning of the window for the FGAT case).

I also had to add a `geos_marine` specific `get_background` task that decompresses these files.

SaveAnalysis task will be a lot more straightforward for geos_marine as it is a single file for 3DVar or FGAT. Might not be the case for 4D analyses fields.

### ALERT 🚨 
⚠️ Important, with this PR, each time the marine cycling suites are executed, new files will be added to R2D2 datastore with unique experiment IDs. Until we sort the scrub and lifetime in R2D2, this will pose a problem on our swell project space, which is currently only 25TB with 500k IOnodes ⚠️  @mranst @ftgoktas


mer-a-o:
@Dooruk, welcome back!

I made a lot of changes to the `saveForecast` task in my PR (mainly removing a significant amount of code): https://github.com/GEOS-ESM/swell/pull/752

I’ll review your changes as well to better understand what Marine needs in terms of saving forecast files. In the meantime, feel free to take a look at the simplified version and let me know whether it makes sense to build on top of that approach.

Given that the R2D2 metadata differs between forecast/background and analysis files, do we need to handle both within the `saveForecast` task?

jeromebarre:
Hi Doruk! Good to have you back.
I think at this point what is mostly needed is save analysis.
https://github.com/GEOS-ESM/swell/issues/722 (at least for GEOS-CF)

Dooruk:
Thanks Jerome and Maryam!

I'm caught up with some other stuff but will review cycling 3dvar_cf and address your points. Short answer is sure, ideal way is to have separate SaveForecast and SaveAnalysis (are you thinking both increment and analysis for this task?). Second point is, are you using the `r2d2.yaml` under the model directory? State names are different in Marine DA, requires more precision due to ice and ocean model handling(and surface, wave, bgc for the future). I see for atm and compo it is a wild card approach (e.g, bkg.*$DATE*). Good discussion for the next Swell Dev.

jeromebarre:
> Thanks Jerome and Maryam!
> 
> I'm caught up with some other stuff but will review cycling 3dvar_cf and address your points. Short answer is sure, ideal way is to have separate SaveForecast and SaveAnalysis (are you thinking both increment and analysis for this task?). Second point is, are you using the `r2d2.yaml` under the model directory? State names are different in Marine DA, requires more precision due to ice and ocean model handling(and surface, wave, bgc for the future). I see for atm and compo it is a wild card approach (e.g, bkg._$DATE_). Good discussion for the next Swell Dev.

I am thinking just saving the analysis as it comes out of JEDI. One can recompute the increment by getting the background files (if they are saved).

For the naming convention we need to use the PT convention for the forecasts: _$DATE_ _ _$PT_ with date being the initialization datetime.
For analysis we can simply decide to save the final analysis (in case of multiple outerloops) with simply _$DATE_ whcih would be the datetime of the analysis. It could be for example several datatimes for an hourly 4DEnVar analysis, or just one datetime for a 3DVar. It doesn't make sense to use PT for anlyses in my opinion.

Dooruk:
Hi all,

As I'm working towards this I noticed an obsolete key. Also I could use some clarifications in certain aspects.

- Is `obs_experiment` obsolete now?

- For `3dvar_cf` `suite_config.py` has `qd.background_experiment("swell_test")` .  Can the unique `r2d2_id` also be used here? I'm testing storing files with r2d2_experiment_id. Our goal was to have these unique id's after all, correct?

- I think you answered this before but @mer-a-o and @jeromebarre you are using `forecast` to store both background and restarts, correct?

Dooruk:
The last step I need to do is to add a toy model resolution compressed background for `3dvar_marine` to work (@ftgoktas may be you can help?). Other than that it passes marine suite and is ready for a review, see the description for more details.

ftgoktas:
Tested `RunCompressForecast` & `SaveForecast` on Discover and both tasks succeeded with the default `3dvar_marine_cycle` config. They compressed and stored a couple of cycles fine and confirmed stored data in R2D2. 

For `GetBackground`, I compressed and ingested manually and it pulled and unpacked it with no issues.

 If anyone wants to try `3dvar_marine` themselves without dealing with any of that, I stored it under `r2d2_experiment_id: pr790_test`, resolution `72x36`, date `2021-07-01T12:00:00Z`. Just set that in your experiment.yaml and `GetBackground` should be able to fetch it. 

Heads up though, it's registered `debug` lifetime and it's just a borrowed file relabeled to that date, so treat it as a quick way to confirm the plumbing works.

Dooruk:
> If anyone wants to try `3dvar_marine` themselves without dealing with any of that, I stored it under `r2d2_experiment_id: pr790_test`, resolution `72x36`, date `2021-07-01T12:00:00Z`. Just set that in your experiment.yaml and `GetBackground` should be able to fetch it.
> 
> Heads up though, it's registered `debug` lifetime and it's just a borrowed file relabeled to that date, so treat it as a quick way to confirm the plumbing works.

We don't have an active scrubber currently though, correct?

Thanks for testing. I'm running tier1 tests, if it passes I'm going to need approvals please. Trying to get this and his remaining PRs in before @ftgoktas leaves.

Dooruk:
This is good to go in.

ftgoktas:
> > If anyone wants to try `3dvar_marine` themselves without dealing with any of that, I stored it under `r2d2_experiment_id: pr790_test`, resolution `72x36`, date `2021-07-01T12:00:00Z`. Just set that in your experiment.yaml and `GetBackground` should be able to fetch it.
> > Heads up though, it's registered `debug` lifetime and it's just a borrowed file relabeled to that date, so treat it as a quick way to confirm the plumbing works.
> 
> We don't have an active scrubber currently though, correct?
> 
> Thanks for testing. I'm running tier1 tests, if it passes I'm going to need approvals please. Trying to get this and his remaining PRs in before @ftgoktas leaves.

Yes, correct, no scrubber as of right now.

ftgoktas (approved):

PR #791, opened 2026-05-18
Bugfix in linking nicas files

### LINKED ISSUE #787: [bug] 3dvar_cf runs with ID and hofx links nicas files
I noticed that in 3dvar_cf experiment, B is set to identity even with https://github.com/GEOS-ESM/swell/blob/develop/src/swell/suites/3dvar_cf/suite_config.py#L44-L45

and also in hofx_cf nicas files get linked during the staging. We need to clean and fix these two bugs.

mer-a-o:
This PR addresses the issue described here: 
closes: https://github.com/GEOS-ESM/swell/issues/787

### DIFF
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/suite_questions.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/suite_questions.yaml
@@ -8,5 +8,11 @@ ensemble_hofx_strategy:
 ensemble_hofx_packets:
   default_value: 1
 
+saber_central_block:
+  default_value: None
+
+saber_outer_block:
+  default_value: None
+
 skip_ensemble_hofx:
   default_value: true
--- a/src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
+++ b/src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml
@@ -330,12 +330,6 @@ produce_geovals:
   - true
   - false
 
-saber_central_block:
-  default_value: None
-
-saber_outer_block:
-  default_value: None
-
 skip_ensemble_hofx:
   default_value: true
   options:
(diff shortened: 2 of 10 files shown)


jeromebarre:
Thanks for spotting this. This might change CI results though. Not sure @mranst ?

mranst:
I don't see any issue with the CI tests, but I'm slightly confused about this fix. In this case, `get_key_for_model` should render the same result as the regular `config.get` method. I'm also getting `bump_nicas` in the default config for `3dvar_cf`, so I may not be fully understanding the issue

mranst:
If the issue is that `bump_nicas` shouldn't be in the config for `hofx_cf`, I would recommend just setting the value to None in `suite_config.py`, or the cf `task_questions.yaml` file

mer-a-o:
@mranst, You're right. Setting `default_value: identity` in `src/swell/configuration/jedi/interfaces/geos_cf/task_questions.yaml` solves the issue.

```
saber_central_block:
  default_value: identity
  options:
  - identity
  - bump_nicas
```

I was trying to make the logic a bit more correct. Fundamentally, hofx shouldn't know about `saber_central_block`. But changing the default value should be fine too, unless you have any other suggestion? :)

mranst:
> I was trying to make the logic a bit more correct. Fundamentally, hofx shouldn't know about `saber_central_block`. But changing the default value should be fine too, unless you have any other suggestion? :)

Ah okay, I understand better now. If you don't want `saber_central_block` to show up in the `hofx` config at all, you could turn it into a suite question instead of a task question. It will then only show up in configs where you specify it.

mer-a-o:
@mranst, thanks for the suggestion. I turned `saber_central_block` and `saber_outer_block` into a suite question. I tested different suites with the updates and it's working as expected.

mranst (approved):
Looks good 👍 I moved the questions to be with the other suite questions

PR #793, opened 2026-05-19
Marine observation ingesting into R2D2

Dooruk:
Work in progress, but works for the most part.

New R2D2 data-store location is:
`/discover/nobackup/projects/gmao/swell/r2d2-experiments-nccs-gmao/observation`

Currently there are two ways to ingest observations into R2D2:

1) Work in this PR allows ingesting files that are described under `ingest_observations` into R2D2. These ODAS files were already converted to IODA with the existing approach and organized under:
`/discover/nobackup/projects/gmao/soca/obs/ioda/ocean`

Note that there are NOAA/EMC observations as well located here: `/discover/nobackup/projects/gmao/soca/obs/ioda/gdas_obs`

The naming convention they have was kept for simplicity. Ingest YAMLs contain date templated `cp_source` entries.

For ODAS obs, source will look like this:
```yaml
cp_source: '/discover/nobackup/projects/gmao/soca/obs/ioda/ocean/temp_profile_xbt/YYYY/MM/ioda-obs-YYYYMMDDHH-temp_profile_xbt.nc'
```

For GDAS obs it will look like this:
```yaml
cp_source: '/discover/nobackup/projects/gmao/soca/obs/ioda/gdas_obs/gdas_YYYY/gdas.YYYYMMDD/HH/ocean/sst/gdas.tHHz.sst_abi_g18.nc'
```

2) A second approach that is only active for GEOS-CF that downloads, converts (using iodaconverters), and then ingests file into R2D2. More details on this can be found here:

https://geos-esm.github.io/swell/#/examples/r2d2/ingest_obs

We (i.e., @vrx-, @DavidSRussell, @lren20) should figure out directly downloading from source, or for now tapping into the raw observations in D-BOSS. Some additional steps will be required in terms of combining files as some L2 files are sources as swaths.


DavidSRussell:
- Does the first commit mean symlinks instead of copies are now the default in R2D2?
- For `time_begin` in the observation YAMLs, do you want a launch date or the date of the first usable data?

Dooruk:
> * Does the first commit mean symlinks instead of copies are now the default in R2D2?

I think better way would be to have that as an option. I do prefer symlinks as R2D2 (SQL) indexing gives files random numbers and it's impossible to know what they contain without ncview or ncdump.
 
>     * For `time_begin` in the observation YAMLs, do you want a launch date or the date of the first usable data?

Good question, I was thinking first usable data. I did use AI so we should take a second look for sure.

Dooruk:
I don't want to create a separate issue for this but @vrx- and @DavidSRussell when you are working on the converters, can you make sure the extensions are `*.nc4`? 

I'm realizing that is not what I was doing with gmao2ioda script and that is creating some annoyance with file ingestion/access. I have to address that in this PR as well.

Because IODA files are NetCDF4*

src/swell/configuration/jedi/interfaces/geos_marine/ingest_observations/adt_swot_nadir.yaml
```diff
@@ -0,0 +1,6 @@
+# Data availability period (for reference only):
+# time_begin: '2022-12-16T00:00:00Z'
```
DavidSRussell:
```suggestion
# time_begin: '2023-07-21T12:00:00Z'
```

DavidSRussell (dismissed):
Looks good, but I haven't actually tested ingestion yet as SWOT KaRIn is not ready.

I changed the SWOT nadir start date to the start of the science phase, but I haven't checked all the other observers. The raw data we have doesn't always match the mission launch date (e.g. 2017 vs. 2013 for Saral); @lren20 probably knows more about that. I could try tracking down the start date from each source if needed.

Dooruk:
Ok, should be good to go now, can you take a look again? I will leave the dates as is for now as they are irrelevant.

DavidSRussell:
Looks fine to me.

Dooruk:
Can I get an approval  then 😄

DavidSRussell (approved):

PR #799, opened 2026-06-01
Update `r2d2-client` module to `05302026` (`gmaoswell` shared path)

ftgoktas:
Moves `r2d2-client` module load from JediOpt to the new gmaoswell shared path on nobackup.
Picks up one bugfix in `r2d2-client` since the previous module (`r2d2-client/112025`).

#### `r2d2-client` changes since 112025
- **Bug fix** ([#106)](https://github.com/JCSDA-internal/r2d2-client/pull/106): Silent no-op bug in `R2D2Client.execute_attr()` when a
  `hasattr` condition was not met, the function would silently do nothing instead
  of raising an error. Now fixed.

Usage:
`module use /discover/nobackup/gmaoswell/r2d2/modulefiles/core`
`module load r2d2-client/05302026`


Dooruk:
I tried loading and running this module yesterday and it didn't work. Could it be a permission problem?

On a general note, I'm wondering if we should switch all installed packages to `gmaoswell`? Does everyone here have access to that space?

mranst:
I don't have any permission to that directory either

ftgoktas:
Just added group permission to the directory. Can you try running the following and let me know if it works? 

```
module use /discover/nobackup/gmaoswell/r2d2/modulefiles/core
module load r2d2-client/05302026
```

mranst:
It works for me now, though I still don't have access to the base directory. If we're going to be using this as a shared location for modules, ideally it should have rX permission for other, as well (which your module does, but just to note)

ftgoktas:
The base directory has been updated to `drwxr-xr-x`, so others have rX access. I've also given group write access to `modulefiles/core/` and its subdirectories, so `g0613` group members should now be able to create and manage modules there.

mranst (dismissed):

Dooruk (changes requested):
Before we merge this in can you change the version in `__init__`as well? then @mranst  or I can create a release tag `v20260602`.

Dooruk (approved):

PR #800, opened 2026-06-01
Adding convert state for GEOS-ATMOS

### LINKED ISSUE #762: Suite to ease generation of diff-resolution background fields for atmos default cases
Changing the default experiment pointer requires generating background input files for corresponding default resolutions. 

For example, in changing the present 3DVAR-ATMOS suite to point to x0054 instead of x0050 we need to create c12 and c90 input background. Similarly, if the exp pointer is going to change for the ensemble SWELL suite, we need to convert the GEOSadas ensemble files to desired resolutions.

Layout of steps required for a suite that does so:

**VARIATIONAL CASE**

- [ ] retrieve tar-ball from GEOS experiment with original cubed background files, the so-called bkgcrst tar ball - and unfold files in run work directory - this task already exists in SWELL (get_background_geos_experiment.py)
- [ ] develop a new task that converters a single background file from its existing resolution to a user-specified resolution - the main JEDI exec for this is fv3jedi_convertstate.x - a templated version of convert state.yaml
- [ ] the main suite proposed here should call the single-conversion task above in such as way as to parallelize all the  conversion of all available background files.
- [ ] tar-up the output with the same name of the input tag ball - files in the new tarball should preserve the name of the original files (notice the get-background task mentioned above renames the files to a swell-desired name; care must be taken for original names to be availabe somehow.

**CUBED ENSEMBLE CASE**

This is similar to the above, except that the fetching ensembe background task should be the one used, ie, 

- [ ] the first item above gets replaced with get_ensemble_geos_experiment.py
- [ ] the single file conversion should be identical to that in the second item of the variational case above
- [ ] the parallelization should now be over instances of background (as above) and instances of ensemble members
- [ ] the last item must create an ens-tar-ball identical to that created in GEOSadas.

mer-a-o:
## Description

*Provide a detailed description of what this PR does.*

## Dependencies

## Impact
closes: https://github.com/GEOS-ESM/swell/issues/762


rtodling:
Sorry: I changed the title of this PR since it doesn't really has much to do w/ FP.

PR #802, opened 2026-06-02
Add SaveBackground task and ingest_background_cf suite for GEOS-CF JDI backgrounds

### LINKED ISSUE #788: Ingest suite for storing geos-cf jdi collection in r2d2
Adding an ingest capability to store geos-cf jdi collection in R2D2. These files need to be stored as symlinks on the geos-cf datastore. This can help us test the geos-cf server and also symlinks more extensively within the workflow. Some more details:

* backgrounds `jdi` collection is stored here: `/css/gmao/geos-cf/NRTv2/priv/ana/Y*/M*/D*/GEOS.cf.ana.jdi_inst_1hr_glo_C360x360x6_v72.YYYYMMDD_HHmmz.nc4`

* How r2d2.store should look like:

The forecast starts each day at 09Z. For a file that is valid at HHZ, the forecast initialization time should be 09Z, with a lead time (or step) equal to the difference between the valid time and 09Z.
For example, for a file valid at 20251002_1800z, the R2D2 date would be 20251002_0900z and the step would be PT9H.
Steps would range from PT0H for files valid at 09Z to PT23H for files valid at 08Z the following day.

```
r2d2.store(
    model='geos_cf'
    item='forecast',
    step='PT9H',
    experiment='geos_cf_v2',
    resolution='c360',
    date='20251002_0900z',
    source_file='GEOS.cf.ana.jdi_inst_1hr_glo_C360x360x6_v72.20251002_1800z.R0.nc4',
    file_extension='nc4',
    file_type='bkg',
    store_as_symlink=True,
)
```
 `save_forecast.py` task can be reused for this purpose.

ftgoktas:
Adds a `SaveBackground ` task and stores GEOS-CF JDI background files into R2D2 as symlinks. The task loops over 24 hourly steps (PT0H–PT23H), resolves each file's path via strftime on a configurable background_source_path template, checks the file exists, and calls `r2d2.store(..., store_as_symlink=...)`. The model is read from `self.__model__`. Also names the suite folder suites/r2d2_ingest/ since the suite now covers both observations and backgrounds.

How to test:

1. `swell create ingest_background_cf`
2. Set `dry_run` in experiment.yaml `false` to actually store files, or keep `true` to preview which files you will be storing.
3. `swell launch ...`

Note on R2D2 `store_as_symlink`: After creating the symlink, `file_util._set_permissions` calls `os.chmod` which follows the symlink to the source file and throws `PermissionError` since we don't own it. The symlink and DB entry are both created before the chmod. We catch and verify the symlink before continuing. I think this is a bug in the R2D2 client (`_set_permissions` should be skipped when `store_as_symlink=True`) and should be fixed there. Created an issue about this here: https://github.com/JCSDA-internal/r2d2-client/issues/112

Closes #788


src/swell/tasks/store_jdi.py
```diff
@@ -0,0 +1,139 @@
...
+                else:
+                    raise
+            stored += 1
+
+        verb = 'Would store' if dry_run else 'Stored'
+        self.logger.info(f'JDI ingest complete: {verb} {stored} files, {skipped} skipped')
+
+    # ------------------------------------------------------------------
+
+    def _resolve_jdi_path(self, template: str, valid_time: dt) -> str:
```
mranst: This feels like it could be made simpler if you just directly used python's datetime templating for jdi_source_template (unless this is an established convention from CF or jedi)

template = `/css/gmao/geos-cf/NRTv2/priv/ana/Y%Y/M%m/D%d/GEOS.cf.ana.jdi_inst_1hr_glo_C360x360x6_v72.%Y%m%d_%H%Mz.R0.nc4`

valid_time.strftime(template)

mranst (commented):
One small comment, but looks good. I was able to test the dry run

src/swell/tasks/store_jdi.py
ftgoktas: Done! Switched to `strftime`.

jeromebarre:
quick comment:
since this ingest suite is not observations but model backgrounds this not really fits under `suites/ingest_obs`.
A solution to this is to remove the `_obs` and call all these simply `ingest` or more specifically `r2d2_ingest` because that is what it is about.

ftgoktas:
> quick comment: since this ingest suite is not observations but model backgrounds this not really fits under `suites/ingest_obs`. A solution to this is to remove the `_obs` and call all these simply `ingest` or more specifically `r2d2_ingest` because that is what it is about.

I agree, just pushed a commit renaming the relevant file/folder names.

mranst (dismissed):
Looks good, thanks!

Dooruk:
Some comments:

- Can you create an r2d2 issue on JCSDA repo on the symlink problem you described above? 
- It's `jedi` not `jdi` 👀 

- In terms of task names, store jedi (or jdi 😄 ) is very ambiguous, an alternative might be `ingest_background` since we have the `ingest_obs` already. I also see this task is written more towards CF in mind (totally fine), so two options:

1) Name the task `IngestCFBackground`
2) Use method @mranst introduced and store model specific tasks under `geos_cf` folder, e.g., /tasks/geos_cf/ingest_background.py`

- Another suggestion regarding `store_symlink`. This should be optional, in other words it should be a key in `experiment.yaml`. Not sure what the default `bool` should be. This is another issue & PR though.

- Another note for the pipeline key. I've been meaning to create an issue regarding this, which we can discuss this over a dedicated R2D2 meeting as well. Currently ingest have two options. Either ingest to R2D2 directly or download, convert and ingest. For marine and atmosphere applications another option might be `convert` and `ingest` if there are netcdf files already on Discover. This maybe the case for operational applications where files are downloaded automatically via D-BOSS (GESDISC). So my suggestion is to have that pipeline key as a list. Let me think about it or let's discuss it so I can describe it better under an issue.

Sorry, long list 😟

src/swell/utilities/question_defaults.py
```diff
@@ -1762,7 +1762,55 @@ class window_type(TaskQuestion):
         prompt: str = "Do you want to use a 3D or 4D (including FGAT) window?"
         widget_type: WType = WType.STRING_DROP_LIST
 
-# --------------------------------------------------------------------------------------------------
+    # --------------------------------------------------------------------------------------------------
+
+    @dataclass
+    class jdi_source_path(TaskQuestion):
```
mer-a-o: Some suggestions for the names here. Feel free to take it or come up with better ones. :)
`jdi_source_path` -> `background_source_path`
`jd_experiment` -> reuse `background_experiment` key
`jdi_resolution` -> reuse `horizontal_resolution` key
`ingest_jdi_pipeline` -> `ingest_background_pipeline`

src/swell/suites/r2d2_ingest/suite_config.py
```diff
@@ -62,3 +62,22 @@ class SuiteConfig(QuestionContainer, Enum):
             qd.dry_run(False),
         ]
     )
+
+    ingest_jdi_cf = QuestionList(
+        list_name="ingest_jdi_cf",
```
mer-a-o: `ingest_background_cf` perhaps?

src/swell/tasks/save_background.py
```diff
@@ -0,0 +1,118 @@
...
+                    model='geos_cf',
+                    item='forecast',
+                    step=step,
+                    experiment=jdi_experiment,
+                    resolution=jdi_resolution,
+                    date=forecast_start.strftime('%Y%m%d_%H%Mz'),
+                    source_file=source_file,
+                    file_extension='nc4',
+                    file_type='bkg',
+                    store_as_symlink=True,
```
mer-a-o: I like @Dooruk idea of making `store_as_symlink` a choice for users.

src/swell/tasks/save_background.py
```diff
@@ -0,0 +1,118 @@
+# (C) Copyright 2021- United States Government as represented by the Administrator of the
```
mer-a-o, line 1: How about changing `store_jdi.py` to `save_background.py` or `store_background.py`. Is there a specific reason for `store` vs `save`?

ftgoktas: done! changed it to `save_background.py`

src/swell/utilities/question_defaults.py
ftgoktas: Renamed all these.

src/swell/tasks/save_background.py
ftgoktas: Done, `store_as_symlink` is now configurable via `experiment.yaml`.

ftgoktas:
> Some comments:
> 
> * Can you create an r2d2 issue on JCSDA repo on the symlink problem you described above?

Done! See https://github.com/JCSDA-internal/r2d2-client/issues/112

> * It's `jedi` not `jdi` 👀

That got me too 😄, @mer-a-o confirmed the `geos-cf` collection is called `jdi`.

> * In terms of task names, store jedi (or jdi 😄 ) is very ambiguous, an alternative might be `ingest_background` since we have the `ingest_obs` already. I also see this task is written more towards CF in mind (totally fine), so two options:
> 1. Name the task `IngestCFBackground`
> 2. Use method @mranst introduced and store model specific tasks under `geos_cf` folder, e.g., /tasks/geos_cf/ingest_background.py`

Switched to `SaveBackground`, which is model agnostic and cleaner. Happy to rename if we want to go in a different direction.

> * Another suggestion regarding `store_symlink`. This should be optional, in other words it should be a key in `experiment.yaml`. Not sure what the default `bool` should be. This is another issue & PR though.

Done, added as a `TaskQuestion` (default `True`) and written to `experiment.yaml`. We have a workaround to mitigate the symlink `PermissionError` so it never triggers.

> * Another note for the pipeline key. I've been meaning to create an issue regarding this, which we can discuss this over a dedicated R2D2 meeting as well. Currently ingest have two options. Either ingest to R2D2 directly or download, convert and ingest. For marine and atmosphere applications another option might be `convert` and `ingest` if there are netcdf files already on Discover. This maybe the case for operational applications where files are downloaded automatically via D-BOSS (GESDISC). So my suggestion is to have that pipeline key as a list. Let me think about it or let's discuss it so I can describe it better under an issue.

Agreed, we can discuss and open up a follow up PR.

> Sorry, long list 😟

🫡

src/swell/utilities/question_defaults.py
ftgoktas: These are all renamed now.

src/swell/utilities/question_defaults.py
```diff
@@ -996,6 +996,19 @@ class dry_run(TaskQuestion):
 
     # --------------------------------------------------------------------------------------------------
 
+    @dataclass
+    class store_as_symlink(TaskQuestion):
```
Dooruk, line 1000: Thanks! One quick comment as I'm working on my marine ingest PR. Can you make this key get used for `ingest_obs` as well?

ftgoktas:
Should we rename `SaveBackground` -> `IngestBackground` to match the `IngestObs` naming convention? They're both `r2d2.store` operations essentially.

mer-a-o:
I suggest using "ingest" for suite and then "save" for tasks, `saveObservation`, `saveBackground` or `saveRestart` etc. I think of "ingest" as a suite that includes several tasks such as (download), (convert), save, etc.

Dooruk:
Hmm, I consider `Save..` tasks more for online cycling storing and Ingest for offline storing (i.e., existing experiment or observations). So it would look like:

- Ingest for suite names (ingest_background, ingest_obs) -> uses Store tasks (storebackground, storeobservations)

- Cycling suites -> uses Save tasks

Does that make sense?

this means ingest_obs task would be renamed store_obs.. maybe another PR.

Also FYI Furkan, the documentation sidebar was messed up and I addressed that in my merged PR.

ftgoktas:
`store_as_symlink` is now wired into `IngestObs` as well. Suite configs for `ingest_obs_marine` and `ingest_obs_cf` both have `store_as_symlink: false` in `experiment.yaml`

mer-a-o (commented):
Changes look good. I'm testing this branch now. Can you please add a section in `r2d2_ingest.md` about ingesting cf backgrounds?

docs/examples/r2d2/r2d2_ingest.md
```diff
@@ -256,6 +256,30 @@ acquisition_method: 'cp'
...
 
+### `ingest_background_cf` — GEOS-CF JEDI hourly backgrounds
+
+```bash
+swell create ingest_background_cf
+# Set dry_run in experiment.yaml false to actually store files, or keep true to preview which files you will be storing.
+swell launch <experiment_path>
+```
+
+Registers GEOS-CF NRT JEDI hourly background files into R2D2 as symlinks. Files already
```
mer-a-o:
```suggestion
Stores GEOS-CF NRT `jdi` collection's hourly files into R2D2 as symlinks. Files already
```

src/swell/tasks/save_background.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def execute(self) -> None:
+        """Ingest NRT background files into R2D2 as symlinks.
+
+        Designed for collections where background files already exist on a
+        shared filesystem and only need to be registered in R2D2 via symlinks
+        rather than copied. Currently used for the GEOS-CF JDI collection.
+
+        The collection contains 1-hourly instantaneous analysis files with a
+        single forecast run initialising at 09Z each day. Steps PT0H (valid
```
mer-a-o:
```suggestion
        single forecast run initializing at 09Z each day. Steps PT0H (valid
```

docs/examples/r2d2/r2d2_ingest.md
```diff
@@ -256,6 +256,30 @@ acquisition_method: 'cp'
...
+exist on the CSS shared filesystem. Each cycle runs a
+single `SaveBackground` task that loops over 24 hourly steps (PT0H–PT23H) from the 09Z
+forecast start and calls `r2d2.store(..., store_as_symlink=True)` for each.
```
mer-a-o:
```suggestion
forecast start and calls `r2d2.store(..., store_as_symlink=True)` for each. Because of this, `start_cylc_point` must be set to the forecast initialization time (`09Z`), e.g. `2025-10-02T09:00:00Z`.
```

src/swell/tasks/save_background.py
```diff
@@ -0,0 +1,126 @@
...
+
+        - ``background_source_path``: strftime path template, e.g.
+          ``/css/gmao/geos-cf/NRTv2/priv/ana/Y%Y/M%m/D%d/
+          GEOS.cf.ana.jdi_inst_1hr_glo_C360x360x6_v72.%Y%m%d_%H%Mz.R0.nc4``
+        - ``background_experiment``: R2D2 experiment name (default ``geos_cf_v2``)
+        - ``horizontal_resolution``: R2D2 resolution string (default ``c360``)
+        - ``store_as_symlink``: if ``True`` (default), register files as symlinks
+          in R2D2 rather than copying them
+
+        The Cylc cycle point must be the forecast initialisation time,
```
mer-a-o:
```suggestion
        The Cylc cycle point must be the forecast initialization time,
```

src/swell/tasks/save_background.py
```diff
@@ -0,0 +1,126 @@
...
+        e.g. ``2025-10-02T09:00:00Z``.
```
mer-a-o, line 50: Can you add a guard to make sure `cycle_time` is always set to `[T09]`? Otherwise if user accidentally removes it, the suite will ingest files with incorrect `step`. I think the code should abort if `cycle_time` is set incorrectly.

ftgoktas: Added a guard that aborts immediately if the cycle time hour is not `09Z`.

mer-a-o (approved):
Thank you @ftgoktas. Looks great now.

Dooruk (approved):

Dooruk:
If Tier1 tests are passing this is good to go in

ftgoktas:
Passes Tier1 tests: https://github.com/GEOS-ESM/swell/actions/runs/28542629181

jeromebarre:
Sorry I was on leave and I didn't find the time to pay attention to this before leaving. I think with this current suite with overrides we can decide to have one experiment for obs or one experiment for bkgs or both. Correct me if I am wrong.
I wondering if we should however keep separate the suite for obs and the suite for bkgs as this could bite us back after some sime. The specs for ingesting backgrounds and obs are very different and could make things complicated in the future with a single suite. But maybe this is handled well enough at the task level `ingest_obs` and `save_background`. Which should btw have consistency with their name `ingest` vs `save`. I would prefer `ingest` but that's just me.

jeromebarre:
> Sorry I was on leave and I didn't find the time to pay attention to this before leaving. I think with this current suite with overrides we can decide to have one experiment for obs or one experiment for bkgs or both. Correct me if I am wrong. I wondering if we should however keep separate the suite for obs and the suite for bkgs as this could bite us back after some sime. The specs for ingesting backgrounds and obs are very different and could make things complicated in the future with a single suite. But maybe this is handled well enough at the task level `ingest_obs` and `save_background`. Which should btw have consistency with their name `ingest` vs `save`. I would prefer `ingest` but that's just me.

Should this be a new issue let me know. I can create one. If we think this is ok for now to have the same suite for all ingest then let's move on.

ftgoktas:
I am fine with renaming `save_background.py`/`SaveBackground` to `ingest_background.py`/`IngestBackground` for consistency.

I think we should keep `r2d2_ingest` as one suite for now, splitting now would duplicate suite boilerplate for two things that are conceptually the same operation (fetch a file, store it in R2D2) just with different source resolution. @Dooruk

Dooruk:
I agree with name change, I may also have mentioned it previously. We have been using Save* type tasks for active cycles and ingest indicates registering offline created files and/or previous experiments in my mind.

Frankly I'm fine with same suite for both ingest operations. From the marine perspective we don't really have many experiment backgrounds outside of SWELL context. But if it gets too convoluted we can revisit it.

PR #804, opened 2026-06-11
EDA workflow

metdyn:
## Description

This is a first step to make EDA workflow run within SWELL framework.  These 32 ensemble members from @rtodling run independently on each node.  Both GSI B and all observation type from 3dvar-atmos are used here.  The single node run for each member is made possible with the current setting:  C90 resolution using thinned observations (rejecting 80% of obs at random). 
The memory usage on each node is roughly 72% (There are still details to work out and improve).

### The overall workflow:

The EDA-3D implementation here is a jedi block-EDA version. It is a very mechanical version of the EDA method.
A conventional 3D-Var minimization is invoked for each member state and the resulting ensemble anlysis states are final products of 3D-EDA. The observation perturbation term is added to the cost function when the member is a perturbed state.

1. The driver for this mechnism is in the .cylc file:
https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/flow.cylc#L79-L81
https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/flow.cylc#L170-L179

Each job submission completes one cost function minization for one member:
"swell task RunJediEdaExecutable $config -d $datetime -m {{model_component}} -imem {{i}}"

2.  The observation thinning can be disabled or tuned by setting obs_thinning_rej_fraction from zero (no thinning) to 1 (reject all).
https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/suite_config.py#L107

Thinning can also be turned off by commenting out this line:
https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/flow.cylc#L76

3.  Each EDA member run uses its own copy of observation files to prevent HDF5 file lock when multiple jobs are accessing a single file.   Hence  cleaning up  the directory for each 3D-Var run is important:
https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/flow.cylc#L88-L89

### The diagnostic tools:

For diagnostic purpose, we have designed handy codelets to compute mean/variance and diffstate (for increment, e.g.) in swell. 

A. To compute the mean and variance of a set of states including bkg and analysis,
user can define the file specs for a list containing multiple entries where each can contain
its own input file, output file, and grid_type. 

There are two ways to create the configure file.

Method-1. modify the EDA suite config file:

https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/suite_config.py#L108-L119

Method-2. create a section in override.yaml 
```
models:
  geos_atmosphere:
    ensmeanvariance_spec:
    - state: bkg
      fn_input: ebkg/mem%mem%/geos.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4
      fn_output_mean: geos.prior.mean
      fn_output_variance: geos.prior.variance
      grid_type: [ 'latlon', 'cs' ]
    - state: analysis
      fn_input: analysis/mem%mem%/eda.ana.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4
      fn_output_mean: eda.ana.mean
      fn_output_variance: eda.ana.variance
      grid_type: [ 'latlon', 'cs' ]
```
and run: `swell create eda_atmos -o override.yaml`  to override the config default

Finally to compute the mean and variance, run 
`swell task RunJediEnsembleMeanVariance  PATHTO/experiment.yaml -d $date -m geos_atmosphere`

B. To compute the difference between two states, similar approaches can be used.

Method-1. modify the EDA suite config file:
https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/suites/eda/suite_config.py#L120-L128

Method-2. create an override section in override.yaml 
```
models:
  geos_atmosphere:
    ensmeanvariance_spec:
    diffstates_spec:
      state1:
        fn_input: geos.prior.mean.%yyyy%mm%dd_%hh%MM%ssz.nc4
      state2:
        fn_input: eda.ana.mean.%yyyy%mm%dd_%hh%MM%ssz.nc4
      state_diff:
        fn_output: eda.mean-inc
        grid_type: [ 'cs', 'latlon' ]  
      state_type: ensemble
```
and run: `swell create eda_atmos -o override.yaml`  to override the config default

Finally to compute the diff states, run 
`swell task RunJediDiffstates PATHTO/experiment.yaml -d $date -m geos_atmosphere`

### How to run this block EDA code:
```
swell create eda_atmos -o override.yaml  -p nccs_discover_sles15 -k
swell launch   your_eda_suite
[override.yaml contains experiment_root and experiment_id]
```

Example output: T mean increment at 500 hPa :
inc_t_500

The work is advised by @rtodling. 
@mranst makes the flow.cylc loop work for ensemble members. 

### Timing results

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<body lang=EN-US style='tab-interval:.5in;word-wrap:break-word'>

Resolution | Random    obs. thinning | NX/NY | Nnode | Perhost | CPU time   Per member   (s)
-- | -- | -- | -- | -- | --
C90 | 80% | 4, 5 | 1 | NA | 347.0

</body>

</html>

In progress

## Additional features:

1. In analogy to fv3jedi ctest for 3d-EDA, we have implemented a choice to use hybrid-B matrix in 3D-Var.  Since this method is not tested, we have coded this function as an option for future use. 

https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py

A technical interesting part of the code is that I am using a string template in .py file, which is very useful here because it avoids writing 72 lines of pure real numbers into python code, and because we will need to tune and update this set of coefficient quite often later on, presumably.  This is viewed as a transition back from python code to yaml style. 

## Issues

1.  This code skips obs perturbation when the obs space contains covariance for R, because we did not find the corresponding source code in JEDI for this function. 

https://github.com/GEOS-ESM/swell/blob/35e846ce2371c9ac360ed13cf852082928bf922e/src/swell/tasks/run_jedi_eda_executable.py#L159-L173

2.  People should note that:
      Adding `air_pressure_levels` to the list of `analysis variable` will break the 3D-Var code with hybrid-B. (I assume there is an issue with bump code)   So I did not include it in the interface model config setup for EDA.  Here my assumption is that always having `air_pressure_at_surface` included in `analysis variable` can be sufficient. 

### To do list:

1.  We can eliminate hofx output,  increment output in the 3D-Var run for each member,  but we want to compare with output from more advanced EDA method.


src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.py
```diff
@@ -41,4 +41,7 @@ def background_ensemble(template_dict: Mapping) -> Mapping:
 
     return background_ensemble
 
+# Note: 4D-window
```
Dooruk, line 44: There are commented out parts throughout this PR. I will point them out as much as possible.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
```diff
@@ -0,0 +1,225 @@
...
+        local interpolator type: oops unstructured grid interpolator
+      state variables to inverse: &bvar
+        {%- for var in state_variables_to_inverse %}
+          - {{ var }}
+        {%- endfor %}
+    linear variable change:
+      linear variable change name: Control2Analysis
+      input variables: *bvar
+  weight:
+    value: 0.1
```
Dooruk, line 79: These hybrid weights could be a `experiment.yaml` key, maybe for a future PR. It would simplify testing of those weights going forward.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
```diff
@@ -0,0 +1,225 @@
...
+          nicas:
+            resolution: 6
+            explicit length-scales: true
+            horizontal length-scale:
+            - groups:
+              - common
+              profile:
+              - 3350515.0
+              - 3350515.0
+              - 3350515.0
```
Dooruk, line 117: Are these c90 and 72 vertical level specific values? There may be applications with 91 and/or 181 levels going forward. Same can be asked for `universe length-scale` as well I suppose.

From SOCA side I remember BUMP being resolution specific but I'm not sure how it is handled in this context.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
```diff
@@ -0,0 +1,225 @@
...
+    # Use ruamel.yaml to load the string instead of standard pyyaml
+    ruamel_yaml = YAML()
+    background_error = ruamel_yaml.load(rendered_yaml_string)
+    cov_template = background_error['components'][1]['covariance']
+    ['members from template']['template']
+    cov_template['field io names'] = field_io_names
+    a = background_error['components'][0]['covariance']['linear variable change']
+    a['output variables'] = template_dict['analysis_variables']
+    a = background_error['components'][1]['covariance']['linear variable change']
+    a['output variables'] = template_dict['analysis_variables']
```
Dooruk, line 221: Could you explain the logic here? Lack of comments and my familiarity with this making it hard for me to understand this.

src/swell/suites/eda_atmos/flow.cylc
```diff
@@ -0,0 +1,257 @@
...
+
+            # Cycling VarBC is inactive, static bias files will be used
+            GetObsNotInR2d2-{{model_component}}: fail? => GetObservations-{{model_component}}
+            {% endif %}
+
+            # Perform staging that is cycle dependent
+            StageJediCycle-{{model_component}}
+
+            # Run Jedi variational executable
+            BuildJediByLinking[^]? | BuildJedi[^]  => SP1-{{model_component}}
```
Dooruk, line 75: This is something I remember from `localensembeda` application as well. I'm not sure the need for the Sync Points? Cylc handles dependencies gracefully, it just looks like this is an extra confusing step?

src/swell/suites/eda/flow.cylc
```diff
@@ -0,0 +1,257 @@
...
+#            # EvaObservations
+#            SP4-{{model_component}} => EvaObservations-{{model_component}}
+#
+#            # EvaJediLog
+#            SP4-{{model_component}} => EvaJediLog-{{model_component}}
+#
+#            # Clean up large files
+#            EvaJediLog-{{model_component}} & EvaIncrement-{{model_component}} &
+#            EvaObservations-{{model_component}} =>
+#            CleanCycle-{{model_component}}
```
Dooruk: commented out parts

src/swell/suites/eda/flow.cylc
```diff
@@ -0,0 +1,257 @@
...
+    {% endif %}
+
+#    [[RunJediEdaExecutable-{{model_component}}]]
+#        script = "swell task RunJediEdaExecutable $config -d $datetime -m {{model_component}}"
+#        platform = {{platform}}
+#        execution time limit = {{scheduling["RunJediEdaExecutable"]["execution_time_limit"]}}
+#        [[[directives]]]
+#        {%- for key, value in scheduling["RunJediEdaExecutable"]["directives"][model_component].items() %}
+#            --{{key}} = {{value}}
+#        {%- endfor %}
```
Dooruk: commented out parts

src/swell/suites/eda_atmos/flow.cylc
```diff
@@ -0,0 +1,257 @@
...
+        script = "swell task RunJediDiffstates $config -d $datetime -m {{model_component}}"
+        platform = {{platform}}
+        execution time limit = {{scheduling["RunJediDiffstates"]["execution_time_limit"]}}
+        [[[directives]]]
+        {%- for key, value in scheduling["RunJediDiffstates"]["directives"][model_component].items() %}
+            --{{key}} = {{value}}
+        {%- endfor %}
+
+    [[CleanEdaFiles-{{model_component}}]]
+        script = "swell task CleanEda $config -d $datetime -m {{model_component}}"
```
Dooruk, line 216: Do we need an extra task for this? There is a `clean` task already.

src/swell/suites/eda_atmos/flow.cylc
```diff
@@ -0,0 +1,257 @@
...
+
+    [[SP2-{{model_component}}]]
+        script = true
+
+    [[SP3-{{model_component}}]]
+        script = true
+
+    [[SP4-{{model_component}}]]
+        script = true
+
```
Dooruk, line 251: 4 sync points?

src/swell/tasks/clean_eda.py
```diff
@@ -0,0 +1,150 @@
...
+        if window_type == '4D':
+            self.jedi_rendering.add_key('background_frequency', self.config.background_frequency())
+
+        # Open the JEDI config file and fill initial templates
+        # ----------------------------------------------------
+        jedi_config_dict = self.jedi_rendering.render_oops_file(f'{jedi_application}{window_type}',
+                                                                window_type,
+                                                                jedi_forecast_model)
+
+        for imem in range(1, nmember+1):
```
Dooruk, line 122: could there be a clean logic with `analysis/*/*` ?

src/swell/tasks/run_jedi_eda_executable.py
```diff
@@ -0,0 +1,269 @@
...
+                print(f'f= {src_file}')
+                shutil.copy(src_file, xdir)
+
+            if imember > 1:
+                obs_cov_model = observer.get('obs error', {}).get('covariance model')
+                print(f'{observation}:  obs_cov_model = {obs_cov_model}')
+                if obs_cov_model and 'cross variable covariances' in obs_cov_model:
+                    print(f"Found cross covariance obs: {obs_cov_model}, skip perturbation")
+                else:
+                    print(f"No cross varaible covariance found: {observation}, Obs Error Diagonal")
```
Dooruk:
```suggestion
                    print(f"No cross variable covariance found: {observation}, Obs Error Diagonal")
```
nitpick

src/swell/tasks/run_jedi_eda_executable.py
```diff
@@ -0,0 +1,269 @@
...
+
+# clean up dir
+#        for observer in jedi_config_dict['cost function']['observations']['observers']:
+#            # Get observation name
+#            observation = observer['observation_name']
+#            # Delete input obsfile .nc4 files
+#            for file_path in glob.glob(os.path.join(mem_dir, f'{observation}*.nc4')):
+#                os.remove(file_path)
+#                print(f"Deleted: {file_path}")
+
```
Dooruk: commented out

src/swell/tasks/run_jedi_ensemble_mean_variance.py
```diff
@@ -71,56 +71,86 @@ def execute(self) -> None:
...
+#            fn_input = spec["fn_input"]
+#            fn_output_mean = spec["fn_output_mean"]
+#            fn_output_variance = spec["fn_output_variance"]
+#
+#            # Use the variables directly
+#            print(f"Input Path:  {fn_input}")
+#            print(f"Output Mean: {fn_output_mean}")
+#            print(f"Output Var:  {fn_output_variance}")
+#            print("-" * 40)
+
```
Dooruk: commented out

src/swell/utilities/scripts/create_mock_configs.py
```diff
@@ -51,9 +51,10 @@
                              'model': 'geos_cf',
                              'executable_type': 'variational'}
 
-defaults_dict['localensembleda'] = {'datetime': atmosphere_default_datetime,
-                                    'model': 'geos_atmosphere',
-                                    'executable_type': 'localensembleda'}
+# WIP: not ready for ctest
+# defaults_dict['localensembleda'] = {'datetime': atmosphere_default_datetime,
+#                                    'model': 'geos_atmosphere',
+#                                    'executable_type': 'localensembleda'}
```
Dooruk: ditto

src/swell/utilities/yaml_utils.py
```diff
@@ -0,0 +1,19 @@
+
```
Dooruk, line 1: Sure, can we erase this then:

https://github.com/GEOS-ESM/swell/blob/7f3ec3090f6789ce4c9733b9f53a51954328e05b/src/swell/tasks/run_jedi_local_ensemble_da_executable.py#L21-L34

src/swell/swell.py
```diff
@@ -268,9 +270,9 @@ def test(test: str) -> None:
 @click.option('-p', '--platform', 'platform', type=click.Choice(get_platforms()),
               default="nccs_discover_sles15", help=platform_help)
 @click.argument('suite', type=click.Choice(("hofx", "3dvar_marine", "3dvar_atmos",
-                                            "localensembleda", "3dvar_cycle")))
+                                            "3dvar_cycle")))
```
Dooruk, line 281: So we don't want `letkf` testing for tier1 anymore?

Dooruk (changes requested):
This is a massive PR. Thanks @metdyn for working on this. I have some comments mostly regarding how this was implemented. 

It worked for me but I will leave the scientific evaluation and JEDI config check to Ricardo. 

I would really like to have a documentation on EDA. This is something I'm aware of nut I'm not sure how `diff_states` work and some broad details on the implementation. You could put something under the `/docs`, you can see my Practical Examples for inspiration.

Dooruk:
Thanks, I just noticed you updated the description of this but a README type doc would be much appreciated.

One other thing, for tier1 testing purposes maybe we can have a simpler 5-10 member setup? Because this will request 32 nodes to run as things stand.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_ensemble.py
metdyn: I spotted quite a few of them and I will do a clean up. 
However, many of them are on purpose that I felt then useful to me. 
E.g., here, the "Note: 4D-window" is related to 4D-EDA and it comes from local ensemble DA.  I am not sure who first wrote, but I guess it comes from @rtodling.  I want to get myself reminded.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
metdyn: As I mentioned in the main page, this file is a backup option not used in current work. The coefficients for 72 levels come from @rtodling and I am amazed by how well it works compared with fv3-jedi ctest inputs related to bump, and I copied to here in an ad hoc way.  It is not tested and not being used.  They are parameterizations.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
metdyn: Correct, before I would add the key, @rtodling has disapproved its use, being not tested and not really useful.  Hence hard coded here.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
metdyn: There are three comments above. It is assigning 'analysis_variables' to 'output variables' for the two covariance components. Due to technical difficulties with template, the dictionary is filled posterior. It is much easier to check the output than reading the source, but this code is not used in EDA so the generated yaml is not shown.

src/swell/suites/eda_atmos/flow.cylc
metdyn: This is needed to delete the observation files and other files I stored in the analysis/mem00x/. directory for 32 independent MPI jobs. The pattern [*.txt, *.nc4] etc. from the `clean` task cannot do the job.

src/swell/suites/eda_atmos/flow.cylc
metdyn: 4 Global Sync Points within a model component, bingo!

src/swell/tasks/clean_eda.py
metdyn: Replacing this py code by`rm -rf analysis` is future option. I need to get familiar with these output files first.

src/swell/tasks/run_jedi_eda_executable.py
metdyn: Good point, deleted.

src/swell/tasks/run_jedi_ensemble_mean_variance.py
metdyn: deleted.

src/swell/utilities/scripts/create_mock_configs.py
metdyn: done

src/swell/utilities/yaml_utils.py
metdyn: deleted the replica, thanks.

src/swell/swell.py
metdyn: Not really, I will be updating the 2-step GETKF (observer + solver) soon, when I can focus on one target and keep the local ensemble da workflow having both 1-step and 2-step options.

src/swell/tasks/run_jedi_eda_executable.py
metdyn: solved.

src/swell/suites/eda_atmos/flow.cylc
metdyn: I find my sync points have a clearer logic than the 3dvar-atmos case.  Cylc has graphic and dependence tree and when I started this file, I tried to reuse the 3dvar-atmos case but I was confused by the cylc there, for there is at least one missed dependency near [ ?| ] segment there. 

So when I try to straight it up, it confused you. 
The logic in this file is the following:

preparation (clone, stage, get bkg, get obs) -> SP1 --> mksi/observation + filter -> SP2 
-> [ submit/wait for 32 job submission till complete]-> SP3 
-> mean/variance + diffstate -> SP4 --> post process, eva

src/swell/suites/eda/flow.cylc
metdyn: deleted.

src/swell/suites/eda/flow.cylc
metdyn: deleted.

rtodling (dismissed):
I think the machinery is working fine. The results still need attention. 

This is what I see for the cost of all 32 members. 

Image

You can see that one member looks very different than all others -  I understand that member one is the one not perturbed ... I belive this difference is a clear consequence of my having asked you to remove treat the members all not using the correlated channel obs errors since JEDI has a bug in the ensemble perturbation algo - which we already discussed and is recorded as an issue in UFO. I just want to have the fig here so we remember and see how simply it is to see the lack of consistency.  For now, I think this is fine.

There is however a behavior that I can't quite explain. This is seen in the gradient norm reduction here:

Image

Notice how for some members, the gradient gets reset. My suspicion is that there is something else odd about the perturbations. 

But I think none of these are relevant at the momember. This PR amounts to a tremendous amout of work and provide a great entry point for what what to do in EDA. 

As @metdyn and I have talked before, this approach is not quite going to be what we end up doing anyway - in all likelhood. 

This is a go to me. With the warning that users should be advised this is not ready for primetime.

metdyn:
Thank you, @rtodling,for your review!   And thank you for checking these details and good to know that!
Thank you, @Dooruk, for proof reading the code and review!  I will put the documentation for ensemble mean variance and for diff state in another PR.  I hope this PR can be merged soon and I have a second PR on EDA which needs the infrastructure from this PR.

jeromebarre:
I will try to review this this week if that's ok.
Ricardo, nice catch on https://github.com/JCSDA-internal/ufo/issues/4182. I think this bug only impacts instruments explicitly configured with `covariance model: cross variable covariances`. Which we don't use so far with composition EDA so we could have never catch that... I am concerned with the jumps in the second figure that seem to happen at 1e-4 NR. Is this linked to https://github.com/JCSDA-internal/ufo/issues/4239 ? Are you sure you don't use in the plotting script extra interative loop post minimization such as GMRESR to get B^-1?

For reference we have EDA J looking like this with skylab/ewok with GEOS-CF:

image

rtodling:
BTW: I just ran a test case without IR, and I still see the split in the cost functions with mem001 (un-perturbed) being very different than the others - or the other way around the others being too different from the unperturbed, just as shown about when all data are used.

There is something odd about this ... I tried using only sondes for this and I still get the same issue - where the Jo of the unperturbed member stands out away from those of the perturbed ensemble. This is counter intuitive to me; it seems wrong. 

Again, this should not hold this PR.

jeromebarre:
> BTW: I just ran a test case without IR, and I still see the split in the cost functions with mem001 (un-perturbed) being very different than the others - or the other way around the others being too different from the unperturbed, just as shown about when all data are used.
> 
> There is something odd about this ... I tried using only sondes for this and I still get the same issue - where the Jo of the unperturbed member stands out away from those of the perturbed ensemble. This is counter intuitive to me; it seems wrong.
> 
> Again, this should not hold this PR.

Yes I agree this issue seems to be potentially intrisinc to oops/ioda/ufo and not a workflow issue. I am however concerned by this and this should the topic of another issue for us to investigate. I am doing swell PR reviews today and tomorrow. I will try to have this PR review done asap.

rtodling:
Actually, I think I understand what's going on here. I forgot that what's really happening here is that not only the obs are perturbed but the backgrounds are different for each member. The behavior I had in my mind - where the unperturbed Jo sits in the middle of all perturbed Jo's - only happens when the backgrounds are the same (as in mean(xb)), ie, the only thing perturbed are the obs.

Now, the odd behavior of the norm of the gradient is present still w/ only soundes data. And even more odd it happens for the unperturbed member as well!

mem001_norm

I think the issue is associated w/ the tolerance set for the CG - present at 1e-8. I suggest changing that to 1e-3.

metdyn:
Hi @jeromebarre,
I have changed suite name from eda to eda_atmos and lowered the gradient norm reduction from 1.e-7 to 1.e-3, according to @rtodling suggestions and we would fix it sooner than later.  I think it will not change the cylc run process, and thanks for review.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/diffstates_output.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def diffstates_output(template_dict: Mapping) -> Mapping:
+
+    # diff states (inc)
+    grid_type = template_dict['diffstates_spec_gridtype']
+    prefix_output = template_dict['diffstates_spec'].get('state_diff', {}).get('fn_output')
+
+    stateType = template_dict.get('diffstates_spec_statetype')
+    if [stateType == 'ensemble']:
```
jeromebarre: Not sure why the if statement is in square brackets []. This will consider only that one element list and will always be true. Thsi might only work if  `state_type: "ensemble"` but break otherwise.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/two_states.py
```diff
@@ -0,0 +1,53 @@
...
+
+# --------------------------------------------------------------------------------------------------
+
+
+def two_states(template_dict: Mapping) -> Mapping:
+    f1 = template_dict['diffstates_spec']['state1']['fn_input']
+    f2 = template_dict['diffstates_spec']['state2']['fn_input']
+
+    stateType = template_dict.get('diffstates_spec_statetype')
+    if [stateType == 'ensemble']:
```
jeromebarre: Not sure why the if statement is in square brackets []. This will consider only that one element list and will always be true. Thsi might only work if  `state_type: "ensemble"` but break otherwise.

src/swell/configuration/jedi/oops/eda3D.py
```diff
@@ -0,0 +1,70 @@
...
+                    'end': self.template_dict['window_end_iso'],
+                    'bound to include': 'begin'
+                },
+                'geometry': self.interface_model('geometry'),
+                'analysis variables': self.template_dict['analysis_variables'],
+                'background': self.interface_model('background_eda'),
+                # GSI:B
+                'background error': self.interface_model('background_error_eda_gsiB'),
+                #
+                # Hybrid-B: not tested
```
jeromebarre, line 35: But the whole point of the EDA is to update and run ensemble or hybrid B. I think this should be activated and therfore motivating cleaning the code in background_error_eda_hybridB.py

src/swell/suites/eda_atmos/flow.cylc
Dooruk: Sorry, I will push back on this. I don't want to hold this PR back from merging but I don't think it helps with flow.cylc. Here is how the graph looks like (part of it) now:

Image

Curious to hear other opinions: @mranst @mer-a-o @jeromebarre

src/swell/tasks/clean_eda.py
```diff
@@ -0,0 +1,150 @@
...
+            mem_dir = f'analysis/mem{imem:003d}/'
+            d1 = os.path.join(self.cycle_dir(), mem_dir)
+            d2 = os.path.join(d1, 'fv3-jedi')
+            if os.path.exists(d2):
+                if os.path.islink(d2):
+                    # Only remove the symlink itself, never follow it
+                    os.unlink(d2)
+                    print(f"Deleted symlink dir: {d2}")
```
Dooruk: Sorry I missed this with my first review but let's swich to logger class for prints for consistency across the code.

Dooruk (changes requested):
I will just repeat that I find SyncPoint redundant. Once @jeromebarre does a pass we can work towards merging this

Also one minor comment on the `print`.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
```diff
@@ -0,0 +1,225 @@
...
+    env = Environment()
+    template = env.from_string(background_error_template)
+
+    # pass the combined dictionary here
+    rendered_yaml_string = template.render(**render_context)
+    # Use ruamel.yaml to load the string instead of standard pyyaml
+    ruamel_yaml = YAML()
+    background_error = ruamel_yaml.load(rendered_yaml_string)
+    cov_template = background_error['components'][1]['covariance']
+    ['members from template']['template']
```
jeromebarre: Is this an orphan line? Or is this supposed to be attached to the line above?
Like:
`cov_template = background_error['components'][1]['covariance']['members from template']['template']`
`field io names` might be at the wriong nesting level.
Also cov_template is not used currently and would likely break if used in eda3D.py

src/swell/suites/eda_atmos/flow.cylc
```diff
@@ -0,0 +1,243 @@
...
+            {% if models[model_component]['ensemble_num_members'] is defined %}
+            {% for i in range( 1, models[model_component]['ensemble_num_members'] + 1 ) %}
+            SP2-{{model_component}} => RunJediEdaExecutable_mem{{i}}-{{model_component}} => SP3-{{model_component}}
+            {% endfor %}
+            {% endif %}
+
+            # bkg/ana : mean/variance
+            SP3-{{model_component}} => RunJediEnsembleMeanVariance-{{model_component}}
+            RunJediEnsembleMeanVariance-{{model_component}} ==> RunJediDiffstates-{{model_component}} => SP4-{{model_component}}
```
jeromebarre: are we supposed to have `==>` here instead of `=>`? I am surprised this doesn't break.

src/swell/tasks/run_jedi_ensemble_mean_variance.py
```diff
@@ -71,56 +71,74 @@ def execute(self) -> None:
         # Ensemble
         self.jedi_rendering.add_key('ensemble_num_members', self.config.ensemble_num_members(None))
 
+        print(f'self.config.ensmeanvariance_spec = {self.config.ensmeanvariance_spec()}')
```
jeromebarre: was this a debug print and should be removed or is this needed?

src/swell/tasks/run_jedi_diffstates.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Geometry
+        self.jedi_rendering.add_key('vertical_resolution', self.config.vertical_resolution())
+        self.jedi_rendering.add_key('npx_proc', npx_proc)
+        self.jedi_rendering.add_key('npy_proc', npy_proc)
+
+        # Diffstates
+        spec_dict = self.config.diffstates_spec()
+        self.jedi_rendering.add_key('diffstates_spec', spec_dict)
+        print(f'diffstates = {spec_dict}')
```
jeromebarre: was this a debug print and should be removed?

src/swell/tasks/run_jedi_diffstates.py
```diff
@@ -0,0 +1,140 @@
...
+            self.jedi_rendering.add_key('experiment_root', 'experiment_root')
+            self.jedi_rendering.add_key('experiment_id', 'experiment_id')
+            self.jedi_rendering.add_key('cycle_dir', 'cycle_dir')
+
+        state_type = spec_dict.get('state_type')
+        self.jedi_rendering.add_key('diffstates_spec_statetype', state_type)
+
+        # loop output grid_type:
+        grid_type = spec_dict['state_diff'].get('grid_type')
+        print(f'grid_type = {grid_type}')
```
jeromebarre: was this a debug print and should be removed?

src/swell/tasks/run_jedi_eda_executable.py
```diff
@@ -0,0 +1,259 @@
...
+            jedi_config_dict['cost function']['observations'].pop('obs perturbations', None)
+        else:
+            jedi_config_dict['cost function']['observations'].update({'obs perturbations': True})
+
+        # create subdir
+        mem_dir = f'analysis/mem{imember:003d}/'
+        xdir = os.path.join(self.cycle_dir(), mem_dir)
+        os.makedirs(xdir, exist_ok=True)
+
+        for observer in jedi_config_dict['cost function']['observations']['observers']:
```
jeromebarre, line 150: in this for loop were the prints for debug reasons and should be removed?

src/swell/tasks/run_jedi_eda_executable.py
```diff
@@ -0,0 +1,259 @@
...
+        with open(jedi_config_file, 'r') as f:
+            yaml_content = f.read()
+
+        dir_list = ["bkg", "fv3files", "gsibec", "rcov"]
+        for i in dir_list:
+            j = f"fv3-jedi/{i}"
+            k = f"{mem_dir}{j}"  # Result: analysis/mem002/fv3-jedi/rcov
+            yaml_content = yaml_content.replace(j, k)
+        with open(jedi_config_file, 'w') as f:
+            f.write(yaml_content)
```
jeromebarre: This is raw string replacement to serialized yaml and therefore fragile. We should avoid that. Couldn't the `jedi_config_dict` be edited before dump to yaml? Other suites are not doing that.

jeromebarre:
> Hi @jeromebarre, I have changed suite name from eda to eda_atmos and lowered the gradient norm reduction from 1.e-7 to 1.e-3, according to @rtodling suggestions and we would fix it sooner than later. I think it will not change the cylc run process, and thanks for review.

Sounds good.
With that last commit the suite names are still '`eda_atmos_tier1`', '`eda`'.

src/swell/tasks/clean_eda.py
metdyn: @Dooruk, thanks for spotting the cylc graph problem!  I will change this along with the self.logger.

src/swell/suites/eda_atmos/suite_config.py
```diff
@@ -0,0 +1,142 @@
...
+                "state_type": "ensemble"
+                }),
+            qd.clean_patterns(['*.txt', '*.csv']),
+        ]
+    )
+
+    # --------------------------------------------------------------------------------------------------
+
+    eda = QuestionList(
+        list_name="eda",
```
jeromebarre: change this to `eda_atmos`?

metdyn:
> > Hi @jeromebarre, I have changed suite name from eda to eda_atmos and lowered the gradient norm reduction from 1.e-7 to 1.e-3, according to @rtodling suggestions and we would fix it sooner than later. I think it will not change the cylc run process, and thanks for review.
> 
> Sounds good. With that last commit the suite names are still '`eda_atmos_tier1`', '`eda`'.

Thanks, @jeromebarre, let me check this.

src/swell/tasks/clean_eda.py
metdyn: I looked at the flow again, and I think it is fine. 
I was worrying that why CloneGeosMksi-{{model_component}} was outside SP1 (sync point 1), but 
looking into the file  flow.cylc reveals that  CloneGeosMksi is embedded inside the [[graph]]/R1 section, hence it is independent from other flow segments.   Especially because it is a clone step, being outside SP1 makes the flow faster.

metdyn:
Thank you, @jeromebarre, you are right about the eda_atmos keyword in suite_config!  I corrected them. 
swell create eda_atmos -o override.yaml  -p nccs_discover_sles15 -k
swell launch   your_eda_suite
[override.yaml contains experiment_root and experiment_id]

src/swell/tasks/run_jedi_ensemble_mean_variance.py
```diff
@@ -71,56 +71,74 @@ def execute(self) -> None:
...
-            run_executable(self.logger, self.cycle_dir(), np, jedi_executable_path,
-                           jedi_config_file, output_log_file)
-        else:
-            self.logger.info('YAML generated, now exiting.')
+        for idx, spec in enumerate(meanvar_spec_dict, start=1):
+
+            self.jedi_rendering.add_key('ensmeanvariance_spec_item', spec)
+            state_name = spec.get('state')
+
+            for output_grid in spec.get('grid_type'):
```
mer-a-o, line 91: If both latlon and cs are selected, then `fv3jedi_ensmeanvariance.x` will run twice. However, it's possible to write both cs and latlon in a single execution of `fv3jedi_ensmeanvariance.x`. I have tested writing `latlon` using `ensvariance to structured grid` (example: https://github.com/JCSDA-internal/fv3-jedi/blob/develop/test/testinput/ensmeanvariance_gfs.yaml#L26).

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/two_states.py
mer-a-o:
```suggestion
    if stateType == 'ensemble':
```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/diffstates_output.py
mer-a-o:
```suggestion
    if stateType == 'ensemble':
```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
mer-a-o: is this line a typo? Perhaps:

```
cov_template = (background_error['components'][1]['covariance']
                ['members from template']['template'])
```

src/swell/suites/eda_atmos/flow.cylc
mer-a-o:
```suggestion
            RunJediEnsembleMeanVariance-{{model_component}} => RunJediDiffstates-{{model_component}} => SP4-{{model_component}}
```

src/swell/utilities/question_defaults.py
```diff
@@ -533,6 +546,19 @@ class ensemble_num_members(TaskQuestion):
 
     # --------------------------------------------------------------------------------------------------
 
+    @dataclass
+    class obs_pert_amplitude(TaskQuestion):
```
mer-a-o, line 550: `obs_pert_amplitude` is not used anywhere in the code.

src/swell/tasks/run_jedi_eda_executable.py
```diff
@@ -0,0 +1,259 @@
...
+                print(f'{observation}:  obs_cov_model = {obs_cov_model}')
+                if obs_cov_model and 'cross variable covariances' in obs_cov_model:
+                    print(f"Found cross covariance obs: {obs_cov_model}, skip perturbation")
+                else:
+                    print(f"No cross variable covariance found: {observation}, Obs Error Diagonal")
+                    obs_error_dict = {
+                        'covariance model': 'diagonal',
+                        'zero-mean perturbations': True,
+                        'member': imember,
+                        'number of members': nmember
```
mer-a-o, line 175: I'm not very familiar with this option in the code but if we are not perturbing mem 0, should this be `nmember -1`?

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/eda_analysis.py
```diff
@@ -7,22 +7,24 @@
...
-        'filetype': 'auxgrid',
-        'gridtype': 'latlon',
-        'datapath': template_dict['cycle_dir'],
-        'filename': 'geos.prior.mean.',
-        'field io names': field_io_names_ensemble
+    imem = template_dict.get('ensemble_imember', None)
+    analysis = {
+        'filetype': 'cube sphere history',
+        'provider': 'geos',
+        'datapath': f'./analysis/mem{imem:03d}',
```
mer-a-o, line 21: This will fail if `imem = None`

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/eda_varincrement1.py
```diff
@@ -0,0 +1,47 @@
...
+    experiment_id = template_dict['experiment_id']
+    imem = template_dict.get('ensemble_imember', None)
+    varincrement1 = {
+        'write increment': True,
+        'increment': {
+            'state component': {
+                'filetype': 'auxgrid',
+                'gridtype': 'latlon',
+                'datapath': '.',
+                'filename': f'analysis/mem{imem:03d}/{experiment_id}.inc_iter1.mem{imem:03d}.',
```
mer-a-o, line 25: This will fail if `imem = None`. Perhaps remove `None` in line 17?

mer-a-o (commented):
Great work @metdyn. I added a few comments to the PR. I'm planning to test it soon.

rtodling:
I have good news. The odd looking figure I showed for the norm of the gradient, suggesting the gradients were getting reset was no correct. My python program was fishy in how it was reading a numer with e-XX notation - it was simply truncating that out ... with a fix the the offending line in my program, here is a fig for the gradient from all 32 members an full observing system as default in this PR:

eda_norm_correct

rtodling:
@metdyn I have two requests:

1) can you please add back the analysis CV the air_pressure_levels variable?

2) can you also please set the default tolerance for the CG to be 1e-3 to be consisrtent w/ other atmos cases?

Thanks

src/swell/tasks/run_jedi_eda_executable.py
rtodling: There is no member 0 at this point.  The only member not being perturbed is mem001. We could create a mem000, but I think at this point this is a uncessary complexity. 

A viable mem000 is the mean, but I don't think that's a good option - the mean is aphysical. The other possibility is to ingratiate the DA w/ a non-perturbed member at the resolution of the ensemble - equivalent to the deterministic hi-res realization. But again, all that are configuration choices to be made in the future.

src/swell/tasks/run_jedi_ensemble_mean_variance.py
rtodling: As I recall, the structured grid  writes a latlon file that is inconsistent w/ GEOS ... I have to look at this again.

mer-a-o: This was merged today: https://github.com/JCSDA-internal/oops/pull/3305
The dimensions should be consistent with GEOS now.

metdyn:
>     1. can you please add back the analysis CV the air_pressure_levels variable?
> 
>     2. can you also please set the default tolerance for the CG to be 1e-3 to be consisrtent w/ other atmos cases?
Hi @rtodling, 
* I have added `air_pressure_levels`, it was taken away because I was trying to make 3D-EnVar case work.
* the current values of `gradient_norm_reduction` are 1.e-3 and 1.e-4 for 3dfgat_atmos and 3dvar_atmos, respectively. I have  set it to 1.e-3 for eda_atmos this morning.

metdyn:
@mer-a-o, thank you very much for your review!  I will get back to your questions tomorrow.

jeromebarre (dismissed):
See my comments. Needs some code cleanup.

I would push to clean the code a bit to make sure the current EDA uses an actual ensemble or hybrid B. Otherwise, it defeats the purpose of EDA.

I would really refrain on editing yamls with raw string replacements if possible.

jeromebarre:
> See my comments. Needs some code cleanup.
> 
> I would push to clean the code a bit to make sure the current EDA uses an actual ensemble or hybrid B. Otherwise, it defeats the purpose of EDA.
> 
> I would really refrain on editing yamls with raw string replacements if possible.

I just realize that a lot of my review comments are almost similar to Maryam's. Mine were done a few hours beofre hers but I didn't pushed the finish review button. I guess we are both using Claude for review... My two main comments above remain different though.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/two_states.py
metdyn: Good catch and thank you @jeromebarre!   Error corrected: it was my typo.  The conventional if a==b:  is what I intended to do. 
Digging further, I see the python rule:
```
a=1
b=1
if [a>b, b<a]:
    print(0)

if any([a > b, b < a]):
    print(1+1)

output:
0
```

ai tool:
In Python, any non-empty list is truthy, regardless of what's inside it.
bool([False, False])   # True
bool([])               # False

If you specifically want to pass a list of conditions and check if at least one of them is true, use Python's built-in any() function:
```
if any([a > b, b < a]):
    print(1+1)
```

metdyn: Thank you, @mer-a-o, you are right!   I have modified this to my local version to be saved.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
metdyn: Thanks for find this, really important here. 
(It is not an orphan, but a parented-child)
My typo, it is one line.  It is intensively used. 
    cov_template = background_error['components'][1]['covariance']['members from template']['template']
is a reference, then 
   cov_template['field io names'] = field_io_names
is used to add to background_error the field io names which I intentionally omitted from L84.

metdyn: Thanks, @mer-a-o, your code is right and adopted!

jeromebarre:
@metdyn @rtodling please share the override.yaml you are currently using for testing this.
Especially for obs. The current run date is not matching any obs that exist in the notR2D2 data dir:
`ioda_locations_not_in_r2d2: /discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda`

Also I understand the option to bypass r2d2 must be activated without it the suite creates both:
`GetObsNotInR2d2-geos_atmosphere`
and 
`GetObservations-geos_atmosphere` which actually tries to use r2d2 and tries to pull `empty_obs.nc4` file.

I am getting increasingly worried that we will use more and more complex overrides over time that are not shared. At least we need to have a directory where those are stored if we decide not have a separate repo for configs/overrides.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/diffstates_output.py
metdyn: Thanks, @jeromebarre.  Your statement is precise, and it is my typo, corrected.

metdyn: Thanks!

mer-a-o:
@metdyn, Do you expect the observations to be fetched from R2D2 for this experiment?

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/eda_analysis.py
metdyn: As this file is for output to eda_analysis only, and you are absolutely right about the undefended variable 'ensemble_imember'.  This variable is passed from jedi_rendering. I modified the upstream code to predefine these variables to None, then assert the failure if it is None for this eda version.

```
class JediConfigRendering():

    def __init__(
        self,
...
        # Predefine eda variables to dictionary
        self.__template_dict__['ensemble_imember'] = None
        self.__template_dict__['ensemble_ichunk'] = None
```

```
class eda3D(OopsConfig):

    def render_oops(self):
        assert self.template_dict['ensemble_imember'] != None, \
            f'ensemble_imember is None and eda3D.py fails'
```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/eda_varincrement1.py
metdyn: Ditto, solved from code modifications to proof the error.

src/swell/configuration/jedi/oops/eda3D.py
metdyn: I discussed with @rtodling before, and the vertical profile for the localization length (I think) was not tested, even though I accidentally ran this version of his profile for hybrid-B vs LGETKF vs 3D-Var, I was happy to see hybrid-B case is resembling 3D-Var in single obs response. Nonetheless, @rtodling wants this hybrid-B version be thoroughly tested first, hence as a backup in this PR.

src/swell/suites/eda_atmos/flow.cylc
metdyn: Good catch and thanks!  (corrected)
The flow was paused there;  Diffstates and Eva afterwards did not run (pretty bad).

metdyn: Thanks!

src/swell/tasks/run_jedi_diffstates.py
metdyn: have changed to logger.info and will be removed later on when we are more confident that the diffstate code runs without error.

src/swell/tasks/run_jedi_diffstates.py
metdyn: Ditto

src/swell/tasks/run_jedi_eda_executable.py
metdyn: Thanks, @rtodling, for the detailed information, very nice to know this!
It is good to get confirmed that the ebkg/mem001 is the non perturbed member. 

Just a heads up, I find that I have to use mem000 as the non-perturbed member in EDA control perturb method, as it was  implemented (hard coded) from UKMO. There, I rerouted the file names by soft links.

src/swell/tasks/run_jedi_eda_executable.py
metdyn: The story is below. 

L223-229 is the AI code I got from a previous inquiry.  ` jedi_config_dict` is complicated. 
I do share the same desire for a direct manipulation on jedi_config_dict but did not find one at that time. 
The process of dumping and rewriting YMAL was an AI solution and it was linux` sed` process, pretty straightforward.  It is not elegant, however, I donot think of it as fragile, at least not now.

On the other hand, I do find the direct dictionary solution to this problem with AI via recursive function. 
I like the python code below LESS than I like the simple `sed` method from dump, rewrite, and readin, because it requires less brain power and it is shorter.

My conclusion is  I still want to use the current ymal dump and rewrite code, for its simplicity. 


```
import yaml
import copy

def update_jedi_paths(data, mem_dir):
    """
    Recursively searches through a dictionary or list and replaces specific
    'fv3-jedi/...' paths within string values.
    """
    if isinstance(data, dict):
        # If it's a dictionary, apply recursively to all values
        return {k: update_jedi_paths(v, mem_dir) for k, v in data.items()}
    
    elif isinstance(data, list):
        # If it's a list, apply recursively to all items
        return [update_jedi_paths(item, mem_dir) for item in data]
    
    elif isinstance(data, str):
        # If it's a string, check for our target keys and replace
        target_subdirs = ["bkg", "fv3files", "gsibec", "rcov"]
        for sub in target_subdirs:
            key = f"fv3-jedi/{sub}"
            if key in data:
                # Replaces 'fv3-jedi/bkg' with 'mem_dir/fv3-jedi/bkg'
                data = data.replace(key, f"{mem_dir}/{key}")
        return data
    
    else:
        # Return integers, booleans, etc., as-is
        return data

# --- Example Usage ---

# (Assuming jedi_config_dict is already defined as in your prompt)
mem_dir = "mem001" 

# It is good practice to pass a copy if you want to preserve the original dict
updated_dict = update_jedi_paths(jedi_config_dict, mem_dir)

# Dump to YAML
yaml_output = yaml.dump(updated_dict, default_flow_style=False, sort_keys=False)
print(yaml_output)
```

Output:
```
cost function:
  cost type: 3D-Var
  jb evaluation: false
  time window:
    begin: '2023-10-09T21:00:00Z'
    end: '2023-10-10T03:00:00Z'
    bound to include: begin
  geometry:
    fms initialization:
      namelist filename: ./mem001/fv3-jedi/fv3files/fmsmpp.nml
      field table filename: ./mem001/fv3-jedi/fv3files/field_table_gmao
    akbk: ./mem001/fv3-jedi/fv3files/akbk72.nc4
    layout:
    - 1
    - 1
    npx: '13'
    npy: '13'
    npz: '72'
  analysis variables:
  - eastward_wind
  # ... [variables omitted for brevity] ...
  background:
    datetime: '2023-10-10T00:00:00Z'
    filetype: cube sphere history
    provider: geos
    compute edge pressure from surface pressure: true
    max allowable geometry difference: 0.001
    datapath:  ...
    filenames:
    - ./ebkg/mem002/geos.mem002.%yyyy%mm%dd_%hh%MM%ssz.nc4
    - ./mem001/fv3-jedi/bkg/geos.crtmsrf.13.nc4
# ...
  background error:
    covariance model: SABER
    covariance type: gsi hybrid covariance
    saber central block:
      saber block name: gsi hybrid covariance
      read:
        gsi akbk: ./mem001/fv3-jedi/fv3files/akbk72.nc4
        gsi error covariance file: ./mem001/fv3-jedi/gsibec/gsi-coeffs-gmao-global-l72x144y91.nc4
        gsi berror namelist file: ./mem001/fv3-jedi/gsibec/cli_gsibec_configuration_l72x144y91.nml
# ...
```

jeromebarre:
> @metdyn @rtodling please share the override.yaml you are currently using for testing this. Especially for obs. The current run date is not matching any obs that exist in the notR2D2 data dir: `ioda_locations_not_in_r2d2: /discover/nobackup/projects/gmao/dadev/rtodling/archive/542/prePP/ioda`
> 
> Also I understand the option to bypass r2d2 must be activated without it the suite creates both: `GetObsNotInR2d2-geos_atmosphere` and `GetObservations-geos_atmosphere` which actually tries to use r2d2 and tries to pull `empty_obs.nc4` file.
> 
> I am getting increasingly worried that we will use more and more complex overrides over time that are not shared. At least we need to have a directory where those are stored if we decide not have a separate repo for configs/overrides.

Please disregard most of this comment as the real issue was an incorrect r2d2 credential setup to access the data stored for this PR. This raised another issue that most of the swell users and experiments are running on JCSDA credentials and data stores. We should avoid that and move away from this JCSDA datastore ASAP. @ftgoktas

The override issue still stands but should be treated separately.

src/swell/tasks/run_jedi_eda_executable.py
metdyn: I check with AI the two method, the preferred version is modifying the dict in memory, the other method: write, replace-keywords, read back to yaml is prone to issues with disk IO, keyword blindness replacement (whenever it happens), and issues for windows. I will change this code

src/swell/tasks/run_jedi_ensemble_mean_variance.py
metdyn: changed to logger

src/swell/tasks/run_jedi_ensemble_mean_variance.py
metdyn: Thanks for letting me know, @mer-a-o. 

I did not use the structured grid, as it needs to specify another set of vertical grid, which is a reason to shy away from. 

I too find it annoying not being able to output cs/ll in one shot within fvejedi, someone needs to add this in jedi. 
Currently the ensmeanvariance exe runs four times, 2 x 2 for 
```
         CS   LL
bkg       1    2 
ana       3   4 
```

Since I find it troublesome to have both cs and ll, I wrote the items in experimetn.yaml to loop over bkg and analysis for both cs and ll. 

Here is the experiment.yaml you can override for your need, then you run 
  swell task RunJediEnsembleMeanVariance  experiment.yaml -d 2023-10-10T00:00:00Z -m geos_atmosphere

```
    # Configure the ensemble mean and variance specifications:
    ensmeanvariance_spec:
    - state: bkg
      fn_input: ebkg/mem%mem%/geos.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4
      fn_output_mean: geos.prior.mean
      fn_output_variance: geos.prior.variance
      grid_type:
      - cs
      - latlon
    - state: analysis
      fn_input: analysis/mem%mem%/eda.ana.mem%mem%.%yyyy%mm%dd_%hh%MM%ssz.nc4
      fn_output_mean: eda.ana.mean
      fn_output_variance: eda.ana.variance
      grid_type:
      - cs
      - latlon
```

We perform 1 task. It runs the exe 4 times and gets 8 out files. Done!
```
eda.ana.mean.20231010_000000z.nc4
eda.ana.mean.ll.20231010_000000z.nc4
eda.ana.variance.20231010_000000z.nc4
eda.ana.variance.ll.20231010_000000z.nc4
geos.prior.mean.20231010_000000z.nc4
geos.prior.mean.ll.20231010_000000z.nc4
geos.prior.variance.20231010_000000z.nc4
geos.prior.variance.ll.20231010_000000z.nc4
```
Files without grid_type labelled are the CS version output.

mer-a-o: You do not need to specify the vertical grid. You can set `all model levels: true`, details here: https://github.com/JCSDA-internal/oops/blob/develop/src/oops/base/StructuredGridWriter.h#L322

I don't think it's just annoying, I think it's re-calculating the same thing multiple time when it's not needed.

src/swell/tasks/run_jedi_eda_executable.py
metdyn: code updated and I changed the print to info.

src/swell/utilities/question_defaults.py
metdyn: I have just added this variable back to the code to produce a yaml in obs err when imem > 1 for perturbation, the default in jedi code is 1.0 and the eda_atmos suit can change it as needed but default to 0.5, same as in jedi ctest input.  Updated yaml input looks like, 
```
      observation_name: amsua_metop-b
      obs error:
        covariance model: diagonal
        obs perturbations amplitude: 0.5
        zero-mean perturbations: true
        member: 2
        number of members: 4
```

I am not sure if `zero-mean perturbations` will take effect when we run each member in a separate batch job, but it should be a second question in obs perturbation that we can come back later on.

src/swell/tasks/run_jedi_ensemble_mean_variance.py
metdyn: `all model levels: true` is more convenient, I can use that when need it. 
If you want to change it, you are welcome to do so. A robust infrastructure is my task goal. 
None the less, this issue did force me for alternatives, including inventing the wheel above. 

A digression here: for me, I hope the  jedi_ensmeanvariance_config_analysis.yaml file would have a list for variance and mean output.

The cpp code would do
  if it is a dict:
      remain same way
 if it is list:
     loop
         call "MODEL".IO(config.listx)


```
mean output:
 - filetype: auxgrid
   gridtype: latlon
   datapath: ./eda/run/20231010T000000Z/geos_atmosphere
   filename: eda.ana.mean.ll.
   field io names: *id001

 - filetype: cube sphere history
   provider: geos
   datapath: ./eda/run/20231010T000000Z/geos_atmosphere
   filename: geos.prior.mean.%yyyy%mm%dd_%hh%MM%ssz.nc4
   first: PT0H
   frequency: PT1H
   field io names: *id001
```

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/eda_analysis.py
metdyn: Again, I have updated the code with `assert` to crash and report the bug. The current design does not have a crash.

src/swell/suites/eda_atmos/suite_config.py
metdyn: You are right, changed.

src/swell/suites/eda_atmos/flow.cylc
mer-a-o: I agree that SPs are confusing. When I run the experiment and look at the tui, something like `SP3` doesn't tell me much about what's going on. The correct way is to use "family" of tasks. I have looked at it briefly when we were talking about running get observations tasks in parallel: https://github.com/GEOS-ESM/swell/pull/724#pullrequestreview-3904396661

mer-a-o:
I tried running this suite and `EvaIncrement-geos_atmosphere` task fails with this error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/discover/nobackup/mabdiosk/SwellExperiments/swell-eda_atmos/run/20231010T000000Z/geos_atmosphere/swell-eda_atmos.increment-iter1.20231010_000000z.nc4'
```

I did not use any override file and simply ran:
` swell create eda_atmos`

and then launched the experiment.

src/swell/tasks/clean_eda.py
metdyn: I think this topic can be resolved, remember that @rtodling and @mranst found the flow in 3d-atmos that I had wanted to retain from (3dvar-atmos or 3dfgat) is defective in its path for obtaining obs.

src/swell/tasks/run_jedi_ensemble_mean_variance.py
metdyn: This a good discussion on JEDI techniques and I will resolve this topic, since we diverge on how different JEDI techniques can handle the mean output for CS and LL, but we converge on the mean outputs from this PR is useful and solid.

src/swell/configuration/jedi/interfaces/geos_atmosphere/model/background_error_eda_hybridB.py
metdyn: Code corrected and topic resolved.

metdyn:
Hi All, as a rule the override file is not needed for PRs (excuse me for the confusion), otherwise the usage is diverging and thanks @jeromebarre for pointing this out. (My ensuring PRs will not contain it)

metdyn:
> @metdyn, Do you expect the observations to be fetched from R2D2 for this experiment?

I am unfamiliar with R2D2 so far, but it is definitely welcome to add more options. Currently as you may have found the EBKG path is hard coded in py and the obs are obtained from old script pointing to local IODA files on disk.  

The flow here only implemented the single point Analysis without any cycle.

metdyn:
>from @jeromebarre:  an incorrect r2d2 credential setup to access the data stored for this PR.
Thanks for pointing out this issue. I did not use r2d2 at all in this PR.  We are focusing on running the plain EDA algorithm and to pave way for the next EDA Lanczos PR.

metdyn:
> I tried running this suite and EvaIncrement-geos_atmosphere task fails with this error:
FileNotFoundError: [Errno 2] No such file or directory: '/discover/nobackup/mabdiosk/SwellExperiments/swell-eda_atmos/run/20231010T000000Z/geos_atmosphere/swell-eda_atmos.increment-iter1.20231010_000000z.nc4'

Thanks @mer-a-o for testing.  There was a typo "==>" issue pointed out from @jeromebarre (corrected) and I think there is a keyword match pattern in `eva_increment.py` that I found yesterday when running the next PR.  I will fix this and run a test.

jeromebarre:
> > @metdyn, Do you expect the observations to be fetched from R2D2 for this experiment?
> 
> I am unfamiliar with R2D2 so far, but it is definitely welcome to add more options. Currently as you may have found the EBKG path is hard coded in py and the obs are obtained from old script pointing to local IODA files on disk.
> 
> The flow here only implemented the single point Analysis without any cycle.

Currently the the suite without overrides fetches obs from r2d2...

metdyn:
@mer-a-o, the workflow now proceeds to generate all the graphics files, hence the eva filename not found problem is solved.

metdyn:
>  the suite without overrides fetches obs from r2d2...

That is nice and thanks for letting me know!

src/swell/suites/eda_atmos/flow.cylc
metdyn: Thanks for suggestions, @mer-a-o and @Dooruk!
I have changed SP2 and SP3 to EDA_start and EDA_end, which should explicit now. SP4 being renamed to SP2 only acts as sync point stand alone. I think this should be good!

Dooruk:
@metdyn, while it is not a strict requirement it would be nice to get this in before next week's SWELL documentation sprint. Do you think this is ready to be merged in soon?

metdyn:
Doruk, I believe it will be ready shortly!  Thanks.

metdyn:
@Dooruk,
This PR is ready. 
(The workflow is tested on discover and has run through EVA plot steps.)

Dooruk:
> @Dooruk, This PR is ready. (The workflow is tested on discover and has run through EVA plot steps.)

Sounds good. @jeromebarre you requested some changes and I don't want to dismiss them without your input. When you get a chance please take a final look so we can merge this.

Dooruk (approved):
Looks good from my end.

mranst (approved):
Looks good!

Dooruk:
Will merge this tomorrow if we don't hear back from Jerome.

jeromebarre:
I am away right now in a conference. Please go ahead and have this merged. Thanks!

PR #808, opened 2026-06-12
Filters are missing error parameter info for hyper IR

rtodling:
## Description

Some filters in the IR are missing the overwrite of the error parameters. This should be automatic in UFO, but it's apparently not - ithat is, when Obs errors are specified in the yaml, they should be passed down the filters - but this is not happening consistently in UFO. 

This  correction had been made for IASI, but somehow never made it into AIRS, CrIS-FSR and AVHRR.

This PR placed the fix in AIRS , CrIS's and AVHRR yaml's

## Dependencies

## Impact

- [ ] repo
- [ ] ...


rtodling:
This is an illustration, e.g., AIRS, for what happens before these changes and after. 

airs_before_fix

notice the high index channel numbers (all window-type)  and water changes (mid indices) have OmA (red larger than OmB (see top left and bottom right panels.

rtodling:
Hum: something is amiss - maybe this got fix in more recent versions of UFO ... I am not finding any difference in results with these changes ... which is not what I recall seeing in the past.

rtodling:
This is now here: [feature/rtodling/combo_for_x0054](https://github.com/GEOS-ESM/swell/pull/835)

https://github.com/GEOS-ESM/swell/pull/835

PR #810, opened 2026-06-15
Add additional parameter, allow directory setting for CreateMockConfigs

mranst:
This PR adds support for an 'additional parameter' to tasks and utilities. I've played around with this concept for a while. It would mostly be used for setting a parameter to `swell task` that would be controlled by cylc (Using it to set observations for GetObservations to operate on individually is a core example). I added it to `swell utility` for the purpose of allowing `CreateMockConfigs` to send output mock files to a specific directory, since for static builds of swell they will go to a build location and not the source directory.


Dooruk (approved):
Hmm, I like the idea. Would be useful while testing ingesting, converting observations. 

I guess mostly useful for observation specific tasks, couldn't think of more use cases.

PR #811, opened 2026-06-16
Set shorter slurm requests

mranst:
## Description

This PR sets more specific defaults for slurm time requests than the previous 1 hour default for most slurm tasks. Most are set to 20 minutes, with a few set to shorter or longer depending on what I've found during testing. Found mixed results, probably depending on cluster load. `3dfgat_atmos` executed almost 6 minutes faster compared to the default running concurrently, and `3dfgat_marine_cycle` showed no difference. It's worth noting that I based these figures on what I've seen running default suites, it's possible that they could take longer than the requested time with different settings. As a side note, I don't like the way that time defaults have to be set in swell currently, #656 makes this more sensible, in my opinion

This doesn't invalidate #697 at all, these time estimates can potentially be used to make more efficient requests to `salloc`, too.


ftgoktas:
Thanks! Looks good to me and I'll experiment these times with `salloc` requests.

ftgoktas (approved):

Dooruk (approved):
I see your new workflow PR also addresses this so fine by me.

Dooruk:
Hmm, I just hit the 20min limit with `3dfgat_atmos` suite. 20minutes might be too strict for var. I know the new workflow PR will address this but can we set the default to 30 or 40 minutes until then (it could be in one of the PRs that is about to go in)?

PR #813, opened 2026-06-17
Remove conda call, point towards obsolete directory

### LINKED ISSUE #399: Fix random module request
This needs to be changed to the `modules` file in the experiment folder:

https://github.com/GEOS-ESM/swell/blob/e951bac683b79b44737431c27a5a8301b991f4a1/src/swell/tasks/gsi_ncdiag_to_ioda.py#L207-L213

thanks @mathomp4 for pointing out

### LINKED ISSUE #812: wired path to personal directory
I am finding that gsi_ncdiag_to_ioda.py has a path wired to the following installation of miniconda:

/discover/nobackup/drholdaw/opt/modulefiles/core/

I believe we want to clear this out, plaese.

mranst:
## Description

Closes #812. I can't figure out the purpose of the call to conda, but as far as I can tell it is unnecessary. 

The `convert_ncdiags` default suite is also broken, as far as I can tell partly due to some directories moving around
`SwellTestData/ufo_testing -> SwellTestData/obsolete/ufo_testing`
so I have updated the defaults, but there still seems to be something wrong with it

```
  File "/discover/nobackup/manstett/SwellExperiments/swell-convert_ncdiags/jedi_bundle/build/lib/python3.11/pyiodaconv/gsi_ncdiag.py", line 889, in toGeovals
    var_out = ncout.createVariable(geovals_vars["surface_height"],
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/netCDF4/_netCDF4.pyx", line 2956, in netCDF4._netCDF4.Dataset.createVariable
  File "src/netCDF4/_netCDF4.pyx", line 4174, in netCDF4._netCDF4.Variable.__init__
  File "src/netCDF4/_netCDF4.pyx", line 2113, in netCDF4._netCDF4._ensure_nc_success
RuntimeError: NetCDF: String match to name in use: (variable 'height_above_mean_sea_level_at_surface', group '/')
```


rtodling:
ok - that's good to know. Then we should be able to verify that we get identical results when this is removed from swell. Have you been able to verify that?

Dooruk (approved):
Thanks, there was another issue on this. Two birds one PR 👍🏼

mranst:
> ok - that's good to know. Then we should be able to verify that we get identical results when this is removed from swell. Have you been able to verify that?

I was able to verify it executes without error. I wasn't sure how to tell the results are the same

Dooruk:
> I was able to verify it executes without error. I wasn't sure how to tell the results are the same

ufo_testing compares final results to this file:

https://github.com/GEOS-ESM/swell/blob/develop/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/ufo_tests.yaml

We abandoned this for tiered testing but it is good to know it still passes. It will fail once we switch to x0054

rtodling:
> > ok - that's good to know. Then we should be able to verify that we get identical results when this is removed from swell. Have you been able to verify that?
> 
> I was able to verify it executes without error. I wasn't sure how to tell the results are the same

I assume you run the convert_ncdiags suite right?

rtodling:
> > I was able to verify it executes without error. I wasn't sure how to tell the results are the same
> 
> ufo_testing compares final results to this file:
> 
> https://github.com/GEOS-ESM/swell/blob/develop/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/ufo_tests.yaml
> 
> We abandoned this for tiered testing but it is good to know it still passes. It will fail once we switch to x0054

This would not play a role in this issue, would it?!

Dooruk:
> > > I was able to verify it executes without error. I wasn't sure how to tell the results are the same
> > 
> > 
> > ufo_testing compares final results to this file:
> > https://github.com/GEOS-ESM/swell/blob/develop/src/swell/configuration/jedi/interfaces/geos_atmosphere/observations/ufo_tests.yaml
> > We abandoned this for tiered testing but it is good to know it still passes. It will fail once we switch to x0054
> 
> This would not play a role in this issue, would it?!

no Michael is addressing two different issues

mranst:
> I assume you run the convert_ncdiags suite right?

Yes, though I had to play around with some settings to make sure that the block in question was triggered. Namely, making sure `single_observation` was set to true

PR #824, opened 2026-06-26
Update pinned versions to June 24, and use ip

mranst:
## Description

Updates pinned versions of JEDI to June 24. Also updates swell modules to use ip, resolving difference in build_jedi suite


Dooruk (approved):
👍🏼

PR #825, opened 2026-06-30
Fix cylc stall if comparison success

mranst:
## Description

This PR fixes an issue with the comparison suite, in which a successful comparison will cause a cylc stall as the scheduler is waiting for tasks that are never triggered. Also bumps the version to the latest tag


src/swell/suites/compare/flow.cylc
```diff
@@ -69,6 +71,14 @@
         script = "exit 1"
 
     {% for model_component in model_components %}
+    [[Pass]]
+        script = """
```
Dooruk, line 76:
```suggestion
    [[Pass]]
        #cylc set ... --out skip tells Cylc: “this task is intentionally not going to run; mark it as skipped.”
        script = """
```

One minor suggestion to add a comment as I had to search for the answer.

Dooruk (approved):
Didn't test it but I believe you 👍🏼

PR #821, opened 2026-06-24
Bypassing aircraft combine when conv from ncdiag

rtodling:
## Description

This aims at bypassing the combine step when converting ncdiag to ioda.

With this change, instead of ending up w/ a file named:

aircraft.yyyymmddThhmmssZ.nc4

you'll end up w/ two files like this:

aircraft_tsen_obs_yyyymmddhh.nc4
and
aircraft_uv_obs_yyyymmddhh.nc4

additional changes are needed to rename these files as:

aircraft_temperature.yyyymmddThhmmssZ.nc4
aircraft_wind.yyyymmddThhmmssZ.nc4

Notice also that the time stamp in the original split files are in the middle of the var time window (synoptic hour) whereas the final file should be named w/ a time at start of window, e.g.

mv aircraft_tsen_obs_2026011512.nc4 aircraft_temperature.20260115T120000Z.nc4

See issue #589 

## Dependencies

None

## Impact

None

PR #827, opened 2026-07-01
Rename aircraft observations

mranst:
## Description

@rtodling I believe this accomplishes the renaming of tsen and uv to temperature and wind, and also using the beginning of the window, as described in #821 .

### DIFF
--- a/src/swell/tasks/gsi_ncdiag_to_ioda.py
+++ b/src/swell/tasks/gsi_ncdiag_to_ioda.py
@@ -267,11 +267,11 @@ class GsiNcdiagToIoda(taskBase):
 
                 pattern = r"\b\w*aircraft\w*\b"
                 if not re.search(pattern, new_name):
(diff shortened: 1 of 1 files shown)


Dooruk:
Tagging @jemccurr for FYI. @rtodling mentioned this being the major obstacle before the x0054 changes

src/swell/tasks/gsi_ncdiag_to_ioda.py
jemccurr, line 1: Everything here looks perfectly fine to me.

jemccurr (commented):
Don't see any obvious problems here. Looks good and will keep note of changes for aircraft bufr converters

rtodling (approved):
This works for me. I tested by running convert_ncdiags and it did what I expected, producing the aircraft_wind and _temperature fles w/ the proper naming conventions - with the variables split as they originally are in GSI. 

Thank you.

Issue #773, opened 2026-04-27
Incorporating SPOC into swell convert_bufr suite


jemccurr:
Following on previous discussions we've had on bufr-query and obsbuilder scripts, I'm opening this issue for incorporating observation pre-processing from the obs group into swell. The obs group has been working on a fork of the NCEP SPOC library that is intended to consolidate the python scripts, modules and configuration yamls needed for pre-processing observations with the bufr-query obsbuilder interface(https://github.com/GEOS-ESM/spoc) This pre-processing has several advantages over approaches using executables. It prevents bloat in the rendered obs.yaml by doing operations that are more concerned with the inner workings of bufr (like retaining surface pressure observations only where the prepbufr data category is zero), as well as operations that involve intermediate calculations - insofar as these can be handled without background information. IMO, these derived variable calculations are awkward in a yaml context and much more natural to deal with in python. Lastly, pre-processing w/ obsbuilder is necessary for correct sorting of bufr data into the obs spaces that we intend to use w/ jedi. Splitting the data in bufr adpupa subsets into sonde and pibal obs spaces, or moving kx 183 observations into the sfcship obs space,for example, cannot (and are not currently) done using the executable approach but are implemented by scripts in our SPOC repo. 

I believe that the easiest way all this could be done in swell is by modifying the convert_bufr suite that Seth has been working on to include (a) a clone spoc task that grabs a tagged version of our SPOC repo and (b) a modified bufr_to_ioda.py that includes logic to execute the bufr conversion using the obsbuilder cli format (python [script.py] --input --output) rather than the format for executables (bufr2netcdf.x --yaml --input --output) if scripts corresponding to the obs space are present in SPOC/dump/scripts/atmosphere. There are a few satellite obs spaces that do not require additional pre-processing, so the ability to use executables could be retained in the case that configuration yamls are present in SPOC/dump/config/atmosphere without a matching obsbuilder script.

PR #830, opened 2026-07-08
Use ObsBuilder scripts from SPOC in bufr conversion

mranst:
## Description

#773

I've been experimenting a bit with using the spoc scripts in the bufr-to-ioda conversion to see if I could get it work. I believe it mostly works, but it is a somewhat crude implementation, so it will probably require some fine-tuning from people more familiar with obs handling

This PR adds the `CloneSpoc` task, which can grab a branch set by `spoc_tag` in `experiment.yaml`, or point to an existing `spoc_source_directory`. `Bufr2Ioda` then uses the scripts to run the conversion.

A few questions/notes:
- It seems like the scripts are expecting the corresponding yamls to live in the same directory they are being executed from? I may be misunderstanding something, but i added this process to CloneSpoc to accomodate:
https://github.com/GEOS-ESM/swell/blob/eca7d5c2088564ae36c0fcc065cebb1ed87dba3e/src/swell/tasks/clone_spoc.py#L66-L71

- The output ioda files seem to be combined with all satId's in one file, not sure how significant this is

- There are some obs types that have separate yamls under the bufr2netcdf.x schema, that don't have individual python scripts (I think this is noted in #773)


cohen-seth:
A few questions/notes:

_1.  "It seems like the scripts are expecting the corresponding yamls to live in the same directory they are being executed from? I may be misunderstanding something, but i added this process to CloneSpoc to accomodate:"_
- This is correct the scripts are expecting the corresponding yamls to live in the same directory they are being executed from.  This was done to preserve the yaml used for the conversion within the run directory for referring to later & for validation; that way, when we add the capability to customize the yamls and/or when spoc yamls are updated there will be a way of knowing how the files were converted. _(See directory tree below)_
- CloneSpoc will be very helpful. Previously, I had to manually add the spoc yamls to the `configuration/jedi/bufr2ioda/bufr2netcdf_x/` directory - should be able to remove these going forward.

run/20231010T000000Z/geos_atmosphere/
     │       
    ├── bufr/
    ├── ioda/
    ├── spoc_radiance_1bamua.yaml
    ├── spoc_radiance_1bmhs.yaml
    ├── spoc_radiance_atms.yaml
    ├── spoc_radiance_avcsam.yaml
    ├── spoc_radiance_avcspm.yaml
    ├── spoc_radiance_gpsro.yaml
    └── spoc_radiance_mtiasi.yaml


_2. "The output ioda files seem to be combined with all satId's in one file, not sure how significant this is"_
-    I believe we want them in separate files. Double check with @rtodling 
-    this functionality of `bufr2netcd.x` is still being developed. There are 2 methods, one method that only partially works and one method that requires more steps. So I had to use the 2nd method, the `--no-gather` method. We will want to switch to the other method once it's functional.

src/swell/tasks/bufr_to_ioda.py
cohen-seth, line 1: Are these obs builder python scripts placeholders?

mranst: Sort of? I tried to point to what was in the spoc scripts folder, mostly based on the name. I assume not all of them will line up correctly

cohen-seth:
Notes from meeting with @jemccurr:

1. obs builder can't handle relative paths. 
    -> will probably need help & guidance on how to handle this @mranst 
2. yaml files for conv types - 'reference time' within yaml needs to be set to the cylce time. 
    -> had to make edits within the yamls in the first version of bufr_to_ioda. should be straight forward.
3. need to be run in the same directory as the the yaml. 
    -> this is how the current bufr_to_ioda.py works. should be ok as is.
4. where is spoc clone stored/linked?

mranst:
@cohen-seth @jemccurr Changed this to copy scripts and yamls to a location in the cycle directory, should be better for making modifications that way.

For development purposes I'd suggest forking this branch, then opening PR's with this as the base when you want to submit changes

jemccurr (approved):
Everything here looks good to go

Issue #831, opened 2026-07-08
Model-dependent question not passing to flow.cylc fails experiment.yaml


metdyn:
`swell create` from develop generates incorrect experiment.yaml if the flow.cylc contains model-dependent questions/variables as condition or as a variable in the for-loop. 

Take for an example
(https://github.com/GEOS-ESM/swell/blob/2ef4a98ab200042a0c9678fe45021bf1dc6edbb9/src/swell/suites/localensembleda/flow.cylc)
https://github.com/GEOS-ESM/swell/blob/2ef4a98ab200042a0c9678fe45021bf1dc6edbb9/src/swell/suites/localensembleda/flow.cylc#L45-L51

If you run `swell create localensembleda  -o override.yaml -s slurmfile.yaml -k` on this branch (feature/ygyu/reproducer_for_num_members), you will find two errors:
1. `ensemble_num_members` is missing in experiment.yaml
2.  code crashes with error:   

```
File "/src/swell/src/swell/swell.py", line 131, in create
    create_experiment_directory(suite, input_method, platform, override_dict,
  File "/src/swell/src/swell/deployment/create_experiment.py", line 283, in create_experiment_directory
    prepare_cylc_suite_jinja2(logger, swell_suite_path, exp_suite_path, experiment_dict,
  File "/src/swell/src/swell/deployment/create_experiment.py", line 550, in prepare_cylc_suite_jinja2
    experiment_dict['models'][model_component][ensemble_aspect]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/python3.11/site-packages/ruamel/yaml/comments.py", line 926, in __getitem__
    return ordereddict.__getitem__(self, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'ensemble_num_members'
```

The reason lies in the design that intentionally only use suite questions to parse flow.cylc, which works fine until in the EDA implementation, we need to expand the capability to model dependent questions when using jinja2 template.  The root cause lies in

https://github.com/GEOS-ESM/swell/blob/cc024e25ea61e85690dc197d7b72ecad627a162c/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L470-L485

PR #832, opened 2026-07-08
 Add a few model dependent questions to parse flow.cylc in jinja2

metdyn:
## Description

This PR addresses issue https://github.com/GEOS-ESM/swell/issues/831, which is required for EDA ControlPert framework. 
It solves the issue in flow.cylc in case of complex rendering. 

       {% for i in range( 1, models[model_component]['ensemble_num_chunks'] + 1 ) %}
         [[RunJediEdaControlPertExecutable_chunk{{i}}-{{model_component}}]]
       {% endfor %}

The code here replaces self.experiment_dict by render_dict, where the latter contains the former dictionary plus two key-value pairs involving ensemble_num_chunks amd ensemble_num_members needed by EDA.

### DIFF
--- a/src/swell/deployment/create_experiment.py
+++ b/src/swell/deployment/create_experiment.py
@@ -537,7 +537,8 @@ def prepare_cylc_suite_jinja2(
                          'dictionary.')
 
     # Check if 'ensemble_hofx_strategy' appears anywhere in suite_file
-    ensemble_list = ['ensemble_'+s for s in ['num_members', 'hofx_strategy', 'hofx_packets']]
+    ensemble_list = ['ensemble_'+s for s in
+                     ['num_members', 'num_chunks', 'hofx_strategy', 'hofx_packets']]
     ensemble_list = ensemble_list + ['skip_ensemble_hofx']
     for ensemble_aspect in ensemble_list:
         if ensemble_aspect in suite_file:
--- a/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
+++ b/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py
@@ -457,6 +457,7 @@ class PrepareExperimentConfigAndSuite:
         # At this point the user should have provided the model components answer. Check that it is
         # in the experiment dictionary and retrieve the response
 
+        render_dict = copy.deepcopy(self.experiment_dict)
         if 'model_components' not in self.experiment_dict:
             self.logger.abort('The model components question has not been answered.')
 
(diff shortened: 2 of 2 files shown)


mranst:
I'm not against merging this if it solves the issue, but I'd like to maybe propose an alternative solution. I believe the issue you're encountering is that trying to use a task question in the runtime section of `flow.cylc` leads to swell not recognizing the task exists. This is because swell sweeps the partially filled flow.cylc for lines containing the phrase "swell task <task_name>", and these are not present if dependent on task questions. You can get around this by making sure that `swell task <task_name>` is present in the file in all cases.

(A) Either as a comment:
```
   # [[RunJediEdaControlPertExecutable-{{model_component}}]]
   #     script = "swell task RunJediEdiControlPertExecutable -m {{model_component}} -d $datetime experiment.yaml
   {% for i in range( 1, models[model_component]['ensemble_num_chunks'] + 1 ) %}
     [[RunJediEdaControlPertExecutable_chunk{{i}}-{{model_component}}]]
   {% endfor %}
```
(B) Or better yet, as a parent task. This has additional benefits such as being able to track the tasks execution as a group in cylc:
```
    [[RunJediEdaControlPertExecutable-{{model_component}}]]
        script = "swell task RunJediEdiControlPertExecutable -m {{model_component}} -d $datetime experiment.yaml
     {% if models[model_component]['ensemble_num_chunks'] is defined
     {% for i in range( 1, models[model_component]['ensemble_num_chunks'] + 1 ) %}
     [[RunJediEdaControlPertExecutable_chunk{{i}}-{{model_component}}]]
        script = "swell task RunJediEdiControlPertExecutable -m {{model_component}} -d $datetime -a {{i}} experiment.yaml
        inherit = RunJediEdaControlPertExecutable-{{model_component}}
     {% endfor %}
     {% endif %}
```

Admittedly, these strategies will not work if your control logic depends on tasks being in an adversarial relationship with each other. For example if you needed to have a case like:
```
{% if models[model_component]['x'] == 'A' %}
[[Task_A]]
{% else %}
[[Task_B]]
{% endif %}
```

metdyn:
@mranst,
We are on the same page for this issue!  The culprit is simply the few critical model-dependent questions that I need for EDA are not being parsed to or understood by flow.cylc, so that `RunJediEdiControlPertExecutable` is not registered as a task, hence its config/question variables are missing in experiment.yaml.  

Let's not use tricks to bypass the scientific need. A simple fix like this will solve the problem. 
Furthermore, I think the list of questions like this to be added to this code will go on and on, when more complex workflows are encountered.  The initial design simply did not foresee future expansions.

BTW, thanks for the workaround you provided.   I am sure case (A) will not work,  case (B) will bypass the issue but leave scientific confusions to others, because the task `RunJediEdiControlPertExecutable` will fail without ichunk being passed as a parameter.

PR #834, opened 2026-07-09
Convert ensemble questions to be suite questions

mranst:
## Description

#831 

Sorry for the back and forth on this. I just wanted to see if there was another way to solve this issue that fits with Swell's current design, as I think special handling of certain variables has potential to be confusing as well. This is also a bug in its own right that I wanted to sort out - there are duplicate suite and task questions for `ensemble_hofx_strategy` and `ensemble_hofx_packets`. Due to the way that the questions work now, only the task versions of these are counted, so we should convert them all to be one or the other as not to be confusing.

@metdyn Is there any reason these fields can't be suite questions? This would completely solve the issue of using them in flow.cylc, and they can be used by the task just like any other task question. I think you mentioned needing to have handling of model-dependent fields, but suite questions should be no different in this regard - you can access them in `flow.cylc` using `{{models[model_component]['ensemble_num_members']}}`.

Locally, I tried merging this branch with your branch `feature/ygyu/reproducer_for_num_members`, and it solved the issue during suite creation

Again, apologies for the back and forth. If this does not work for you, I will merge #832.


metdyn:
@mranst,

Thanks for your work, however, it is not what I want.

A simple question:
I have an override file, how do I specify  ensemble_num_members at suite level, if I want atmos and marine use EDA and LETKF, repsectively.

```
experiment_id: pert
generate_yaml_and_exit: true
models:
  geos_atmosphere:
    horizontal_resolution: "91"
    obs_thinning_rej_fraction: 0.98
    ensemble_num_members: 4
    ensemble_num_chunks: 2
    npx_proc: 1
    npy_proc: 1
    gradient_norm_reduction: 1.e-10
  geos_marine:
    horizontal_resolution: "91"
    ensemble_num_members: 16
    ensemble_num_chunks: none
    npx_proc: 1
    npy_proc: 1
```

mranst:
> @mranst,
> 
> Thanks for your work, however, it is not what I want.
> 
> A simple question: I have an override file, how do I specify ensemble_num_members at suite level, if I want atmos and marine use EDA and LETKF, repsectively.
> 
> ```
> experiment_id: pert
> generate_yaml_and_exit: true
> models:
>   geos_atmosphere:
>     horizontal_resolution: "91"
>     obs_thinning_rej_fraction: 0.98
>     ensemble_num_members: 4
>     ensemble_num_chunks: 2
>     npx_proc: 1
>     npy_proc: 1
>     gradient_norm_reduction: 1.e-10
>   geos_marine:
>     horizontal_resolution: "91"
>     ensemble_num_members: 16
>     ensemble_num_chunks: none
>     npx_proc: 1
>     npy_proc: 1
> ```

This can work, there are model-dependent suite questions, no different than tasks. The only reason this wouldn't work before is that the question was set to be registered exclusively under the `geos_atmosphere` config

https://github.com/GEOS-ESM/swell/blob/cc024e25ea61e85690dc197d7b72ecad627a162c/src/swell/utilities/question_defaults.py#L523-L532

mranst:
@metdyn I deleted the other special handling for ensemble variables in `create_experiment.py`.

I created a crude demonstrator branch just so I could make sure this works as expected: `demonstrator/mranst/ensemble_questions`, where I set different values per-model and use them in flow.cylc to set tasks

metdyn:
> @metdyn I deleted the other special handling for ensemble variables in `create_experiment.py`.
> 
> I created a crude demonstrator branch just so I could make sure this works as expected: `demonstrator/mranst/ensemble_questions`, where I set different values per-model and use them in flow.cylc to set tasks

@mranst, Thanks and let me check this code.

metdyn:
@mranst, 
I setup a more realistic flow.cylc using two models based on your demonstrator branch (feature/ygyu/demonstrator_mranst_ensemble_questions),
The output show:
    We made progress in parsing flow.cylc
    yes,  num_members appears in exp.yaml
    no,   for loop over atmos and marine,  the atmos script was solely overwritten by marine.

mranst:
> @mranst, I setup a more realistic flow.cylc using two models based on your demonstrator branch (feature/ygyu/demonstrator_mranst_ensemble_questions), The output show: We made progress in parsing flow.cylc yes, num_members appears in exp.yaml no, for loop over atmos and marine, the atmos script was solely overwritten by marine.

@metdyn The issue is that the template is looking for `geos_atmos`, which should be `geos_atmosphere`

Dooruk (dismissed):
This seems to be the right strategy for @metdyn's request.

metdyn:
@Dooruk ,  @mranst,

After digging into the SWELL code on `suite_questions`, I feel super confused and I would completely delete any suite_questions.yaml  and only leave suite_questions.py in SWELL.  This should serve the purpose and stop future confusions from new users. 

Here is the reason and questions:

1.  what is the difference between  suite_question  and  task_questions?
2.  The name `suite_questions` appears in 
```    
./suites/suite_questions.py  [ model independent ]
./configuration/jedi/interfaces/geos_atmosphere/suite_questions.yaml  [ model dep ]
./configuration/jedi/interfaces/geos_marine/suite_questions.yaml  [ model dep ]
./configuration/jedi/interfaces/geos_cf/suite_questions.yaml  [ model dep ]
```

I am confused why `./suites/suite_questions.py` is model independent and cylc-only questions,  while
`./configuration/jedi/interfaces/geos_atmosphere/suite_questions.yaml` is 100% model dependent, because it is under `interface/model`.  It was thought to be less frequently used variables, however, if you check deeper, how many people change cycle times from this line in suite_question:
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/configuration/jedi/interfaces/geos_atmosphere/suite_questions.yaml#L1-L3

how many people change analysis varaiables from this line in task_question:
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/configuration/jedi/interfaces/geos_atmosphere/task_questions.yaml#L2-L19

In here,  both `suite_questions.yaml` and `task_questions.yaml` are 100% model dependent within interface/model,  we can only keep one and delete the other. 


3.   Why does `./suites/suite_questions.py` contain `qd.marine_models`?
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/suites/suite_questions.py#L53-L59

Digging deeper, I see this breaks the rule and use a shortcut to parse Eva yaml, which is inconsistent with general treatment in flow.cylc,
```
3dfgat_marine_cycle/flow.cylc:    {% if 'cice6' in models["geos_marine"]["marine_models"] %}
3dfgat_marine_cycle/eva/increment-geos_marine.yaml:{% if 'cice6' in marine_models %}
```

mranst:
1. The difference between suite questions and task questions are the order that they are asked in. Suite questions are assumed to be the highest level, and are asked in order to define how the workflow takes shape (i.e. which tasks are going to be run). This is used in the step where swell takes the initial template, and figures out which tasks will be run. Based on this, the list of questions associated with each task are then added to create the complete set of configurations for experiment.yaml. Task questions are assumed to be of lower priority, and are asked later in the process. 

I don't really love that this distinction exists, but there are some reasons for it. Having a set of questions that is asked first is necessary to define the shape of the workflow. Consider `model_components`, as an example. This is used to determine which tasks are running under which model, which is then used to generate the complete list of questions.  This is why I think it makes sense to have these `ensemble` questions exist as suite questions, so they can be asked first to dictate the shape of the workflow.

To put it another way, suite questions are always asked, whereas task questions are only asked if they belong to a task that is being run. Thus, having a task question define which tasks will be run is problematic, as you have seen.

3. This is definitely confusing, so thank you for bringing this up. `suite_questions.py` just contains some common lists of questions that are used by a lot of the suites. You can see most suites in Swell import one of these lists to use as a base. It would probably be better named `common_question_lists.py` or similar.

https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/suites/localensembleda/suite_config.py#L13-L27

`suite_questions.py` is completely unrelated to the function of the `suite_questions.yaml` within each model folder, which is used to set the defaults for model-dependent questions if they are set to `defer_to_model`.

Since you mention it, `task_questions.yaml` and `suite_questions.yaml` could probably be combined into one file at this point. The distinction between suite and task questions was actually larger in the past (you couldn't access suite questions from within a task at all, and there were actually both task and suite versions of the same question, very confusing), so this is a leftover from then. I think this is probably out of scope for this PR, though

3. `marine` forms the basis for several suites (`3dvar_marine`, `3dfgat_marine_cycle`, etc.), so it's telling the list to use the question. I'm not sure why localensembleda uses the marine config, though. Probably my mistake when migrating to suite configs

> Digging deeper, I see this breaks the rule and use a shortcut to parse Eva yaml, which is inconsistent with general treatment in flow.cylc

The eva tasks each construct their own dictionary that is separate from `experiment.yaml`, and are called under a single model at runtime, so I don't see a need to change this

metdyn:
All the difficulties come from this file (@Dooruk)
https://github.com/GEOS-ESM/swell/blob/develop/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py

A **circle** exists in the code:

\# 2. Perform a non-exhaustive resolving of suite file templates
**self.experiment_dict** --> jinja partial parse -- >  suite_str -->  model_ind_tasks

\# 7. Perform a more exhaustive resolving of suite file templates
**self.experiment_dict** --> jinja partial parse -- >  suite_str -->  model_dep_tasks 

all tasks -->  **self.experiment_dict** [experiment.yaml]


This is why you have to pre-define suite questions to partially parse self.suite_str, which is vulnerable when users invent new flow.cylc.

In `prepare_config_and_suite.py`, because self.suite_str is origin and unique, but the partially parse suite_str can be full of errors, the `suite_question` is not easily defensible. 

A clean solution is to stop using the immature `suite_str`. 

**step-1**. Define a written rule in suite/flow.cylc: 
             SWELL only resolves `explicit task names` and possibly those containing `{{model_component}}` in task names.

meaning, this is valid:

localensembleda/flow.cylc:        script = "swell task GetEnsembleGeosExperiment $config -d x1 -m x2

or 

localensembleda/flow.cylc:        script = "swell task GetEnsembleGeosExperiment_{{model_component}} $config -d x1 -m x2


this is invalid:
localensembleda/flow.cylc:        script = "swell task GetEnsembleGeosExperiment_{{ensemble_name}} $config -d xxx -m xxx


**step-2**.  delete
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L426-L432

use:    model_ind_tasks = self.get_suite_task_list_model_ind(**self**.suite_str)

**step-3**.  delete
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L476-L485
and
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py#L500-L502

use:   model_dep_tasks = self.get_suite_task_list_model_dep(**self**.suite_str)




In summary, 

Partially resolving flow.cylc is the culprit.    Grep and parse directly flow.cylc tasks will resolve it. 
This shall save creating extra steps and efforts for  expanding suite_level questions. 

```
3dvar_cf_cycle/flow.cylc:        script = "swell task CloneJedi $config"
3dvar_cf_cycle/flow.cylc:        script = "swell task BuildJediByLinking $config"
3dvar_cf_cycle/flow.cylc:        script = "swell task BuildJedi $config"
3dvar_cf_cycle/flow.cylc:        script = "swell task StageJedi $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task GetBackground $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task GetObservations $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task RenderJediObservations $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task RunJediVariationalExecutable $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task GetRestartCf $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task PrepForecastCf $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task SaveForecastCf $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task SaveRestartCf $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task EvaJediLog $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task EvaIncrement $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task EvaObservations $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task SaveObsDiags $config -d $datetime -m {{model_component}}"
3dvar_cf_cycle/flow.cylc:        script = "swell task CleanCycle $config -d $datetime -m {{model_component}}"
localensembleda/flow.cylc:        script = "swell task CloneJedi $config"
```

How do you think?   @mranst @Dooruk

mranst:
@metdyn 
The reason it is designed this way is so that you can have control over which tasks are run, and thus which task questions are asked. If I understand correctly, what you propose would dispose of this entirely, only allowing `model_components` to be considered. Consider cases where you want to want to either run one task or another:

```
{% if x %}
[[Task_A]]
    …
{% else %}
[[Task_B]]
    …
{% endif %}
```

Under the current system, `x` being a suite question allows us to add only questions associated with Task_A or Task_B to experiment.yaml. Under your proposal, we would lose this capability, and force swell to ask all questions regardless of if they are needed or not. 

Here are some examples of this that are currently in swell. In 3dfgat_marine_cycle:
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/suites/3dfgat_marine_cycle/flow.cylc#L208-L219

`skip_ensemble_hofx` is a suite question that controls whether `RunJediHofxEnsembleExecutable` is run or not:
https://github.com/GEOS-ESM/swell/blob/debfec080d675faa2dd7d542b08f57adf0c74e81/src/swell/suites/localensembleda/flow.cylc#L183-L206

I appreciate you taking the time to think through this problem and look for solutions, but in my opinion there is utility to the current design that should not be quickly discarded. On that note, my PR #656 does away with this parsing of `flow.cylc` for "swell task" entirely, and allows for the use of python to directly tell swell which tasks are run. Suite questions are still necessary for this, but it solves the issues arising from template being partially resolved with missing variables, as you would get a clear error when the template is trying to access a variable that isn't defined at that stage.

metdyn:
>>Under your proposal, we would lose this capability, and force swell to ask all questions regardless of if they are needed or not.

Yes, questions from Task-A/B/C  will all be asked when CLI is used.  If you want the CLI questions right, for the flow:
```
{% if x %}
[[Task_A]]
    …
{% else %}
[[Task_B]]
    …
{% endif %}
```

Here is a solution:
```
1.  grep and find inclusive task list   [ A + B + all others ]
2.  create self.dict_inclusive  and donot ask questions
3.  repeat current process: 
       self.dict_inclusive ---> parse flow.cylc -->  "exact tasks  [ A or B + others ]" -->  CLI ask quesetions
     -->  self.experiment_dict
```

Does this solve this problem?

PR #837, opened 2026-07-17
Marine LETKF for the toy ocean model

Dooruk:
We are replicating https://github.com/JCSDA-internal/soca/blob/develop/test/testinput/letkf.yml with some changes here

----

This creates a deterministic LETKF option for the marine (ocean-only for now) DA applications. Caveats:

- Only for a toy model ensemble bkgr (5-deg tripolar grid)
- Only for a single time step
- This suite combines observer + solver together, which may bog down for 0.25-deg

So it's merely a first step.

Builds upon @metdyn's tremendous work, marine application required some exceptions and modifications, such as:

- No vertical localizations
- Horizontal Localization and Halo parts were directly added to the observation yamls. This didn't cause problems with the variational approach, but might warrant a conditional option.
- I didn't touch the linear observer part but that is most likely not exercised for marine?
- The task `obsFilters` is part of the suite but not exercised for marine, mainly for reference purposes for now.

There are also many options in the driver part in terms of which prior, posterior, ensemble, increment fields to write, which we need to decide depending on our workflow and EVA plots:

```yaml
driver_save_posterior_mean: true
driver_save_prior_mean: true
driver_save_posterior_mean_increment: true
driver_save_posterior_ensemble_increments: true
driver_save_posterior_ensemble: false
driver_save_prior_variance: true
driver_save_posterior_variance: true
driver_do_posterior_observer: true
```

FYI @DavidSRussell @vrx- 

## Impact
Huge


Dooruk:
@mranst I'm using the model specific folder for the first time here but the code test doesn't like it:

https://github.com/GEOS-ESM/swell/actions/runs/29612176850/job/87989120665?pr=837#step:7:22

`FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.local/lib/python3.11/site-packages/swell/tasks/get_ensemble.py'`

src/swell/utilities/question_defaults.py
```diff
@@ -1620,9 +1581,6 @@ class single_observations(TaskQuestion):
     class swell_static_files(TaskQuestion):
         default_value: str = "defer_to_platform"
         question_name: str = "swell_static_files"
-        models: List[str] = mutable_field([
-            "all_models"
-        ])
```
mranst, line 1690: Changing this to be a model-independent question wouldn't be a big deal, but `geos_cf` does use a different location for this, and since the `suite_config.py` for `3dvar_cf_cycle` sets this location under the `geos_cf` list, this explains the other issue with the code tests you're having

Dooruk: Great catch. I don't remember editing this :)

Did you figure this out from the Swell code tests message? I ran it in DEBUG mode but I didn't get it.

mranst: Yes, it kinda gets buried in a bunch of stuff since it doesn't cause immediate exit, but I saw it from the code test output:

https://github.com/GEOS-ESM/swell/actions/runs/29612176850/job/87989120665#step:7:51

```
   File "/home/runner/.local/lib/python3.11/site-packages/swell/deployment/prepare_config_and_suite/prepare_config_and_suite.py", line 264, in prepare_question_dictionaries
    if val['models'] != ['all_models'] and model not in val['models']:
       ~~~^^^^^^^^^^
KeyError: 'models'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/runner/.local/lib/python3.11/site-packages/swell/test/code_tests/suite_creation_test.py", line 40, in runTest
    raise Exception(f'Error generating {suite} experiment: {e}')
Exception: Error generating 3dvar_cf_cycle experiment: 'models'
```

rtodling (dismissed):
From my (ATMOS) perspective, I don't have a problem with anything here. This seems good - no intersections with Atmos application.

DavidSRussell:
Thanks, great progress! It ran fine for me.

What is in the bkg directory? Is it background members? Or background perturbations? And why are there 17 of them if ensemble_num_members is set to 3?

I don’t think I have enough information to judge whether the increments look as they should. The SSH increment seems out of whack compared to the SST and SSS, more than +/- 5 m in the Kuroshio and near the poles. There are large deep T and S increments from Argo in those places though, so that’s a good sign. As a side note, looks like we are not rejecting Argo under sea ice, which maybe we should revisit at some point?

src/swell/tasks/run_jedi_obsfilters_executable.py
```diff
@@ -24,6 +24,11 @@ class RunJediObsfiltersExecutable(taskBase):
 
     def execute(self, ensemble_members: Optional[list] = None) -> None:
 
+        # skip this task if the model is geos_atmosphere
```
DavidSRussell: You mean "is not" geos_atmosphere?

Dooruk: Yep, will address it

Dooruk:
> What is in the bkg directory? Is it background members? Or background perturbations? And why are there 17 of them if ensemble_num_members is set to 3?

bkg members, like in `ocean_das`. I created 17 just used 3 of them for testing, all test data gets linked though.

> I don’t think I have enough information to judge whether the increments look as they should. The SSH increment seems out of whack compared to the SST and SSS, more than +/- 5 m in the Kuroshio and near the poles. There are large deep T and S increments from Argo in those places though, so that’s a good sign. As a side note, looks like we are not rejecting Argo under sea ice, which maybe we should revisit at some point?

I wouldn't spend time on the results, this is only for plumbing but ARGO under sea ice we should put a pin on it.

Dooruk:
I could use an approval 😄

src/swell/configuration/jedi/interfaces/geos_marine/task_questions.yaml
```diff
@@ -116,7 +173,13 @@ observations:
...
 
 path_to_ensemble:
-  default_value: test
+  default_value: /discover/nobackup/projects/gmao/advda/SwellTestData/ensemble_marine
+
+skip_ensemble_hofx:
+  default_value: true
+  options:
+  - true
+  - false
```
mranst:
```suggestion
```

`skip_ensemble_hofx` is a suite question, so this setting won't be used

mranst (commented):
One note, otherwise happy to approve

mranst (approved):
👍

DavidSRussell (approved):

DavidSRussell:
FYI I just tried setting `local_ensemble_save_posterior_ensemble` and `local_ensemble_save_posterior_ensemble_increments` to `true`, and I don't think it's working properly. I get the files `ocn.mem%member%.an.2023-01-02T12:00:00Z.nc` and `ocn.mem%member%.inc.an.2023-01-02T12:00:00Z.nc` instead of one file for each ensemble member.

PR #840, opened 2026-07-20
Update pinned versions to July 16

mranst:
## Description

Zero diff in 3dvar_atmos and 3dfgat_marine_cycle


Dooruk (approved):
👍🏼 thanks for making the comparisons. 

I now remember we never addressed updating the installed module automatically, in other words calling `git pull` in a folder like `/discover/nobackup/projects/gmao/advda/JediOpt/src/jedi_bundle/sles15_skylab9`

PR #844, opened 2026-07-22
Add per-observation R2D2 provider mappings

### LINKED ISSUE #843: Add per-observation R2D2 provider overrides
`GetObservations` currently determines each obs's R2D2 provider exclusively from `observation_ioda_names.yaml`. This prevents suites and experiments from selecting a different provider through `suite_config.py` or `override.yaml`.

I propose having an `observation_providers` mapping mechanism that associates individual observation names with providers.

mer-a-o:
## Description

This PR adds configurable, per-observation R2D2 provider overrides to `GetObservations`.
Previously, the task always obtained the provider from `observation_ioda_names.yaml`.
Experiments can now select providers through either `suite_config.py` or `override.yaml` without changing the shared observation metadata.

Closes: https://github.com/GEOS-ESM/swell/issues/843

examples:

in `suite_config.py`:

```
qd.observation_providers({
    "tempo_no2_tropo": "nasa_v4",
    "tropomi_s5p_no2_tropo": "esa",
})
```

or in `override.yaml`:

```
models:
  geos_cf:
    observation_providers:
      tempo_no2_tropo: nasa_v4
      tropomi_s5p_no2_tropo: esa
```

Documentation and unit tests are also added. 

## Dependencies
None

## Impact
None


ftgoktas:
Thanks for this, looks good to me.

ftgoktas (approved):

Dooruk:
Sorry, I will challenge this a little bit (I'm slowly becoming a Ricardo 😅 ). My main reason is to avoid additional config/mapping variables whenever we can. I already think we are providing a lot of knobs and when I introduce this to people I receive lots of questions about all the keys but that's another issue. (Idea, maybe hide some default, mostly untouched keys).

For this issue, is this mainly for OSE/OSSE type experiments? I'm not sure about the ultimate use case. Changing of providers is not common for marine after initial tests (version changes are few and far in between) and AFAIK similar for NWP obs. It would be mainly adding a new observer and hence for our case another obs.yaml. I just don't know for your case.

I think the proposed approach will make comparing experiments a bit more complicated as one wouldn't know from IODA outputs or EVA plots alone and require accessing these keys for clarification. If `observation_ioda_names.yaml` is cumbersome I think having this information in the obs name mighe be better approach.

I propose we follow a strict obs name nomenclature, something like following, TBD:
{obs_type}_{inst_type}_{provider}_{(optional)version}

Clearly this nomenclature is not the case for the marine side and/or FP side currently but happy to work towards that (can't promise for NWP!) as we are actually in the process of testing different providers for the same obs.

mer-a-o:
> Sorry, I will challenge this a little bit (I'm slowly becoming a Ricardo 😅 ). My main reason is to avoid additional config/mapping variables whenever we can. I already think we are providing a lot of knobs and when I introduce this to people I receive lots of questions about all the keys but that's another issue. (Idea, maybe hide some default, mostly untouched keys).
> 
> For this issue, is this mainly for OSE/OSSE type experiments? I'm not sure about the ultimate use case. Changing of providers is not common for marine after initial tests (version changes are few and far in between) and AFAIK similar for NWP obs. It would be mainly adding a new observer and hence for our case another obs.yaml. I just don't know for your case.
> 
> I think the proposed approach will make comparing experiments a bit more complicated as one wouldn't know from IODA outputs or EVA plots alone and require accessing these keys for clarification. If `observation_ioda_names.yaml` is cumbersome I think having this information in the obs name mighe be better approach.
> 
> I propose we follow a strict obs name nomenclature, something like following, TBD: {obs_type}_{inst_type}_{provider}_{(optional)version}
> 
> Clearly this nomenclature is not the case for the marine side and/or FP side currently but happy to work towards that (can't promise for NWP!) as we are actually in the process of testing different providers for the same obs.

I fully agree that we should be mindful of the number of knobs in SWELL. One idea is to maintain a doc that lists the available configuration keys, their purpose, and whether they are commonly changed or mostly default. Then, for each PR that adds or changes config options, we could make reviewing that doc part of the checklist. LLMs can probably help keep this kind of documentation in sync.

For this specific issue, my main use case is testing different observation versions, or evaluating the impact of an update to an IODA converter. In those cases, I would like to be able to change the observation version through an override, rather than having to change the SWELL code/config itself.

We had something close to this capability last year, but the issue was that all the "obs x provider" combinations were checked by SWELL. So I agree with the general concern. I’m mostly trying to find a way to keep version/provider testing flexible without adding too much extra mapping/config complexity.

I think the nomenclature idea could be a good direction, especially because the observation name already propagates into IODA outputs, JEDI configs, and EVA plots. Having the version/provider visible in the obs name would make experiment comparisons more transparent. I think we need to talk more about it.

jeromebarre:
I understand the concern about having too many knobs, but at this point I think swell is far from minimalistic in that regard.......

As far as I can tell, this PR introduces a simple knob that only needs to be touched for non-default cases. From our future users perspective, I expect most people to use the ready-to-go suites and corresponding overrides we provide, so I'm not too worried about that. The concern is more valid from a code development/management standpoint, but this PR introduces minimal changes.

What I think we should keep in mind is: **we shouldn't be thinking in terms of file name conventions, but rather in terms of metadata conventions.**   A name like `no2_tropo_tempo_nasa_v4` could make things a bit more harder to parse reliably and creates a convention that has to be maintained everywhere downstream. My understanding is that this PR simply exposes control of an existing metadata field for R2D2...

Ideally such metadata info should also exist inside the ioda file inside the metadata group. The ioda format is flexible enough so it can accomodate this easily. Metadata travels with the data and is directly inspectable. We then wouldn't have to worry about file name conventions with additional snake/camel/the_whole_zoo cases...

jeromebarre (approved):

Dooruk (approved):
Sounds good, I see both of your points. We will be having to test different versions for some marine obs, so it is useful for us too. Changes look good on my end.

Dooruk:
Triggered this with tier1 tests.

github.com/GEOS-ESM/swell/actions/runs/30837931035

Something to clarify to developers in https://github.com/GEOS-ESM/swell/pull/847 perhaps is testing PRs with tier1 tests before merging. It is currently obscure in between other tabs:
https://geos-esm.github.io/swell/#/code_tests/suite_tests

PR #845, opened 2026-07-22
Special handling of goes-cf v2 backgrounds

mer-a-o:
## Description

This PR updates `get_background.py` to handle the special case of fetching GEOS-CF v2 backgrounds. GEOS-CF v2 backgrounds have a forecast start time of 09Z, and the `step` is calculated relative to that time. See the [SWELL R2D2 ingest documentation](https://geos-esm.github.io/swell/#/examples/r2d2/r2d2_ingest?id=ingest_background_cf-geos-cf-jedi-hourly-backgrounds) for details.

This PR also updates CF observation file extensions from `nc` to `nc4` for consistency.


src/swell/tasks/get_background.py
```diff
@@ -124,6 +145,15 @@ def execute(self) -> None:
         # Get name of this model component
         # --------------------------------
         model_component = self.get_model()
+        use_geos_cf_v2_background = (
+            model_component == 'geos_cf'
+            and background_experiment == 'geos_cf_v2'
```
jeromebarre: Should this have a more specific name for these backgrounds from the actual geos_cf_v2 runs. Should we use the word `oper` or `nrt` to identify what those are more distinctly than just `geos_cf_v2` which can be likely used somewhere else.

jeromebarre (dismissed):

src/swell/tasks/get_background.py
mer-a-o: good point. I changed it to `geos_cf_oper`

mranst (approved):
Looks good to me

jeromebarre (approved):

PR #835, opened 2026-07-16
Combined changes for x0054 update

rtodling:
## Description

This brings together to following PRs:

 - https://github.com/GEOS-ESM/swell/pull/816
 - https://github.com/GEOS-ESM/swell/pull/815
 - https://github.com/GEOS-ESM/swell/pull/808
 - https://github.com/GEOS-ESM/swell/pull/776
 - https://github.com/GEOS-ESM/swell/pull/678
 - https://github.com/GEOS-ESM/swell/pull/806
 - https://github.com/GEOS-ESM/swell/pull/805
 - https://github.com/GEOS-ESM/swell/pull/849
 - https://github.com/GEOS-ESM/swell/pull/821
 

## Dependencies

This requires JEDI now pinned to: 07162026  - found to be binary identical to present pinned version of JEDI.

PR #851, opened 2026-07-24
Update build with gsibec 1.4.3

mranst:
## Description

Updated July 16 build with gsibec 1.4.3. Zero diff.


mathomp4:
@rtodling just recently (today?) released gsibec 1.4.4. Should swell maybe move to use that? @rtodling will know.

rtodling:
> @rtodling just recently (today?) released gsibec 1.4.4. Should swell maybe move to use that? @rtodling will know.

yes - we should be using 1.4.4 - thanks @mathomp4 I mean to make this change here. The update is to no consequence to SWELL and its applications.

mranst:
Closed by #835

PR #853, opened 2026-07-31
Compress and ingest background files via R2D2

ftgoktas:
#790 already handles forecast compression for marine, so this PR just picks add ingest suite the feature to compress background files on the way in, and decompress them on the way back out, so we're not using too much disk space on backgrounds.

`SaveBackground` now takes `compress_output`/`compress_algorithm` (gzip or pigz)/`compress_pigz_threads`. When it's on, the file gets staged locally, compressed, stored, and cleaned up. 

`GetBackground` knows to look for the compressed record and decompress it back. It's off by default, so nothing changes unless you opt in.

##### Testing:

  ```bash
swell create ingest_background_cf
```
Find a 09Z NRT file to test against first:
`ls /css/gmao/geos-cf/NRTv2/priv/ana/Y<year>/M<month>/D<day>/ | grep _0900z`

Edit `experiment.yaml` with that date as `start_cycle_point`/`final_cycle_point`, and set `compress_output: true`. 
Leave `dry_run` at its default (`true`) for the first pass:

Then run: `swell launch <path>`
Check the logs, it should log `[DRY RUN] Would store for all 24 hourly steps...`

Once that looks right, set `dry_run: false`, and re-launch, and confirm that it stores background files.


Dooruk:
CF folks might want to handle this in a different way

Issue #854, opened 2026-08-04
3dvar_atmos_cycle has been lost


rtodling:
At some point we had a suite that cycled the 3dvar atmos analysis, giving the user the opt to cycle the VarBC results from one analysis to the next. It seems to me this got lost when we adopted the more python-oriented version of the suites and redefined the suites to be an explicit python class.

Can we please reinstate the suite?

mranst:
I looked back at before the PR you're referring to went in, and a bit before that, but couldn't find a `3dvar_atmos_cycle` suite. It looks like there's an option `cycling_varbc` in `3dvar_atmos` that does what you're suggesting, though. 

https://github.com/GEOS-ESM/swell/blob/69394f066a1b0d2e36ee9692db3d3e4048d3570c/src/swell/suites/3dvar_atmos/flow.cylc#L58-L67

Is it possible that these are essentially two suites in one that are controlled by this parameter?

Dooruk:
@mranst, you are right, there wasn't a suite named like that. This switch also impacts the `get_observations` task.

Bit of a warning though, this may cause some headache as we skipped R2D2 for x0054 and not work right off the bat.

PR #857, opened 2026-08-05
Fixes for 3dvar_atmos cycling_varbc

mranst:
## Description

This PR fixes a few issues related to setting `cycling_varbc` to true in `3dvar_atmos`. `flow.cylc` is fixed for this option, and `GetObsNotInR2d2` has been adapted to link previous cycle bias files the same way `GetObservations` does. 

I think there was a bug in the way that previous cycle bias files were linked that has been present for a while. The issue was that when determining the files to link, the process would take the current cycle's bias input file and only replace the cycle directory where it was stored. This would result in the file it was looking for having the wrong background time offset. For example, when running the `20251230T060000Z` cycle, the file would start as:
```
root_directory/20251230T060000Z/geos_atmosphere/aircraft_temperature.20251229T210000Z.acftbias_cov 
```
It would then only replace the cycle directory, so you would get
```
root_directory/20251230T000000Z/geos_atmosphere/aircraft_temperature.20251229T210000Z.acftbias_cov 
```
When in fact the filename should be:
```
root_directory/20251230T000000Z/geos_atmosphere/aircraft_temperature.20251229T150000Z.acftbias_cov 
```
As this is the `previous cycle time` minus `background time offset` for that period, though I could be wrong about how this is supposed to work

`abi_g16` is switched off, as it doesn't look like `x0054` has observation files for it. Should this normally be a failure state?


mranst:
#854

PR #858, opened 2026-08-10
Get observations individually for cf

mranst:
## Description

Changes `geos-cf` suites to fetch observations individually. Modifies `get_observations` to provide option of grabbing single observations per command-line argument. I figured this would be the best way to reduce redundant code.


mer-a-o:
Nice!! 

image

Can you also please add it to `3dvar_cf_cycle`?

mer-a-o (approved):

PR #859, opened 2026-08-11
docs: move How Swell Uses R2D2 earlier in R2D2 overview

ftgoktas:
Reorders sections for new users to see why Swell uses R2D2 right after what R2D2 is, before getting into deeper architecture. No wording changed, only section location and the Table of Contents order.


mer-a-o:
Closing as it will be covered in other PRs

PR #861, opened 2026-08-11
Document the `--skip-r2d2` flag for first-time users

ftgoktas:
Documents the `-k`/`--skip-r2d2` flag for new users.


mranst:
Don't you need credentials to fetch observations? `--skip-r2d2` doesn't get around that, only registering and storing

ftgoktas:
You're right, will fix the doc to say it skips registering/storing only, not fetching.

mranst (approved):

mer-a-o:
I did not merge this PR as part of the code sprint. I'm hoping to better define `skip-r2d2` behavior before updating the doc. This PR https://github.com/GEOS-ESM/swell/pull/876 will improve `skip-r2d2` and and we can update doc in there.

PR #873, opened 2026-08-17
Fix t1test and t2test commands

mranst:
## Description

Fixes test commands


Dooruk (approved):
Thanks, it's working for me now. I will merge this in.

PR #875, opened 2026-08-17
(docs-adjacent) Update PR template

Dooruk:
I copied @mathomp4's PR:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/739/changes

I think overall this is a good structure but we can be descriptive in terms expectations on the user for tier testing and define different labels?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,13 +1,43 @@
+<!--- These lines are comments. You can delete or ignore them -->
+<!--- NOTE: If your PR is trivial, feel free to delete the "Related Issue" -->
+<!---       "Testing" or other sections. -->
+
+<!--- Provide a brief summary of your changes in the Title above -->
+
 ## Description
+<!--- Describe your changes in detail -->
 
-*Provide a detailed description of what this PR does.*
+## Related Issue
+<!--- This project primarily accepts pull requests related to open issues -->
+<!--- If suggesting a new feature, observer, or change, please discuss it in an issue first -->
+<!--- If fixing a bug, there should be an issue describing it with steps to reproduce -->
+<!--- Please add the issue number below or delete the Closes line if this is not relevant -->
 
-## Dependencies
+Closes #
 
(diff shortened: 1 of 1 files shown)


mathomp4:
Obviously, you call can change the comments, etc. to whatever works best for swell in terms of labels, tests...

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,13 +1,43 @@
...
-- [ ] ...
+## Checklist:
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
+<!--- For the required labels: -->
+<!---   0 diff trivial: this change is to documentation or obviously does not affect the results -->
+<!---   0 diff structural: this change only moves code around -->
+<!---   0 diff: this change is to code and has no impact on the results -->
+<!---   non 0-diff: this change affects the results -->
+- [ ] I have tested this change with a run of SWELL suite (if non-trivial), see documentation xxx for testing details
```
mer-a-o, line 42: Another checkpoint for adding documentation? Something like

```suggestion
- [ ] I have tested this change with a run of SWELL suite (if non-trivial)
- [ ] I have added or updated the documentation to describe my new feature
```

mer-a-o (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
Dooruk: Good idea, I will change it once the sidebar is finalized.

mranst:
I like it, I guess we'll need to add the new labels this references. Maybe we could also push people to use the existing labels (atmosphere DA, observations, compo, documentation etc.). I know I could definitely be better about this

mathomp4:
> I like it, I guess we'll need to add the new labels this references. Maybe we could also push people to use the existing labels (atmosphere DA, observations, compo, documentation etc.). I know I could definitely be better about this

@mranst You can use a label enforcer as well if you have labels you require (see https://github.com/GEOS-ESM/GEOSgcm/blob/main/.github/workflows/enforce-labels.yml for an example)

PR #855, opened 2026-08-04
Use new soca2cice 

Dooruk:
Introduces the new comprehensive Postprocess method for soca2cice, introduced by Anna Shlyaeva. Only tested with 0.25-deg resolution.

Details 👇🏼 
https://github.com/JCSDA-internal/soca/pull/1246

or

https://github.com/JCSDA/soca/tree/develop/src/soca/PostProcess

Uses the basic setup, more capabilities to be added. Once we are happy the older method will be fully erased.

`fields_metadata.yaml` is updated in the static folder for file io.

### Dependencies
⚠️ Requires an August JEDI build after this PR was merged. SWELL has the proper version now

Need to use a version built after the following is merged:
https://github.com/JCSDA-internal/soca/pull/1259

FYI @mjkagnes123 @zhaobin74

PR #880, opened 2026-08-20
Update JEDI to August 14, change state variable names

mranst:
## Description

Moves the pinned versions of JEDI to August 14 (with GSIbec v1.5.0), and changes state variable names to account for changes in `fv3-jedi`. Adjusts some of the code tests to be in line with x0054 experiment.

Needed by #855 

Associated: GEOS-ESM/jedi_bundle#121


Dooruk:
TODO another time: maybe we should take out the nccs_cascade platform? not sure if anyone uses it as it is going to be decomissioned

Dooruk (dismissed):
Good to go in in my end, lot of changes in geos_atmosphere.

mranst:
> TODO another time: maybe we should take out the nccs_cascade platform? not sure if anyone uses it as it is going to be decomissioned

Sounds good. When is that supposed to be?

Dooruk:
> > TODO another time: maybe we should take out the nccs_cascade platform? not sure if anyone uses it as it is going to be decomissioned
> 
> Sounds good. When is that supposed to be?

This was sent out in March:

"
The NCCS is in the process of procuring the next generation of compute hardware for Discover. This hardware will be a departure from the traditional Discover Scalable Compute Units (SCUs) as we will be procuring a standalone Cray EX4000 solution from HPE. This system will contain 256 core AMD Turin CPUs and the high-speed interconnect will be Cray Slingshot instead of InfiniBand. We’ll send more information out in a separate communication.

There are facilities upgrades needed to power and cool the Cray so we will need to decommission the Cascade Lake (SCU16) nodes in advance of the delivery. Dates are still to be determined but since the Cascade Lakes are the last of our Intel CPUs, we wanted our users to have as much notice as possible that this change is coming. We anticipate that the new hardware will be delivered in the June to August timeframe, depending on the scheduling and completion of the facilities work.
"

Dooruk (approved):
Looks good!

Issue #852, opened 2026-07-29
Consolidating paths for CF cycling


jeromebarre:
https://github.com/GEOS-ESM/swell/blob/c46583fc8dc2de2ee274aa588ff1182c129938c1/src/swell/suites/3dvar_cf_cycle/suite_config.py#L95-L98

Most of this should be under 
`/discover/nobackup/projects/gmao/geos_cf_dev`
create a subdir called `swell`
and have:
`geos_cf_install_dir,  geos_cf_run_dir, swell_static_files, geosfp_path` in it.

We should probably have jedi_builds in it...

PR #881, opened 2026-08-20
Feature/consolidate cf paths

jeromebarre:
## Description

Consolidate paths to use a single a space under the geos-cf-dev space. This addresses issue: https://github.com/GEOS-ESM/swell/issues/852
This has been tested for the 3 cf suites, hofx, 3dvar and 3dvar_cycled.

I also took the opportunity to clean a bit the .j script.

Note: this is using an old build of geos-cf. The actual build of the current geos-cfv2 will be ported in a subsequent PR to swell and this will change the CI. 

## Dependencies

None

## Impact

None just cleaner


Dooruk (dismissed):
Looks good, seems like very local changes. Will trigger tier1 tests just in case.

mranst:
> Looks good, seems like very local changes. Will trigger tier1 tests just in case.

Already did, they're good.

mranst (dismissed):
Looks good, tier1 tests pass. After this, we can add `3dvar_cf_cycle` to tier2 tests

src/swell/suites/3dvar_cf_cycle/suite_config.py
```diff
@@ -89,13 +89,21 @@ class SuiteConfig(QuestionContainer, Enum):
...
-                '/discover/nobackup/mabdiosk/rundir/handle_inc/'
+                '/discover/nobackup/projects/gmao/geos_cf_dev/GEOS-CF3/handle_inc/'
                 'GCC_c90_FPens.geoscf_jedi.20210805_0600z.nc4'
             ),
-            qd.geos_cf_install_dir('/discover/nobackup/mabdiosk/GEOS-mil/GEOSgcm/install'),
-            qd.geos_cf_run_dir('/discover/nobackup/mabdiosk/rundir/GCv14.0_GCMv1.17_c90_Skylab'),
-            qd.swell_static_files('/discover/nobackup/projects/gmao/geos_cf_dev/SwellStaticFiles'),
-            qd.geosfp_path('/discover/nobackup/projects/gmao/geos_cf_dev/jbarre')
+            qd.geos_cf_install_dir(
+                '/discover/nobackup/projects/gmao/geos_cf_dev/GEOS-CF3/GEOSgcm_31062024/install'
```
mer-a-o: why `GEOS-CF3`?

src/swell/suites/3dvar_cf_cycle/suite_config.py
```diff
@@ -89,13 +89,21 @@ class SuiteConfig(QuestionContainer, Enum):
...
+                ),
+            qd.geos_cf_run_dir(
+                '/discover/nobackup/projects/gmao/geos_cf_dev/GEOS-CF3/GCv14.0_GCMv1.17_c90_Skylab'
```
mer-a-o: Can you drop `_Skylab`?

src/swell/suites/3dvar_cf_cycle/suite_config.py
mer-a-o: I'm looking at the directory. My suggestion would be to add a SWELL directory in `/discover/nobackup/projects/gmao/geos_cf_dev` and in that have everything you added for `GEOS-CF3`
so 
`/discover/nobackup/projects/gmao/geos_cf_dev/SWELL/GEOSgcm_31062024`, ...

mer-a-o (changes requested):

src/swell/suites/3dvar_cf_cycle/suite_config.py
jeromebarre: That is just name of the experiment that was in your local dir. That will go away in my next PR on GEOS model update. I can change it but at this point we lived with that name for quite a bit.

src/swell/suites/3dvar_cf_cycle/suite_config.py
jeromebarre: This a directory space for overall GEOS-CF v3 developments. IMO SWEL would be more confusing. What about GEOS-CF3-dev?

mer-a-o (approved):
Looks great. Thanks @jeromebarre

PR #882, opened 2026-08-24
fix docsify links

mer-a-o:
## Description

Docsify was not rendering the links in docs correctly. I went over all the links within the doc to make sure they are correct and change a few thinks in `index.html` to make sure the sidebar and the links render corrcetly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 (C) Copyright 2021- United States Government as represented by the Administrator of the National
 Aeronautics and Space Administration. All Rights Reserved.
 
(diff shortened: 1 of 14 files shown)


mranst:
These filepath changes end up breaking how the docs are rendered in Github, is there a way to make it work for both?

mer-a-o:
@mranst, the links in `_sidebar.md` are broken on GH because of `relativePath: true` in `index.html`. After my latest commit all the other pages should render fine on GH, except for `_sidebar.md`. We can have two versions of this file for one for docsify and on for GH but that means maintaining two files. I'm fine with either options.

mranst (approved):
That is fine with me, thanks

PR #883, opened 2026-08-24
bugfix for when obs provider is in override

mer-a-o:
## Description

The observation provider specified in the override wasn’t being picked up. This PR fixes that issue and corrects some of the default values for ingest_obs_cf.


mranst (approved):
Looks good to me

PR #885, opened 2026-08-24
5day window, background frequency and bug fix

### LINKED ISSUE #493: Get_background should avoid trying to use a CICE background if CICE is not in the marine_models key.
Get_background should avoid trying to use a CICE background if CICE is not in the marine_models key.

Dooruk:
I've been testing 5day configurations, they still don't work due to memory constraints but I'm making a PR because of the IAU fix.

Also adds capability to edit `diag_table` and `ice_in` automatically depending on the `background_frequency`. This allows switching between 6-h, 1-day or 5-day DA windows without changing the `geos_homdir`

Introduces `suite_to_run` as a renderable key. This is needed for some LETKF (ensemble) specific entries in observation YAMLs

Important: While adding some simple features I noticed a bug in terms if MOM6 IAU, so this fixes that as well.

### DIFF
--- a/src/swell/tasks/prep_coupled_geos_run_dir.py
+++ b/src/swell/tasks/prep_coupled_geos_run_dir.py
@@ -67,7 +67,7 @@ class PrepCoupledGeosRunDir(taskBase):
         # ----------------
         self.get_static()
 
-        # Modify ice_in and diag_table to allow different DA windows without changing the GEOSgcm 
+        # Modify ice_in and diag_table to allow different DA windows without changing the GEOSgcm
         # experiment directory
         # --------------------------------
         bkgr_freq = self.config.get_key_for_model('background_frequency', 'geos_marine', 'PT00')
--- a/src/swell/tasks/prep_coupled_geos_run_dir.py
+++ b/src/swell/tasks/prep_coupled_geos_run_dir.py
@@ -32,11 +32,12 @@ class PrepCoupledGeosRunDir(taskBase):
         As the name suggests, this task is geared towards coupled GEOSgcm simulations but
         the only difference between this and dataOcean ones should be in the get_static() method.
 
-        xx) Changes HOMDIR and EXPDIR in gcm_run.j to point to the forecast directory
-        xx) Provides consistency for GEOSgcm version by copying GEOSgcm.x from EXPDIR
-        xx) (Optional) Checks GEOSDIR, GEOSBIN, GEOSETC, GEOSUTIL are consistent with
+        01) Changes HOMDIR and EXPDIR in gcm_run.j to point to the forecast directory
+        02) Provides consistency for GEOSgcm version by copying GEOSgcm.x from EXPDIR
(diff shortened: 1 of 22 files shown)


src/swell/tasks/task_questions.py
```diff
@@ -475,6 +475,7 @@ class TaskQuestions(QuestionContainer, Enum):
             qd.obs_experiment(),
             qd.observation_providers(),
             qd.observing_system_records_path(),
+            qd.suite_to_run(),
```
mranst: Since this is a suite question, you don't need to add it to task questions

Dooruk: Ah I see

Dooruk: I took it out but it failed my active run task:

Image

Dooruk: Idk if it would not happen with a freshly created experiment. I can test and see

Dooruk: This doesn't work if I don't have it in `task_questions`, even with a new experiment. I might be doing something wrong though.

mranst: Oh, it's a weird interaction with the config class. Since it pre-defines `self.__suite_to_run__`, the method never gets added to the config. 

https://github.com/GEOS-ESM/swell/blob/3022db07c56891a6f9546b1336361bfe7094222c/src/swell/utilities/config.py#L65

https://github.com/GEOS-ESM/swell/blob/3022db07c56891a6f9546b1336361bfe7094222c/src/swell/utilities/config.py#L122

mranst: Just remove line 65 and it should fix it

Dooruk: claude suggested an alternative, your approach didn't work unfortunately..

Dooruk:
This is ready to go in with an approval, @DavidSRussell let me know if you have comments. I need to merge this with https://github.com/GEOS-ESM/swell/pull/855

mranst (approved):
All good on my end

src/swell/tasks/prep_coupled_geos_run_dir.py
```diff
@@ -24,19 +24,20 @@ class PrepCoupledGeosRunDir(taskBase):
     def execute(self) -> None:
 
         """
-        Copies GEOS HOMDIR files to the cycle forecast directory to prepare before executing
+        Copies GEOS HOMDIR files to the GEOSfcm forecast directory and prepares it before executing
```
DavidSRussell, line 27:
```suggestion
        Copies GEOS HOMDIR files to the GEOSgcm forecast directory and prepares it before executing
```

DavidSRussell (commented):
Looks good but I haven't tested yet. Perhaps we can test 5-day windows with a much-reduced observer list? Also I'm assuming we will want to reduce the large halo size--if it caused issues with 3DFGAT, I imagine it will with LETKF too.

PR #891, opened 2026-08-31
add override to download obs

### LINKED ISSUE #890: DownloadObs doesn't allow per-obs config overrides from experiment YAML
`DownloadObs` reads each observation's download configuration directly from `download_observations/<obs>.yaml ` with no way to override individual fields (e.g. s3_source) at experiment runtime. This is a problem for datasets like TROPOMI S5P where the MEEO S3 bucket organises files under either NRTI/ (Near Real-Time) or OFFL/ (Offline) depending on latency, and the correct product stream can vary by date. 

A new optional key `download_obs_config_overrides` needs to be added to the task config; after loading the per-obs YAML, any matching entry is shallow-merged into the config before the download is dispatched. Allowing, for example, switching s3_source from `s3://meeo-s5p/NRTI/…` to `s3://meeo-s5p/OFFL/…` entirely through the experiment override YAML without touching the source files.

This override functionality can be helpful for other types of downloads.

jeromebarre:
Resolves https://github.com/GEOS-ESM/swell/issues/890

It can be tested with https://github.com/GEOS-ESM/swell-config/pull/2

doing `swell create ingest_obs_cf -o <path>/swell-config/geos_cf/ingest_obs_geoscfv3.yaml`

### DIFF
--- a/docs/practical_examples/r2d2/r2d2_ingest.md
+++ b/docs/practical_examples/r2d2/r2d2_ingest.md
@@ -171,6 +171,30 @@ machine urs.earthdata.nasa.gov login <username> password <password>
 
 `DownloadObs` task places files in `<cycle_dir>/download/<obs_name>/`.
 
+#### Overriding download config per-observation
+
+`download_obs_config_overrides` lets you override fields of a
+`download_observations/<obs>.yaml` file from the experiment's suite config,
+without editing the source YAML. This is useful for switching a source's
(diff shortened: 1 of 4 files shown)


mer-a-o:
Looks good and works as designed. Thanks @jeromebarre. Can you please update the docs with this new feature in this PR?

jeromebarre:
I have added a section to the docs

mer-a-o (approved):

PR #895, opened 2026-09-01
Remove injection for JediOopsLogParser

mranst:
## Description

Very minor PR removing shell injection for single task. I'm not really sure why but this never worked with `shell=False`, even though it's not doing any kind of wildcard expansion or anything. I just switched it to python logic.


mer-a-o (approved):
Looks good.

PR #889, opened 2026-08-26
(docs) Marine Coupled GEOSgcm run description

Dooruk:
With the help of Claude and some manual input created a long documentation on executing GEOSgcm in SWELL. Take a look, hope it is helpful while I'm gone.

PR #900, opened 2026-09-15
Add hofx_marine suite (4D to work with 3DFGAT experiment outputs)

Dooruk:
I tried adding this in `hofx` but `geos_atmos` have certain tasks that always fail for `marine` so I skipped doing that. Perhaps we can rename `hofx` `hofx_atmos` and have a separate `hofx_coupled` suite going forward.

Note; I also tested having my `suite_config.py` under the `hofx_cf` suite. However, I noticed https://github.com/GEOS-ESM/swell/pull/858 has two potential issues while running multiple cycles with a lot of observers:

- Cylc struggles to run many jobs (>30-40) simultaneously. 
- If an individual observer fails the suite fails but maybe empty obs handling didn't work properly for me.

Anyways, thats' why for now this is a separate suite folder.

Of course the key entry is `qd.background_experiment(["swell-3dfgat_marine_cycle-ef812e05"]),`. This is my experiment but users can point to their previous experiments.

~TODO~
- [x] docs! (or let me do this in https://github.com/GEOS-ESM/swell/pull/889)


DavidSRussell (approved):
Thanks, this is cool. The default case runs fine for me. I might choose a different colormap and limits for temperature, but that’s nitpicking.

PR #903, opened 2026-09-23
Allow override of r2d2_id

### LINKED ISSUE #902: Can swell create use the same r2d2_experiment_id?
Can we reuse the unique r2d2-id if it is specified in `override.yaml` during `swell create`? The purpose is to pick up from a previously stalled or finished experiment. I don't think this is possible currently as swell checks if an experiment name is taken first. I guess this necessitates other R2D2 users shouldn't be able to write over to another user's existing experiment.

mranst:
## Description
#902. Accept override value as r2d2. Does NOT reregister experiment if it already exists, I believe this is the intended behavior but I can change it.


Dooruk (approved):
Perfect, it worked for me!

FYI @jeromebarre and @mer-a-o

Dooruk:
Not a big deal once the scrubber becomes active but perhaps we could use this change to use same `r2d2_experiment_id` for `gmao_ci`, for the verified base experiment at least.

